PR #2, opened 2024-08-13
feature: use generated conjure API to interact with scout

alxhill:


jhoch-nmnl:
We chatted in the office but, this is great!

PR #6, opened 2024-08-22
feat: add support for dataset ref names in a run

alkasm:
You can pass file paths to upload or a list of datasets (where the ref_name is pulled from `dataset.filename`) as before, or a dict mapping the ref_name to the dataset, which becomes the preferred ref_name in the run if it's included.

## Testing

```python
>>> import nominal as nm
>>> from nominal import Dataset, Run
>>> dataset = Dataset(nm.data.penguins(), "penguins.csv")
>>> r = Run(datasets=[dataset])
>>> r.upload()

Link to Run:
https://app-staging.gov.nominal.io/runs/
<Response [200]>
>>> r = Run(datasets={"test-ref-name": dataset})
>>> r.upload()

Link to Run:
https://app-staging.gov.nominal.io/runs/
<Response [200]>
```

As expected, the first run uploaded has the dataset ref_name "penguins.csv", the dataset filename:

Screenshot 2024-08-22 at 11 16 42 AM

and the second run uploaded has the dataset ref_name "test-ref-name":

Screenshot 2024-08-22 at 11 16 44 AM

### DIFF
--- a/nominal/nominal.py
+++ b/nominal/nominal.py
@@ -3,6 +3,7 @@ import io
 import os
 from datetime import datetime
 from math import floor
+from typing import Mapping, Sequence
 
 import jsondiff as jd
 import keyring as kr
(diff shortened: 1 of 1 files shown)


nominal/nominal.py
```diff
@@ -355,20 +357,20 @@ def __getattr__(self, k: str) -> None:
 
     def __init__(
         self,
-        rid: str = None,
-        path: str = None,
-        paths: list[str] = [],
-        datasets: list[Dataset] = [],
-        properties: dict = {},
-        title: str = None,
+        rid: str | None = None,
```
alkasm, line 360: I added all these optionals because pyright in vscode wasn't happy with `str = None` - and for a client lib I think it's best to be explicit on the types

nominal/nominal.py
```diff
@@ -355,20 +357,20 @@ def __getattr__(self, k: str) -> None:
 
     def __init__(
         self,
-        rid: str = None,
-        path: str = None,
-        paths: list[str] = [],
-        datasets: list[Dataset] = [],
-        properties: dict = {},
```
alkasm, line 362: fixed the mutable defaults here

nominal/nominal.py
```diff
@@ -355,20 +357,20 @@ def __getattr__(self, k: str) -> None:
...
-        title: str = None,
+        rid: str | None = None,
+        path: str | None = None,
+        paths: Sequence[str] = (),
+        datasets: Sequence[Dataset] | Mapping[str, Dataset] = (),
+        properties: dict[str, str] | None = None,
```
alkasm, line 364: we should prefer the generic containers (e.g. `Mapping` over `dict`, `Sequence` over `list`) in the client signatures to be more flexible for consumers. However I only updated the ones in this PR that I was actively changing - we can do that in a later cleanup PR

jackparmer (approved):

PR #9, opened 2024-08-26
bugfix: use uniform datetime fmt with 8601 type on the wire

alkasm:
Previously, `dataset.upload()` only worked for timezone-naive timestamps. Now we explicitly write all timestamps as timezone-aware and the backend can successfully parse them.

### DIFF
--- a/nominal/nominal.py
+++ b/nominal/nominal.py
@@ -18,12 +18,12 @@ from rich import print
 
 from ._api.ingest.ingest_api import (
     AbsoluteTimestamp,
-    CustomTimestamp,
     IngestService,
     IngestSource,
     RelativeTimestamp,
     S3IngestSource,
     TimestampMetadata,
+    Iso8601Timestamp,
     TimestampType,
     TriggerIngest,
 )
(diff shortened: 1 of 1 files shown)


nominal/nominal.py
```diff
@@ -130,7 +130,7 @@ def __upload_file(self, overwrite: bool) -> requests.Response:
             self.filename = default_filename("DATASET")
 
         csv_file_buffer = io.BytesIO()
-        self.write_csv(csv_file_buffer)
+        self.write_csv(csv_file_buffer, datetime_format=r"%Y-%m-%dT%H:%M:%S.%fZ")
```
alkasm, line 133: Ensures that timestamps will always be written with the same format. In the future this should be restricted to the specific column of interest, but this works for now.

nominal/nominal.py
```diff
@@ -192,9 +192,7 @@ def upload(self, overwrite: bool = False):
 
         nominal_ts_metadata = TimestampMetadata(
             "_python_datetime",
-            TimestampType(
-                absolute=AbsoluteTimestamp(custom_format=CustomTimestamp("yyyy-MM-dd['T']HH:mm:ss.SSSSSS", 0))
-            ),
+            TimestampType(absolute=AbsoluteTimestamp(iso8601=Iso8601Timestamp())),
```
alkasm, line 195: A custom format string is no longer required with the above format

alxhill (approved):

PR #10, opened 2024-08-27
fix: allow for relative timestamps in runs

alkasm:
Previously, datasets could use relative timestamps but the lib would fail if attaching those datasets to a run. Now the runs should be able to utilize datasets with relative or absolute timestamps.

Tested by uploading runs with mixed absolute and relative timestamps.

### DIFF
--- a/nominal/nominal.py
+++ b/nominal/nominal.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 import copy
 import io
 import os
-from datetime import datetime
+from datetime import datetime, timedelta
 from math import floor
 from pathlib import Path
 from typing import Mapping, Optional, Sequence
@@ -362,7 +362,7 @@ class Run:
         Print the Run datetime endpoints in a human-readable form
         """
         print("Run {} time:".format(endpoint))
-        unix_seconds = self._domain[endpoint]["SECONDS"] + self._domain[endpoint]["NANOS"] * 10e9
+        unix_seconds = self._domain[endpoint]["SECONDS"] + self._domain[endpoint]["NANOS"] // 1e9
         print("Unix: ", unix_seconds)
         datetime_endpoint = datetime.fromtimestamp(unix_seconds)
         print("Datetime: ", datetime_endpoint)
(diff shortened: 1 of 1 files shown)


nominal/nominal.py
```diff
@@ -491,7 +515,7 @@ def __set_run_unix_timestamp_domain(self, endpoints=["START", "END"]):
             unix = dt.timestamp()
             seconds = floor(unix)
             self._domain[key]["SECONDS"] = seconds
-            self._domain[key]["NANOS"] = floor((unix - seconds) / 1e9)
+            self._domain[key]["NANOS"] = floor((unix - seconds) * 1e9)
```
alkasm, line 518: this was inverted

nominal/nominal.py
```diff
@@ -362,7 +362,7 @@ def __print_human_readable_endpoint(self, endpoint):
         Print the Run datetime endpoints in a human-readable form
         """
         print("Run {} time:".format(endpoint))
-        unix_seconds = self._domain[endpoint]["SECONDS"] + self._domain[endpoint]["NANOS"] * 10e9
+        unix_seconds = self._domain[endpoint]["SECONDS"] + self._domain[endpoint]["NANOS"] // 1e9
```
alkasm, line 365: this was inverted, + an extra factor of 10 :)

nominal/nominal.py
```diff
@@ -455,12 +455,36 @@ def __init__(
...
+            start_dt = parser.parse(start)
+
+        # if you pass in an end, we use that
+        # if you don't, we can compute the end from the datasets
+        # if they use relative timestamps, we can compute the end since we know the start
+        if end is None:
+            for ds in self.datasets.values():
+                if ds.relative:
+                    assert ds.ts_col in ds.columns
+                    assert ds.relative_units == "SECONDS", "relative timestamps are only currently supported in seconds"
```
helenyugithub, line 478: why?

helenyugithub (approved):

nominal/nominal.py
alxhill: nit: I think we can ignore the nanos if we're converting to seconds as it's just intended to be the fractional part

nominal/nominal.py
```diff
@@ -455,12 +455,36 @@ def __init__(
...
+        else:
+            start_dt = parser.parse(start)
+
+        # if you pass in an end, we use that
+        # if you don't, we can compute the end from the datasets
+        # if they use relative timestamps, we can compute the end since we know the start
+        if end is None:
+            for ds in self.datasets.values():
+                if ds.relative:
+                    assert ds.ts_col in ds.columns
```
alxhill, line 477: nit: feels like this should be an assertion within the dataset itself

nominal/nominal.py
alxhill: okay the word "endpoints" really confused me here - I thought this meant service endpoints, not "start" and "end" 😆 

suggested follow up to call this "bounds" instead of "domain" and/or just have `self.start_ts` and `self.end_ts` as fields

alxhill (approved):
nits

nominal/nominal.py
alkasm: I wanted this PR to be purely a fix and not update functionality, though seeing this today...it's small enough that it probably would have been fine to add

nominal/nominal.py
```diff
@@ -455,12 +455,36 @@ def __init__(
...
+                    assert ds.relative_units == "SECONDS", "relative timestamps are only currently supported in seconds"
+                    absolute_end = start_dt + timedelta(seconds=ds[ds.ts_col].max())
+                    end_dt = max(end_dt, absolute_end)
+                else:
+                    absolute_end = ds["_python_datetime"].max()
+                    end_dt = max(end_dt, absolute_end)
```
alkasm, line 483: absolute_end here may be tz-aware, and for the first iteration end_dt is tz-naive, and thus cannot be compared.

we can do the naive conversion, but unclear if users expect the timestamps to be local or UTC

PR #11, opened 2024-08-28
Add new SDK

alkasm:
Adds a new developer-focused SDK for creating and manipulating runs, datasets, attachments. Keeps the existing client lib in-place, we can merge this without breaking the existing functionality. Then we'll create a new PR to remove and replace the existing library code with equivalent functionality.

- [x] strict mypy check all new code on 3.8, 3.9, 3.10, 3.11, 3.12
- [x] test on a local nominal instance
- [x] test on a staging cloud environment


nominal/sdk.py
```diff
@@ -0,0 +1,403 @@
...
+from ._api.combined import scout
+from ._api.combined import scout_catalog
+from ._api.combined import scout_run_api
+from ._api.ingest import upload_api
+
+IntegralNanosecondsUTC = int
+IntegralNanosecondsRelative = int
+
+
+class AbsoluteTimestamp:
```
alkasm, line 20: this might make more sense as a few functions rather than a method that just stores a single int.

pyproject.toml
alkasm, line 1: we probably need to add a py.typed marker?

nominal/sdk.py
```diff
@@ -0,0 +1,508 @@
...
+    title: str
+    description: str
+    properties: Mapping[str, str]
+    labels: Sequence[str]
+    start: IntegralNanosecondsUTC
+    end: IntegralNanosecondsUTC | None
+    _auth_header: str = field(repr=False)
+    _run_client: scout.RunService = field(repr=False)
+
+    def add_datasets(self, datasets: Mapping[str, str]) -> None:
```
alxhill, line 58: thoughts on this supporting `[str, Union[str, Dataset]]` instead to make it more intuitive to use with existing objects?

alkasm: good idea

alkasm: can do that for all the method args that take RIDs

nominal/sdk.py
```diff
@@ -0,0 +1,508 @@
...
+        dataset: BinaryIO,
+        name: str,
+        timestamp_column_name: str,
+        timestamp_column_type: TimestampColumnType,
+        file_extension: _AllowedFileExtensions = ".csv",
+        *,
+        description: str | None = None,
+        labels: Sequence[str] = (),
+        properties: Mapping[str, str] | None = None,
+    ) -> str:
```
alxhill, line 404: should this return a `Dataset` to align up with `create_run` which returns a `Run`?

nominal/sdk.py
```diff
@@ -0,0 +1,508 @@
...
+        self._run_client.update_run_attachment(self._auth_header, request, self.rid)
+
+    def replace(
+        self,
+        *,
+        title: str | None = None,
+        description: str | None = None,
+        properties: Mapping[str, str] | None = None,
+        labels: Sequence[str] | None = None,
+    ) -> Self:
```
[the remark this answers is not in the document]
alkasm: Yeah, I waffled on this myself. Since you're also bringing it up I think I'll mutate the instance. 

Initially I wrote all the attributes with a @property decorator so I could make them "immutable" to the user but still update them. But then I went to frozen dataclasses for simplicity. I can still mutate the members in the object using __setattr__ even if using a frozen dataclass.

nominal/sdk.py
```diff
@@ -0,0 +1,508 @@
...
+
+        Timestamp column types must be a `CustomTimestampFormat` or one of the following literals:
+            "iso_8601": ISO 8601 formatted strings,
+            "epoch_{unit}": epoch timestamps in UTC (floats or ints),
+            "relative_{unit}": relative timestamps (floats or ints),
+            where {unit} is one of: nanoseconds | microseconds | milliseconds | seconds | minutes | hours | days
+        """
+
+        if isinstance(dataset, TextIOBase):
+            raise TypeError(f"dataset {dataset} must be open in binary mode, rather than text mode")
```
alxhill, line 417: should we have a version of this method that just takes in a path & opens the file on behalf of the user? helps avoid awkwardness like this too

nominal/sdk.py
alkasm: Ya I wasn't sure if I should mirror the APIs directly (for runs, the APIs return a run, for datasets, the APIs return a rid) or if I should make additional API calls to make them behave the same.

pyproject.toml
```diff
@@ -1,10 +1,11 @@
 [tool.poetry]
 name = "nominal"
-version = "0.0.39"
+version = "1.0.0-beta"
 description = "Automate Nominal workflows in Python"
-authors = ["JP <jack@nominal.io>", "Michael Bauer <michael@nominal.io>"]
+authors = ["JP <jack@nominal.io>", "Michael Bauer <michael@nominal.io>", "Alexander Reynolds <alex.reynolds@nominal.io>"]
 readme = "README.md"
 packages = [
+    {include = "nominal/py.typed"},
```
alxhill, line 8: what does this file do btw? not familiar with mypy

nominal/sdk.py
```diff
@@ -0,0 +1,508 @@
...
+class Dataset:
+    rid: str
+    name: str
+    description: str | None
+    properties: Mapping[str, str]
+    labels: Sequence[str]
+    _auth_header: str = field(repr=False)
+    _catalog_client: scout_catalog.CatalogService = field(repr=False)
+
+    def poll_until_ingestion_completed(self, dataset_rid: str, interval: timedelta = timedelta(seconds=2)) -> None:
```
alkasm, line 155: Oops, fallout from moving this from a function to a method; this should just use `self.rid` rather than passing it in

pyproject.toml
alkasm: it marks that this package is typed for consumers; when they call this lib, mypy knows it contains types

nominal/sdk.py
alkasm: I intended to add that as a separate method in the future; will add it as a todo for now!

nominal/sdk.py
alkasm: should I rename them back to `update()` then? maybe more intuitive?

nominal/sdk.py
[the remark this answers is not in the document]
alkasm: should I rename them back to `update()` then? maybe more intuitive?

nominal/sdk.py
alkasm: added as a todo here and n create_attachment

nominal/sdk.py
alkasm: done. I also added it to `get_<x>()` methods, which is maybe a bit silly, e.g. `attachment = get_attachment(attachment)`, but figure I might as well keep the API uniform.

PR #12, opened 2024-09-03
v1 function-oriented library

alkasm:
This PR adds a function-oriented interface around the new SDK for simple and short high-level scripting. This is intended to replace the prior v0 client library.

### Migration guide

The initial v0 client library and the new v1 are architected differently, so there isn't a direct 1-to-1 translation. However, the v1 has the same capabilities and more. Some examples are given below for transitioning to the new v1. Please contact your friendly neighborhood Nomineer if you'd like some help translating to the new client library!

Everything you need is now under the main nominal namespace: `import nominal as nm`

### Set your token

**v0**
```py
nm.cloud.set_token("eyJ...")
```

**v1**
Instead of setting this token in Python, you set it at the command-line:
```sh
python3 -m nominal auth set-token
```

### Create and upload a dataset

The v0 library tried to guess the timestamp column, but you could also specify it manually. Different versions looked like:

**v0 (a)**
```py
dataset = Ingest().read("data.csv")
dataset = Ingest.set_ts_index(dataset)
dataset.upload()
```

**v0 (b)**
```py
dataset = Ingest().read("data.csv", ts_col="timestamp")
dataset.upload()
```

**v0 (c)**
```py
df = pl.read_csv("data.csv")
df = Ingest.set_ts_index(df, ts_col="timestamp")
dataset = Dataset(df, "data.csv", relative=True, relative_units="SECONDS")
dataset.upload()
```

The v1 library requires you to specify the timestamp column and type:

**v1**
```py
dataset = nm.upload_csv(
    "data.csv",
    "Dataset Name",
    timestamp_column="timestamp",
    timestamp_type="iso_8601",  # or "epoch_seconds", "relative_microseconds", etc
)
```

There are also `nm.upload_polars()` and `nm.upload_pandas()` functions.

### Get a dataset

**v0**
Not available

**v1**
```py
dataset = nm.get_dataset("ri.scout....")
```

### Get a run

The initial client library had slightly different functionality depending on how you created the `Run` object. You could get a run by its rid, _if_ you didn't specify files to upload:

**v0**
```py
run = Run(rid="ri.scout.run...")
```

**v1**
```py
run = nm.get_run("ri.scout.run...")
```

### Create a run

**v0**
```py
run = Run(path="data.csv")  # from a single CSV file
run = Run(paths=["data1.csv", "data2.csv"])  # from multiple CSV files
run = Run(datasets=[dataset_a, dataset_b])  # from multiple datasets that were already uploaded
```

**v1 (a)** - create a run without datasets (optionally add them later)
```py
run = nm.create_run("My Run", start="2024-09-05T12:35:00Z", end="2024-09-05T13:15:00Z")
run.add_dataset("test-stand-a", dataset_a)
run.add_dataset("test-stand-b", dataset_b)
```

**v1 (b)** - create a run and a dataset simultaneously
```py
run = nm.create_run_csv(
    "data.csv",
    "My Run",
    timestamp_column="timestamp",
    timestamp_type="iso_8601",
)
```

The v1 client library does not support creating a run with multiple CSV files simultaneously. Instead, use `nm.create_run()` with `run.add_dataset()` to add multiple datasets.

### Set the base URL

_Generally, users aren't expected to use this; it's more for Nomineers for local development or staging environments._

**v0**
```py
nm.cloud.set_base_url("PROD")
```

**v1**
```py
nm.set_base_url("https://api.gov.nominal.io/api")
```


nominal/__init__.py
```diff
@@ -1,13 +1,48 @@
...
-from .nominal import Ingest as Ingest
-from .nominal import Run as Run
-
-os.environ["NOMINAL_BASE_URL"] = "https://api-staging.gov.nominal.io/api"
+__all__ = [
+    "set_default_connection",
+    "get_default_connection",
+    "upload_dataset_from_pandas",
+    "upload_dataset_from_polars",
+    "upload_dataset",
```
alxhill, line 29: naming suggestions:
* `upload_polars`
* `upload_pandas`
* `upload_csv`
I broadly like the idea of the `upload_` methods being super specific, so if we e.g support new file types, they would all be new individual methods (`upload_parquet`, etc).

nominal/__init__.py
```diff
@@ -1,13 +1,48 @@
...
+    "get_dataset_by_rid",
```
alxhill, line 30: should this be a "high level" method? 🤔 

if so, I think it should be thought of as the "primary" way to get datasets - maybe `find_dataset` would be the name for like, search by name or channels or something else

nominal/__init__.py
```diff
@@ -1,13 +1,48 @@
...
+    "update_dataset",
```
alxhill, line 31: thoughts on making the function-based version of this more targeted? e.g `update_dataset_name` / etc for the fields we want people to think of as mutable

(also worth noting that there are a bunch of catalog endpoints that ideally wouldn't be public at all - such as updating bounds, ingest status, etc, to lets make sure they're excluded from both python apis)

nominal/__init__.py
```diff
@@ -1,13 +1,48 @@
...
+    "create_run",
+    "get_run_by_rid",
+    "search_runs",
+    "update_run",
+    "add_dataset_to_run",
```
alxhill, line 36: for the function api, should we go for a higher level abstraction? like, `create_run_csv` / `create_run_polars` that handles upload + create directly?

nominal/_utils.py
```diff
@@ -246,3 +139,47 @@ def update_dataclass(self: T, other: T, fields: Iterable[str]) -> None:
     """
     for field in fields:
         self.__dict__[field] = getattr(other, field)
+
+
+class FileType(NamedTuple):
```
alxhill, line 144: why `NamedTuple` vs frozen dataclass? (just curious, only loosely familiar with them)

nominal/nominal.py
```diff
@@ -1,628 +1,316 @@
...
-
-    Methods
-    -------
-    upload(overwrite=False)
-        Uploads and registers the dataset on the Nominal platform.
+
+_default_connection: NominalClient | None = None
+
+
+def set_default_connection(base_url: str, token: str) -> None:
```
alkasm, line 30: should I create separate functions for setting the base URL and token? should I do that in addition to, or instead of, set_default_connection()?

nominal/_utils.py
```diff
@@ -246,3 +139,47 @@ def update_dataclass(self: T, other: T, fields: Iterable[str]) -> None:
...
+    extension: str
+    mimetype: str
+
+    @classmethod
+    def from_path(cls, path: Path | str, default_mimetype: str = "application/octect-stream") -> FileType:
+        ext = "".join(Path(path).suffixes)
+        mimetype, _encoding = mimetypes.guess_type(path)
+        if mimetype is None:
+            return cls(ext, default_mimetype)
+        return cls(ext, mimetype)
```
alxhill, line 154: nit: would this work? `cls(ext, mimetype or default_mimetype)`

nominal/nominal.py
alkasm: in examples, it's a bit easier to copy/paste and edit one line at a time when they are separate functions. if you copy/paste a multiline thing into a repl, you can't really go back to edit it

nominal/nominal.py
```diff
@@ -1,628 +1,316 @@
...
+    global _default_connection
+    if _default_connection is None:
+        raise NominalError("No default connection set: initialize with `set_default_connection(base_url, token)`")
+    return _default_connection
+
+
+def upload_dataset_from_pandas(
+    df: pd.DataFrame,
+    name: str,
+    description: str,
```
alkasm, line 55: technically description is optional -- should I default it to None for this API for fewer required args, or should we push users into being good stewards by making it required?

nominal/nominal.py
```diff
@@ -1,628 +1,316 @@
...
+    TimestampColumnType,
+    _datetime_to_integral_nanoseconds,
+    reader_writer,
 )
-from ._utils import PayloadFactory, create_service, default_filename
-from .cloud import ENDPOINTS, _auth_help_blurb, get_app_base_url, get_base_url
 
+if TYPE_CHECKING:
+    import pandas as pd
+    import polars as pl
```
alxhill, line 24: interesting, when is this `TYPE_CHECKING` var defined? is this is mypy thing

nominal/nominal.py
alkasm: similarly, should `name` be required or should we feel free to randomize it if not provided?

nominal/nominal.py
```diff
@@ -1,628 +1,316 @@
...
+    Updates the current instance and returns it.
+
+    Note: This replaces the metadata (rather than adding to it).
+    """
+    return dataset.update(name=name, description=description, properties=properties, labels=labels)
+
+
+def create_run(
+    title: str,
+    description: str,
```
alkasm, line 173: run descriptions are not optional (as per the APIs) but we could similarly default to empty string.

I feel like run titles should always be mandatory/not autogenerated

alxhill (commented):
pushing this out as a checkpoint, will switch to comments vs a review given size

some of these comments might be answered later in the code, so feel free to ignore while I continue reviewing if there are obvious oversights

nominal/nominal.py
alxhill: I quite liked the `set_token` functionality of the previous api, and nice to have `keyring` secure it and retrieve it so it doesn't have to be manually included in every script. I think setting tokens should be separate from setting URIs (every customer has to set their token - most customers will not need to override the api URI)

my personal opinion is that the python client should assume the public API url, and only require configuration for on-prem uris (which we'll happen to use too for development)

nominal/nominal.py
```diff
@@ -1,628 +1,316 @@
...
+
+def set_default_connection(base_url: str, token: str) -> None:
+    """Set the default global connection to the Nominal platform.
+
+    base_url: The URL of the Nominal API platform, e.g. "https://api.gov.nominal.io/api".
+    token: An API token to authenticate with. You can grab a client token from the Nominal sandbox, e.g.
+        at https://app.gov.nominal.io/sandbox.
     """
+    global _default_connection
+    _default_connection = NominalClient.create(base_url, token)
```
alxhill, line 38: thoughts on `NominalClient.get()` or `.default()` instead of a global?

nominal/nominal.py
alxhill: name: required unless we can infer it from something (e.g do dataframes have names or keep track of the csv they came from?)
description: optional, if you make it required people just put empty string or "." - if it's not a valuable feature to them, I don't think we should try and force usage

nominal/__init__.py
alkasm: you're suggesting `get_dataset_by_rid()` -> `get_dataset()` since it's the main way (so the verbosity doesn't help disambiguate), correct?

nominal/__init__.py
alkasm: oh interesting, didn't think about that. we could do that, but might be kinda polluting the namespace? I'm not sure. users can always do each one individually:

```py
update_dataset(dataset, name="new name")
update_dataset(dataset, description="new description")
```

nominal/__init__.py
alkasm: I was shying away from this initially, primarily so that the function signatures remained small. I didn't love the idea of including disambiguating parameter names like `dataset_name`, `run_name`, etc so figured it was best to keep them separate. Also have a q here about usage that I'll ask you about directly

nominal/__init__.py
alxhill: so specifically this could be called `get_dataset` & ignore the "by_rid" bit if there's no other way to get a single dataset

nominal/_utils.py
alkasm: here I think it would be equivalent. By default I always explicitly check optionals with None checks since I usually want to allow for falsey values (e.g. empty string). That's not the case here but I try to maintain consistency throughout

nominal/_utils.py
alkasm: Ah good point. Generally I use them for the opposite reason I did here - to give structure to an API that you can otherwise pass a tuple, e.g. pass me a Point(x, y) so that I can shuttle it to an underlying API that only knows about tuples. But that's not the case here.

nominal/nominal.py
alkasm: > (e.g do dataframes have names or keep track of the csv they came from?)
nope

nominal/nominal.py
alxhill: agreed re: run titles

I think we should make run description optional. as of right now, there are ~32k runs and 20k of them have an empty string as their description. my hottake is that the backend api should reflect this & make the field optional too 👀

nominal/__init__.py
alxhill: summary from async chat: 
- maybe not necessary to have an update flow as part of the function-oriented api
- so long as fields like labels/properties can be set on creation, updates can be part of the object-oriented version (I.e `dataset.update(description="new description")` would be the way to do an update)

nominal/nominal.py
alkasm: > I quite liked the set_token functionality of the previous api, and nice to have keyring secure it and retrieve it so it doesn't have to be manually included in every script.

yeah agreed, was balancing this need vs relying on mutable system state which I don't really like. Biasing against someone running a script that "works on their machine" and then a coworker runs the same script and it doesn't work.

nominal/nominal.py
alkasm: made optional

nominal/nominal.py
alkasm: Yeah. If users aren't using pandas, they don't pay an import penalty for pandas/polars just because I included it in a type hint

nominal/_utils.py
alxhill: yeah don't mind the explicit version either, just curious - thx

nominal/nominal.py
alkasm: still need a global for the url or token either way (or data persisted to disk) so I don't think this is majorly different

nominal/nominal.py
alxhill: 🔥

nominal/__init__.py
alkasm: for uploading from a run - we can probably just generate the name for the dataset, and remove the descriptions. then the signature is small, and they have more customizability through composing functions. but this keeps the barrier to entry super low.

nominal/nominal.py
alkasm: counterpoint: biasing against folks checking in tokens

nominal/nominal.py
```diff
@@ -1,628 +1,316 @@
...
+        t.start()
+        dataset = conn.create_dataset_from_io(
+            reader,
+            name,
+            timestamp_column=timestamp_column,
+            timestamp_type=timestamp_type,
+            file_type=FileTypes.CSV,
+            description=description,
+        )
+        t.join()
```
alxhill, line 82: to confirm the `reader_writer` is here to proxy logs from the other thread to this main one?

also, wondering if we need the complexity of a separate thread given this method blocks on the upload and ingest anyway, so the main thread will always be waiting for the thread to complete

nominal/nominal.py
```diff
@@ -1,628 +1,316 @@
...
+    if wait_until_complete:
+        dataset.poll_until_ingestion_completed()
+    return dataset
+
+
+def upload_dataset(
```
alxhill, line 122: if we do want a singular method for file upload (vs `upload_csv` / `upload_parquet`), it should have `file` in the name for consistency with the dataframe methods

alxhill: (ignore if you've updated this is updated already locally!)

nominal/nominal.py
alkasm: the multithreading isn't for speed, but rather to avoid copying the dataframe to an in-memory string

nominal/nominal.py
```diff
@@ -1,628 +1,268 @@
...
+
+
+def search_runs(
+    *,
+    start: str | datetime | IntegralNanosecondsUTC | None = None,
+    end: str | datetime | IntegralNanosecondsUTC | None = None,
+    exact_title: str | None = None,
+    label: str | None = None,
+    property: tuple[str, str] | None = None,
+) -> list[Run]:
```
alxhill, line 183: what will running `search_runs` with no parameters return? good to avoid the case where users call the method accidentally and it suddenly blocks for ages trying to read every single run

nominal/sdk.py
```diff
@@ -5,6 +5,7 @@
 from dataclasses import dataclass, field
 from datetime import datetime, timedelta
 from io import TextIOBase
+from io import TextIOBase
```
alxhill, line 8:
```suggestion
from io import TextIOBase
```

nominal/nominal.py
alkasm: it actually throws, conjure doesn't allow it

nominal/__init__.py
alkasm: I'm going to add this, but I'll do that in a v1.1. Want to get this merged without much further new testing

nominal/nominal.py
alkasm: added keyring functionality back in

nominal/nominal.py
alkasm: (but this is also why I prefer the `-> Iterable` signatures!)

alkasm: I now throw if none are provided

nominal/nominal.py
```diff
@@ -1,628 +1,263 @@
...
+@cache
+def _get_or_create_connection(base_url: str, token: str) -> NominalClient:
+    return NominalClient.create(base_url, token)
 
-        Returns:
-        Response object from the REST call.
-        """
 
-        s3_upload_resp = self.__upload_file(overwrite)
+def set_base_url(base_url: str = _DEFAULT_BASE_URL) -> None:
```
alkasm, line 41: hmm now this is kinda wonky, `set_base_url` behaves differently than `set_token`.

I think I'm just going to roll in `~/.nominal.yml` now

alkasm: Okay, so now there's _only_ a `set_base_url()`. Setting a token should be done on the system / outside of the Python API to discourage checking in tokens. I included a CLI utility for creating/setting the config file.

```py
python -m nominal auth set-token
```

alkasm:
@shaun-sweet

nominal/nominal.py
alkasm: I will eventually build out a more fully featured rust/clap CLI for this and other functionality in the near future, but this works for now

alkasm:
Keeping the migration guide as the top-level comment. Here's the rest of the PR info:


- [x] top-level functions for most common ops (create, get, list, update on datasets, runs, attachments)
- [x] e2e test every top level function
- [x] use a custom type for (file type, mime type) pairs in SDK
- [x] consistent ordering of labels, props in SDK
- [x] remove ability to create runs with datasets simultaneously in SDK and top-level
- [x] change `run.add_datasets()` to `run.add_dataset()`
- [x] add `get_attachments()` and `list_attachments()` to SDK
- [x] bump minimum python version to 3.9 (some dependencies, like pandas, don't support 3.8 anymore)
- [x] overhaul README
- [x] write/read through top-level docstrings
- [x] map v0 functions to v1 functions
- [x] poll for completion on all dataset uploads on top level functions
- [x] use "name" over "title" for everything
- [x] remove data (should add back later with a focus on timeseries data)
- [x] add function for going straight from csv -> (dataset and run)

## Testing

### Unit tests

Most tests are e2e, there's one unit test though:

```sh
$ poetry run pytest
```

### End-to-end tests

Requires a local Nominal instance running. Set your auth token to $DEVTOKEN

```sh
$ poetry run pytest tests/e2e --auth-token $DEVTOKEN                 
```

### Static analysis

```sh
$ poetry run mypy nominal/nominal.py nominal/_utils.py nominal/sdk.py nominal/__init__.py
Success: no issues found in 4 source files
```

```sh
$ poetry run ruff format --check nominal
7 files already formatted
```

alkasm:
> This PR adds a function-oriented interface around the new SDK for simple and short high-level scripting. This is intended to replace the prior v0 client library.
> 
> ### Migration guide
> The initial v0 client library and the new v1 are architected differently, so there isn't a direct 1-to-1 translation. However, the v1 has the same capabilities and more. Some examples are given below for transitioning to the new v1. Please contact your friendly neighborhood Nomineer if you'd like some help translating to the new client library!
> 
> Everything you need is now under the main nominal namespace: `import nominal as nm`
> 
> ### Set your token
> **v0**
> 
> ```python
> nm.cloud.set_token("eyJ...")
> ```
> 
> **v1** Instead of setting this token in Python, you set it at the command-line:
> 
> ```shell
> python3 -m nominal auth set-token
> ```
> 
> ### Create and upload a dataset
> The v0 library tried to guess the timestamp column, but you could also specify it manually. Different versions looked like:
> 
> **v0 (a)**
> 
> ```python
> dataset = Ingest().read("data.csv")
> dataset = Ingest.set_ts_index(dataset)
> dataset.upload()
> ```
> 
> **v0 (b)**
> 
> ```python
> dataset = Ingest().read("data.csv", ts_col="timestamp")
> dataset.upload()
> ```
> 
> **v0 (c)**
> 
> ```python
> df = pl.read_csv("data.csv")
> df = Ingest.set_ts_index(df, ts_col="timestamp")
> dataset = Dataset(df, "data.csv", relative=True, relative_units="SECONDS")
> dataset.upload()
> ```
> 
> The v1 library requires you to specify the timestamp column and type:
> 
> **v1**
> 
> ```python
> dataset = nm.upload_csv(
>     "data.csv",
>     "Dataset Name",
>     timestamp_column="timestamp",
>     timestamp_type="iso_8601",  # or "epoch_seconds", "relative_microseconds", etc
> )
> ```
> 
> There are also `nm.upload_polars()` and `nm.upload_pandas()` functions.
> 
> ### Get a dataset
> **v0** Not available
> 
> **v1**
> 
> ```python
> dataset = nm.get_dataset("ri.scout....")
> ```
> 
> ### Get a run
> The initial client library had slightly different functionality depending on how you created the `Run` object. You could get a run by its rid, _if_ you didn't specify files to upload:
> 
> **v0**
> 
> ```python
> run = Run(rid="ri.scout.run...")
> ```
> 
> **v1**
> 
> ```python
> run = nm.get_run("ri.scout.run...")
> ```
> 
> ### Create a run
> **v0**
> 
> ```python
> run = Run(path="data.csv")  # from a single CSV file
> run = Run(paths=["data1.csv", "data2.csv"])  # from multiple CSV files
> run = Run(datasets=[dataset_a, dataset_b])  # from multiple datasets that were already uploaded
> ```
> 
> **v1 (a)** - create a run without datasets (optionally add them later)
> 
> ```python
> run = nm.create_run("My Run", start="2024-09-05T12:35:00Z", end="2024-09-05T13:15:00Z")
> run.add_dataset("test-stand-a", dataset_a)
> run.add_dataset("test-stand-b", dataset_b)
> ```
> 
> **v1 (b)** - create a run and a dataset simultaneously
> 
> ```python
> run = nm.create_run_csv(
>     "data.csv",
>     "My Run",
>     timestamp_column="timestamp",
>     timestamp_type="iso_8601",
> )
> ```
> 
> The v1 client library does not support creating a run with multiple CSV files simultaneously. Instead, use `nm.create_run()` with `run.add_dataset()` to add multiple datasets.
> 
> ### Set the base URL
> _Generally, users aren't expected to use this; it's more for Nomineers for local development or staging environments._
> 
> **v0**
> 
> ```python
> nm.cloud.set_base_url("PROD")
> ```
> 
> **v1**
> 
> ```python
> nm.set_base_url("https://api.gov.nominal.io/api")
> ```



> This PR adds a function-oriented interface around the new SDK for simple and short high-level scripting. This is intended to replace the prior v0 client library.
> 
> ### Migration guide
> The initial v0 client library and the new v1 are architected differently, so there isn't a direct 1-to-1 translation. However, the v1 has the same capabilities and more. Some examples are given below for transitioning to the new v1. Please contact your friendly neighborhood Nomineer if you'd like some help translating to the new client library!
> 
> Everything you need is now under the main nominal namespace: `import nominal as nm`
> 
> ### Set your token
> **v0**
> 
> ```python
> nm.cloud.set_token("eyJ...")
> ```
> 
> **v1** Instead of setting this token in Python, you set it at the command-line:
> 
> ```shell
> python3 -m nominal auth set-token
> ```
> 
> ### Create and upload a dataset
> The v0 library tried to guess the timestamp column, but you could also specify it manually. Different versions looked like:
> 
> **v0 (a)**
> 
> ```python
> dataset = Ingest().read("data.csv")
> dataset = Ingest.set_ts_index(dataset)
> dataset.upload()
> ```
> 
> **v0 (b)**
> 
> ```python
> dataset = Ingest().read("data.csv", ts_col="timestamp")
> dataset.upload()
> ```
> 
> **v0 (c)**
> 
> ```python
> df = pl.read_csv("data.csv")
> df = Ingest.set_ts_index(df, ts_col="timestamp")
> dataset = Dataset(df, "data.csv", relative=True, relative_units="SECONDS")
> dataset.upload()
> ```
> 
> The v1 library requires you to specify the timestamp column and type:
> 
> **v1**
> 
> ```python
> dataset = nm.upload_csv(
>     "data.csv",
>     "Dataset Name",
>     timestamp_column="timestamp",
>     timestamp_type="iso_8601",  # or "epoch_seconds", "relative_microseconds", etc
> )
> ```
> 
> There are also `nm.upload_polars()` and `nm.upload_pandas()` functions.
> 
> ### Get a dataset
> **v0** Not available
> 
> **v1**
> 
> ```python
> dataset = nm.get_dataset("ri.scout....")
> ```
> 
> ### Get a run
> The initial client library had slightly different functionality depending on how you created the `Run` object. You could get a run by its rid, _if_ you didn't specify files to upload:
> 
> **v0**
> 
> ```python
> run = Run(rid="ri.scout.run...")
> ```
> 
> **v1**
> 
> ```python
> run = nm.get_run("ri.scout.run...")
> ```
> 
> ### Create a run
> **v0**
> 
> ```python
> run = Run(path="data.csv")  # from a single CSV file
> run = Run(paths=["data1.csv", "data2.csv"])  # from multiple CSV files
> run = Run(datasets=[dataset_a, dataset_b])  # from multiple datasets that were already uploaded
> ```
> 
> **v1 (a)** - create a run without datasets (optionally add them later)
> 
> ```python
> run = nm.create_run("My Run", start="2024-09-05T12:35:00Z", end="2024-09-05T13:15:00Z")
> run.add_dataset("test-stand-a", dataset_a)
> run.add_dataset("test-stand-b", dataset_b)
> ```
> 
> **v1 (b)** - create a run and a dataset simultaneously
> 
> ```python
> run = nm.create_run_csv(
>     "data.csv",
>     "My Run",
>     timestamp_column="timestamp",
>     timestamp_type="iso_8601",
> )
> ```
> 
> The v1 client library does not support creating a run with multiple CSV files simultaneously. Instead, use `nm.create_run()` with `run.add_dataset()` to add multiple datasets.
> 
> ### Set the base URL
> _Generally, users aren't expected to use this; it's more for Nomineers for local development or staging environments._
> 
> **v0**
> 
> ```python
> nm.cloud.set_base_url("PROD")
> ```
> 
> **v1**
> 
> ```python
> nm.set_base_url("https://api.gov.nominal.io/api")
> ```

PR #13, opened 2024-09-04
[sdk] Add to multifile dataset

alkasm:
- update API definitions (moves ingest/ and upload/ folders into combined/)
- add method to append data to dataset
- pass the nominal client everywhere (rather than each individual service client). wanted to stay pure here by only injecting the necessary dependencies but it just makes the code a lot uglier; can refactor in the future with a better pattern

TODO:
- [x] test on local instance
- [x] test on staging


nominal/sdk.py
```diff
@@ -210,24 +203,51 @@ def update(
...
+            destination=ingest_api.IngestDestination(
+                existing_dataset=ingest_api.ExistingDatasetIngestDestination(dataset_rid=self.rid)
+            ),
+            source=ingest_api.IngestSource(s3=ingest_api.S3IngestSource(path=s3_path)),
+            source_metadata=ingest_api.IngestSourceMetadata(
+                timestamp_metadata=ingest_api.TimestampMetadata(
+                    series_name=timestamp_column_name,
+                    timestamp_type=_timestamp_type_to_conjure_ingest_api(timestamp_column_type),
+                ),
+            ),
```
alxhill, line 238: can we add some quick validation to throw if it's not an absolute timestamp?

alxhill (approved):
one nit, otherwise looks 🔥

nominal/sdk.py
alkasm: realized that i forgot that as soon as I asked you to take a look lol

nominal/sdk.py
```diff
@@ -433,22 +454,29 @@ def create_dataset_from_io(
...
-                series_name=timestamp_column_name,
-                timestamp_type=_timestamp_type_to_conjure_ingest_api(timestamp_column_type),
+            source_metadata=ingest_api.IngestSourceMetadata(
+                timestamp_metadata=ingest_api.TimestampMetadata(
+                    series_name=timestamp_column_name,
+                    timestamp_type=_timestamp_type_to_conjure_ingest_api(timestamp_column_type),
+                ),
             ),
         )
         response = self._ingest_client.trigger_ingest(self._auth_header, request)
```
alkasm, line 482: wait shouldn't this be `trigger_file_ingest`?

alkasm: how did this work??

nominal/sdk.py
```diff
@@ -210,24 +204,54 @@ def update(
...
+        self,
+        dataset: BinaryIO,
+        timestamp_column_name: str,
+        timestamp_column_type: TimestampColumnType,
+        file_extension: _AllowedFileExtensions = ".csv",
+    ) -> None:
+        """Append to a dataset from a file-like object."""
+
+        if timestamp_column_type.startswith("relative"):
+            raise ValueError("multifile datasets with relative timestamps are not yet supported by the client library")
```
alxhill, line 223: I think we should check that the dataset is in the success state _before_ we start uploading the file - otherwise will have to re-upload the file on failure as we don't cache it anywhere

alxhill (approved):
one note, otherwise lgtm

nominal/sdk.py
alkasm: oh I guess I jsut changed it. mypy caught this, along with some other issues that I just fixed.

nominal/sdk.py
alkasm: done! good catch, removed this while testing.

PR #18, opened 2024-09-09
feat: partially support log set service

ritwikdixit:
This PR will add support for a few endpoints within the Log Set Service.
- create
- attachLogsAndFinalize
- getLogSetMetadata
- searchLogs

### DIFF
--- a/nominal/_utils.py
+++ b/nominal/_utils.py
@@ -7,12 +7,12 @@ Internal utility functions for Nominal Python client
 import random
 import string
 from datetime import datetime
+from enum import Enum
 from typing import Type, TypeVar
 
 import requests
 from conjure_python_client import ServiceConfiguration
 from requests.utils import CaseInsensitiveDict
-from enum import Enum
 
 
 def default_filename(nominal_file_class):
@@ -116,11 +116,10 @@ class PayloadFactory:
 import logging
 import sys
 from dataclasses import dataclass
-from datetime import datetime, timezone, timedelta
-from typing import Literal, Union, Iterable
-from ._api.combined import ingest_api
-from ._api.combined import scout_run_api
-from ._api.combined import datasource
+from datetime import timedelta, timezone
+from typing import Iterable, Literal, Union
+
+from ._api.combined import datasource, ingest_api, scout_run_api
 
 if sys.version_info >= (3, 11):
     from typing import Self as Self
--- a/nominal/_utils.py
+++ b/nominal/_utils.py
@@ -12,6 +12,7 @@ from typing import Type, TypeVar
 import requests
 from conjure_python_client import ServiceConfiguration
 from requests.utils import CaseInsensitiveDict
+from enum import Enum
 
 
 def default_filename(nominal_file_class):
@@ -115,10 +116,11 @@ class PayloadFactory:
 import logging
 import sys
 from dataclasses import dataclass
-from datetime import datetime, timezone
+from datetime import datetime, timezone, timedelta
 from typing import Literal, Union, Iterable
 from ._api.combined import ingest_api
 from ._api.combined import scout_run_api
+from ._api.combined import datasource
 
 if sys.version_info >= (3, 11):
     from typing import Self as Self
@@ -164,6 +166,13 @@ TimestampColumnType: TypeAlias = Union[
     CustomTimestampFormat,
 ]
 
+DataSourceType: TypeAlias = Union[
+    Literal[
+        "dataset"
+        "log_set",
+    ],
+]
+
 
 def _timestamp_type_to_conjure_ingest_api(
     ts_type: TimestampColumnType,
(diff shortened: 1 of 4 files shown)


nominal/_utils.py
```diff
@@ -198,12 +207,20 @@ def _flexible_time_to_conjure_scout_run_api(
...
 def _conjure_time_to_integral_nanoseconds(
     ts: scout_run_api.UtcTimestamp,
 ) -> IntegralNanosecondsUTC:
     return ts.seconds_since_epoch * 1_000_000_000 + (ts.offset_nanoseconds or 0)
 
+def _datasource_api_timestamp_to_datetime(
+    ts:  datasource.Timestamp
+) -> datetime:
+    seconds, nanos = ts.seconds, ts.nanos
+    return datetime.fromtimestamp(seconds, tz=timezone.utc) + timedelta(microseconds=nanos // 1000)
```
ritwikdixit, line 223: `datetime` doesn't support nanoseconds, should I just use `IntegralNanosecondsUTC`?

ritwikdixit: `datetime` also doesn't make much sense with relative timestamps if there are already logs in relative time..

nominal/sdk.py
```diff
@@ -271,6 +279,85 @@ def _from_conjure(cls, client: NominalClient, dataset: scout_catalog.EnrichedDat
...
+    time: datetime
+    body: str
+    properties: Mapping[str, str] | None = None
+
+    def _to_conjure(self) -> datasource_logset_api.Log:
+        properties = {} if not self.properties else dict(self.properties)
+        return datasource_logset_api.Log(
+            time = _datetime_to_conjure_datasource_api(self.time),
+            body = datasource_logset_api.LogBody(
+                basic = datasource_logset_api.BasicLogBody(
```
ritwikdixit, line 335: `BasicLogBody` is the only type in the union right now

nominal/sdk.py
```diff
@@ -392,17 +483,28 @@ def create_run(
...
                 )
-                for ref_name, rid in datasets.items()
-            },
+                for ref_name, rid in data_sources.items()
+            } if data_sources else {}
+
+        combined_data_sources = { 
+            **create_run_data_sources_for_data_source_type(datasets, "dataset"),
+            **create_run_data_sources_for_data_source_type(log_sets, "log_set"),
+        }
```
ritwikdixit, line 503: Is there a more fluent way to do this? I thought that since we'd be supporting 4-5+ types I'd combine them somehow but the only type safety on providing arg name as a string is that it's of the literal `DataSourceType`

nominal/sdk.py
```diff
@@ -458,6 +560,59 @@ def search_runs(
...
+
+        Timestamp type must be a string equal to either 'ABSOLUTE' or 'RELATIVE'.
+        """
+        timestamp_type_enum = None
+        if timestamp_type == "ABSOLUTE":
+            timestamp_type_enum = DataSourceTimestampType.ABSOLUTE
+        elif timestamp_type_enum == "RELATIVE":
+            timestamp_type_enum = DataSourceTimestampType.RELATIVE
+        else:
+            raise TypeError(f"timestamp type {timestamp_type} must be one of [RELATIVE, ABSOLUTE]")
```
ritwikdixit, line 582: I also don't love taking in a string to do this check for abs vs relative but I didn't know what to do off the top of my head. In the API there's a type `UNKNOWN` though I thought to just throw for now.

nominal/sdk.py
```diff
@@ -458,6 +560,59 @@ def search_runs(
...
+            next_page_token: str | None = None,
+        ):
+        request = datasource_logset_api.SearchLogsRequest(
+            token = next_page_token,
+            page_size = page_size,
+        )
+        response = self._logset_client.search_logs(
+            self._auth_header,
+            log_set_rid = log_set_rid,
+            request = request,
```
ritwikdixit, line 611: So for pagination I didn't do anything here it largely mirrors the Conjure API. I can go take a look at the pattern you provided for runs though!

nominal/sdk.py
```diff
@@ -458,6 +560,59 @@ def search_runs(
...
+        )
+        response = self._logset_client.create(self._auth_header, request)
+        return LogSetMetadata._from_conjure(self, response)
+
+    def get_log_set_metadata(self, log_set_rid) -> LogSetMetadata:
+        """Returns metadata about a log set given its RID."""
+        response = self._logset_client.get_log_set_metadata(self._auth_header, log_set_rid)
+        return LogSetMetadata._from_conjure(self, response)
+
+    def search_logs(
```
ritwikdixit, line 598: what's the convention on what to put under the Client vs under the LogSetMetadata object itself?

nominal/sdk.py
```diff
@@ -458,6 +583,43 @@ def search_runs(
...
+
+        Timestamp type must be a string equal to either 'ABSOLUTE' or 'RELATIVE'.
+        """
+        timestamp_type_enum = None
+        if timestamp_type == "ABSOLUTE":
+            timestamp_type_enum = DataSourceTimestampType.ABSOLUTE
+        elif timestamp_type_enum == "RELATIVE":
+            timestamp_type_enum = DataSourceTimestampType.RELATIVE
+        else:
+            raise TypeError(f"timestamp type {timestamp_type} must be one of [RELATIVE, ABSOLUTE]")
```
ritwikdixit, line 605: ehh I don't think I'm doing the right thing here by taking in a string and mapping to an enum defined in `utils`.. open to suggestions

nominal/sdk.py
```diff
@@ -271,6 +279,108 @@ def _from_conjure(cls, client: NominalClient, dataset: scout_catalog.EnrichedDat
             _client=client,
         )
 
+@dataclass(frozen=True)
+class LogSet:
```
ritwikdixit, line 283: I called it `LogSet` because I added methods inside to do stuff with it. I figure that's better than keeping with the conjure definition of `LogSetMetadata`, and then adding attach logs / search logs to the client itself

nominal/sdk.py
```diff
@@ -271,6 +279,108 @@ def _from_conjure(cls, client: NominalClient, dataset: scout_catalog.EnrichedDat
...
+        """
+        request = datasource_logset_api.SearchLogsRequest(
+            token = next_page_token,
+            page_size = page_size,
+        )
+        response = self._client._logset_client.search_logs(
+            self._client._auth_header,
+            log_set_rid = self.rid,
+            request = request,
+        )
```
ritwikdixit, line 328: For now the pagination functionality just mirrors conjure, AKA not very intuitive .. - I saw you were using yield for Runs and so maybe I should emulate that?

nominal/sdk.py
```diff
@@ -382,6 +497,7 @@ def create_run(
         end: datetime | IntegralNanosecondsUTC,
         *,
         datasets: Mapping[str, str] | None = None,
+        log_sets: Mapping[str, str] | None = None,
```
alkasm, line 500: I removed the dataset argument in the new functional API PR, just makes the SDK simpler to use and combining this API call isn't really solving an important headache (i.e. it's okay to do these operations step-by-step)

PR #38, opened 2024-09-13
feat: add csv to dataset method

alxhill:
simple helper to avoid pitfall of needing to open in binary mode

### DIFF
--- a/nominal/core.py
+++ b/nominal/core.py
@@ -230,6 +230,11 @@ class Dataset:
         update_dataclass(self, dataset, fields=self.__dataclass_fields__)
         return self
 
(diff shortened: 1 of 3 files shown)


alkasm (approved):

nominal/core.py
```diff
@@ -230,6 +230,11 @@ def update(
         update_dataclass(self, dataset, fields=self.__dataclass_fields__)
         return self
 
+    def add_csv_to_dataset(self, path: str | Path, timestamp_column: str, timestamp_type: TimestampColumnType) -> None:
```
alkasm, line 233: we should add a similar method to `NominalClient` for the initial upload as well

PR #41, opened 2024-09-13
fix: detect .csv.gz and add helper method for initial csv creation

alkasm:
Building off the comments in #38

### DIFF
--- a/nominal/core.py
+++ b/nominal/core.py
@@ -230,10 +230,11 @@ class Dataset:
         update_dataclass(self, dataset, fields=self.__dataclass_fields__)
         return self
 
-    def add_csv_to_dataset(self, path: str | Path, timestamp_column: str, timestamp_type: TimestampColumnType) -> None:
+    def add_csv_to_dataset(self, path: Path | str, timestamp_column: str, timestamp_type: TimestampColumnType) -> None:
         """Append to a dataset from a csv on-disk."""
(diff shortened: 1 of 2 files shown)


nominal/nominal.py
```diff
@@ -143,26 +146,20 @@ def upload_csv(
...
+    name: str | None,
     timestamp_column: str,
     timestamp_type: TimestampColumnType,
     description: str | None = None,
     *,
     wait_until_complete: bool = True,
 ) -> Dataset:
-    path = Path(file)
-    file_type = FileType.from_path_dataset(path)
-    if file_type.extension not in (".csv", "csv.gz"):
```
alkasm, line 155: this was an error before (should be `.csv.gz` rather than `csv.gz`)

nominal/core.py
```diff
@@ -671,3 +704,11 @@ def _create_search_runs_query(
         q = scout_run_api.SearchQuery(property=scout_run_api.Property(name=name, value=value))
         queries.append(q)
     return scout_run_api.SearchQuery(and_=queries)
+
+
+def _verify_csv_path(path: Path | str) -> tuple[Path, FileType]:
+    path = Path(path)
+    file_type = FileType.from_path_dataset(path)
+    if file_type.extension not in (".csv", ".csv.gz"):
```
alkasm, line 712: this is the hotfix

alxhill (approved):

PR #42, opened 2024-09-16
feat: add video support

alkasm:
Add:

- `Video`
- `Video.poll_until_ingestion_completed()`
- `Video.update()`
- `get_video()`
- `upload_video()`


nominal/_utils.py
```diff
@@ -7,10 +7,12 @@
...
 from datetime import datetime, timezone
 from pathlib import Path
-from typing import BinaryIO, Iterable, Iterator, Literal, NamedTuple, TypeVar, Union
+from typing import BinaryIO, Iterable, Iterator, Literal, NamedTuple, Type, TypeVar, Union
 
 import dateutil.parser
-from typing_extensions import TypeAlias  # typing.TypeAlias in 3.10+
+from typing_extensions import TypeAlias
+
+from nominal._api.combined import scout_video_api  # typing.TypeAlias in 3.10+
```
alkasm, line 15:
```suggestion
from typing_extensions import TypeAlias # typing.TypeAlias in 3.10+
```

not quite sure how the comment moved...but also this second import is unused

nominal/core.py
```diff
@@ -364,6 +371,90 @@ def _from_conjure(
...
 
 
+@dataclass(frozen=True)
+class Video:
+    rid: str
+    name: str
+    description: str | None
+    properties: Mapping[str, str]
+    labels: Sequence[str]
+    _client: NominalClient
```
alkasm, line 381: repr=False

nominal/core.py
```diff
@@ -364,6 +371,90 @@ def _from_conjure(
...
+
+            new_labels = ["new-label-a", "new-label-b"]
+            for old_label in video.labels:
+                new_labels.append(old_label)
+            video = video.update(labels=new_labels)
+        """
+        # TODO(alkasm): properties SHOULD be optional here, but they're not.
+        # For uniformity with other methods, will always "update" with current props on the client.
+        request = scout_video_api.UpdateVideoMetadataRequest(
+            description=description,
```
ritwikdixit, line 435: Yeah we can maybe block on this becoming optional

nominal/_utils.py
```diff
@@ -169,6 +175,7 @@ class FileTypes:
     CSV_GZ: FileType = FileType(".csv.gz", "text/csv")
     # https://issues.apache.org/jira/browse/PARQUET-1889
     PARQUET: FileType = FileType(".parquet", "application/vnd.apache.parquet")
+    MP4: FileType = FileType(".mp4", "video/mp4")
```
ritwikdixit, line 178: Is it worth making MOV and/or AVI container types a first-class FileType?

nominal/nominal.py
```diff
@@ -292,6 +292,23 @@ def download_attachment(rid: str, file: Path | str) -> None:
...
 
+def upload_video(
+    file: Path | str, name: str, start: datetime | str | IntegralNanosecondsUTC, description: str | None = None
+) -> Video:
+    """Upload a video to Nominal from a file."""
+    conn = get_default_client()
+    path = Path(file)
+    file_type = FileType.from_path(path)
+    with open(file, "rb") as f:
+        return conn.create_video_from_io(f, name, _parse_timestamp(start), description, file_type)
```
ritwikdixit, line 303: I think `mimetypes.guess_type(path)` should work for other types of video containers aside from mp4 but maybe worth a test? We don't really enforce anything for container type on the backend (though maybe we should lol)

nominal/core.py
```diff
@@ -536,6 +630,59 @@ def create_dataset_from_io(
...
+        filename = f"{urlsafe_name}{file_type.extension}"
+
+        s3_path = put_multipart_upload(self._auth_header, video, filename, file_type.mimetype, self._upload_client)
+        request = ingest_api.IngestVideoRequest(
+            labels=list(labels),
+            properties={} if properties is None else dict(properties),
+            sources=[ingest_api.IngestSource(s3=ingest_api.S3IngestSource(path=s3_path))],
+            timestamps=ingest_api.VideoTimestampManifest(
+                no_manifest=ingest_api.NoTimestampManifest(
+                    starting_timestamp=_flexible_time_to_conjure_ingest_api(start)
```
ritwikdixit, line 662: There's this concept of a "scale parameter" which we mainly did for Moria but it allows people to specify if videos are sped up / slowed down. It's worth including if anyone else will use it

ritwikdixit: But honestly I'm not sure if anyone else will use it lol

nominal/_utils.py
alkasm: This enum is actually just for default/well-known values. When using the functions that actually take a path, it's parsed from the path.

nominal/nominal.py
alkasm: yes, it does!

nominal/core.py
alkasm: I don't think we have to block, since we can just send it the current props

PR #47, opened 2024-09-17
feat: add bulk add datasource method to run

ross-barrett:


nominal/core.py
```diff
@@ -67,6 +67,14 @@ class Run:
     def add_dataset(self, ref_name: str, dataset: Dataset | str) -> None:
         """Add a dataset to this run.
 
+        Datasets map "ref names" (their name within the run) to a Dataset (or dataset rid). The same type of datasets
+        should use the same ref name across runs, since checklists and templates use ref names to reference datasets.
+        """
+        self.add_datasets({ref_name: dataset})
+
+    def add_datasets(self, datasets: dict[str, Dataset | str]) -> None:
```
alkasm, line 75:
```suggestion
    def add_datasets(self, datasets: Mapping[str, Dataset | str]) -> None:
```

`Mapping` is more generic/allows subclasses of dicts/etc

alkasm (approved):
ty!

PR #50, opened 2024-09-18
feat: add support for expression grammar checklists

ritwikdixit:
Checks as code, checks from YAML

Refactor of https://github.com/nominal-io/nominal-client/compare/main...rb/checklist-expressions?expand=1 from @ross-barrett


nominal/_utils.py
```diff
@@ -48,6 +48,20 @@ class CustomTimestampFormat:
     CustomTimestampFormat,
 ]
 
+Priority = Literal["P0", "P1", "P2", "P3", "P4"]
```
ritwikdixit, line 51: How's using a literal for typing?

nominal/core.py
```diff
@@ -311,6 +322,146 @@ def _from_conjure(cls, client: NominalClient, dataset: scout_catalog.EnrichedDat
         )
 
 
+# TODO(ritwikdixit): add support for more Checklist metadata and versioning
```
ritwikdixit, line 325: There's a few things left out here. Namely, commit # and checklist metadata - for now, we're not really letting users do anything with the checklist other than create it.

nominal/core.py
```diff
@@ -755,6 +917,94 @@ def _search_datasets(self) -> Iterable[Dataset]:
...
+        )
+        response = self._authentication_client.search_users_v2(
+            self._auth_header,
+            request
+        )
+        if len(response.results) == 0:
+            raise ValueError(f"User {user_email} not found")
+        if len(response.results) > 1:
+            raise ValueError(f"Found multiple users with email {user_email}")
+        return response.results[0].rid
```
ritwikdixit, line 935: eventually we may want to change the checklists api to take in a user email but I think we're just gonna do this to unblock for now

nominal/core.py
```diff
@@ -311,6 +322,174 @@ def _from_conjure(cls, client: NominalClient, dataset: scout_catalog.EnrichedDat
...
+
+    # we do this afterward to batch the request
+    def _resolve_expressions(self) -> None:
+        variable_expressions = self._client._compute_representation_client.batch_compute_to_expression(
+            self._client._auth_header,
+            {var.name: var.compute_graph for var in self.variables}
+        )
+        check_expressions = self._client._compute_representation_client.batch_compute_to_expression(
+            self._client._auth_header,
+            {check.rid : check.compute_graph for check in self.checks}
```
ritwikdixit, line 345: we send out 2 network requests here, could be done in one, but I think it's fine

nominal/core.py
```diff
@@ -751,6 +941,96 @@ def _search_datasets(self) -> Iterable[Dataset]:
...
+            functions={},
+            properties={} if properties is None else dict(properties),
+            labels=list(labels),
+            checks=conjure_checks,
+            checklist_variables=conjure_checklist_variables,
+            is_published=is_published,
+        )
+
+        response = self._checklist_api_client.create(self._auth_header, request)
+        # TODO(ritwikdixit): remove the need for an extraneous network request
```
ritwikdixit, line 998: Not ideal for now - we technically send a request to the representation service to resolve the expression representations of the checklist variables and the checks, when we had them from the start.. but it's not super easy to match them up between the pre / post request checks, variables

nominal/_utils.py
```diff
@@ -48,6 +48,20 @@ class CustomTimestampFormat:
...
+        return scout_checks_api.Priority.P0
+    elif priority == "P1":
+        return scout_checks_api.Priority.P1
+    elif priority == "P2":
+        return scout_checks_api.Priority.P2
+    elif priority == "P3":
+        return scout_checks_api.Priority.P3
+    elif priority == "P4":
+        return scout_checks_api.Priority.P4
+
```
alkasm, line 64: Should throw here for unexpected values

nominal/nominal.py
```diff
@@ -339,3 +348,41 @@ def _get_start_end_timestamp_csv_file(
         IntegralNanosecondsUTC(start.to_datetime64().astype(int)),
         IntegralNanosecondsUTC(end.to_datetime64().astype(int)),
     )
+
+def create_checklist(
+    assignee_email: str,
```
alkasm, line 353: preference to put this right before the args with defaults so that we can eventually turn this optional - i don't think this should need to be required in the future

nominal/nominal.py
```diff
@@ -339,3 +348,41 @@ def _get_start_end_timestamp_csv_file(
...
+    title: str,
+    checks: Sequence[CreateCheck],
+    checklist_variables: Sequence[CreateChecklistVariable],
+    default_ref_name: str | None = None,
+    commit_message: str | None = None,
+    description: str | None = None,
+    properties: Mapping[str, str] | None = None,
+    labels: Sequence[str] = (),
+    is_published: bool = True,
```
alkasm, line 362: what is is_published?

nominal/nominal.py
```diff
@@ -339,3 +348,41 @@ def _get_start_end_timestamp_csv_file(
         IntegralNanosecondsUTC(start.to_datetime64().astype(int)),
         IntegralNanosecondsUTC(end.to_datetime64().astype(int)),
     )
+
+def create_checklist(
+    assignee_email: str,
+    title: str,
+    checks: Sequence[CreateCheck],
+    checklist_variables: Sequence[CreateChecklistVariable],
+    default_ref_name: str | None = None,
```
alkasm, line 357: what is a default_ref_name?

nominal/nominal.py
```diff
@@ -339,3 +348,41 @@ def _get_start_end_timestamp_csv_file(
         IntegralNanosecondsUTC(start.to_datetime64().astype(int)),
         IntegralNanosecondsUTC(end.to_datetime64().astype(int)),
     )
+
+def create_checklist(
+    assignee_email: str,
+    title: str,
```
alkasm, line 354: everything in the client library is uniformly using "name" even if the api uses "title"

nominal/nominal.py
```diff
@@ -339,3 +348,41 @@ def _get_start_end_timestamp_csv_file(
...
+    checks: Sequence[CreateCheck],
+    checklist_variables: Sequence[CreateChecklistVariable],
+    default_ref_name: str | None = None,
+    commit_message: str | None = None,
+    description: str | None = None,
+    properties: Mapping[str, str] | None = None,
+    labels: Sequence[str] = (),
```
alkasm, line 361: this function has a ton of params which will make it more cumbersome to use and document; unless we need everything here I'd prefer to make the signature smaller, and provide the .update() method to add props and labels like we do with the other functions

nominal/core.py
```diff
@@ -751,6 +941,96 @@ def _search_datasets(self) -> Iterable[Dataset]:
...
+        with open(filePath, "r") as file:
+            checklist_dict = yaml.load(file, Loader=yaml.FullLoader)
+            data = checklist_dict["checklist"]
+            return self.create_checklist(
+                title=data["title"],
+                assignee_email=data["assignee_email"],
+                checklist_variables=[checklist_variable_constructor(var) for var in data["variables"]],
+                checks=[check_constructor(chk) for chk in data["checks"]],
+
+                description=data.get("description", None),
```
alkasm, line 1023: `None` is the default second argument for `get`, so you can just `get("key")`

nominal/core.py
```diff
@@ -751,6 +941,96 @@ def _search_datasets(self) -> Iterable[Dataset]:
...
+            is_published=is_published,
+        )
+
+        response = self._checklist_api_client.create(self._auth_header, request)
+        # TODO(ritwikdixit): remove the need for an extraneous network request
+        return Checklist._from_conjure(self, response)
+        
+
+    def create_checklist_from_yaml(self, filePath: str) -> Checklist:
+        def remove_newlines(s: str) -> str:
```
alkasm, line 1003: these functions should be utility functions rather than inlined.

nominal/core.py
```diff
@@ -751,6 +941,96 @@ def _search_datasets(self) -> Iterable[Dataset]:
...
+        def check_constructor(data: Mapping[str, str]) -> CreateCheck:
+            data["expression"] = remove_newlines(data["expression"])
+            return CreateCheck(**data)
+
+        def checklist_variable_constructor(data: Mapping[str, str]) -> CreateChecklistVariable:
+            data["expression"] = remove_newlines(data["expression"])
+            return CreateChecklistVariable(**data)
+
+        with open(filePath, "r") as file:
+            checklist_dict = yaml.load(file, Loader=yaml.FullLoader)
```
alkasm, line 1015: use `yaml.safe_load(f)` - we do not need the full loader.

ref:
- https://github.com/yaml/pyyaml/wiki/PyYAML-yaml.load(input)-Deprecation
- https://nvd.nist.gov/vuln/detail/CVE-2017-18342

nominal/core.py
```diff
@@ -751,6 +941,96 @@ def _search_datasets(self) -> Iterable[Dataset]:
...
+            self._auth_header,
+            request
+        )
+        if len(response.results) == 0:
+            raise ValueError(f"User {user_email} not found")
+        if len(response.results) > 1:
+            raise ValueError(f"Found multiple users with email {user_email}")
+        return response.results[0].rid
+
+    def create_checklist(
```
alkasm, line 961: this should be a `classmethod` not an instance method

nominal/core.py
```diff
@@ -751,6 +941,96 @@ def _search_datasets(self) -> Iterable[Dataset]:
...
+            checklist_variables=conjure_checklist_variables,
+            is_published=is_published,
+        )
+
+        response = self._checklist_api_client.create(self._auth_header, request)
+        # TODO(ritwikdixit): remove the need for an extraneous network request
+        return Checklist._from_conjure(self, response)
+        
+
+    def create_checklist_from_yaml(self, filePath: str) -> Checklist:
```
alkasm, line 1002: I would probably keep this functionality in the toplevel namespace for now

alkasm (changes requested):
marking this as needs changes specifically to track `yaml.load` -> `yaml.safe_load`

nominal/core.py
```diff
@@ -311,6 +322,174 @@ def _from_conjure(cls, client: NominalClient, dataset: scout_catalog.EnrichedDat
...
+
+class CreateCheck(TypedDict):
+    title: str
+    priority: Priority
+    description: str | None
+    expression: str
+
+
+# TODO(ritwikdixit): add support for more fields i.e. lineage
+@dataclass(frozen=False)
```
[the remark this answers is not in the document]
ritwikdixit: Yeah I did a kinda weird thing to be able to send the batch request but I think there's a way of circumventing that

alkasm:
do you have an example of invoking this handy?

nominal/_utils.py
ritwikdixit: I get a `Pylance error` that the code is unreachable in VSCode but I'll just try keeping it for now

nominal/nominal.py
ritwikdixit: agreed

nominal/nominal.py
ritwikdixit: In our backend checklist version control, you can save a workbook as a draft, or publish it, analogous to committing and merging to main (I think).

nominal/nominal.py
ritwikdixit: Because the compute representation service does not fetch any actual series on the backend, we refer to series by their channel name and ref name - similar to unresolved variable locators in the checklist service. When the checklist is applied to a run, they end up getting resolved into actual series.

nominal/core.py
ritwikdixit: why would `create_run` and `create_dataset_from_io` be instance methods but this be a class method?

nominal/nominal.py
ritwikdixit: Discussed offline: switching to a builder

nominal/core.py
alkasm: because i can't read

sorry! false alarm

nominal/nominal.py
alkasm: I guess the fact that it's called "default" is what's tripping me up? it's not really a default right?

nominal/nominal.py
alkasm: got it so you could potentially `get_checklist()` which might return a draft?

alkasm: potentially we should defer touching drafts at all, and filter out drafts--unless folks think otherwise

nominal/nominal.py
ritwikdixit: It's provided out of utility here - i.e. use this ref name when you go to create the unresolved checklist variable so that you don't have to specify the ref name every time you reference a channel

nominal/nominal.py
ritwikdixit: I think that's a good idea. I'm going to just throw if they try to get a checklist that isn't published

nominal/nominal.py
```diff
@@ -339,3 +348,33 @@ def _get_start_end_timestamp_csv_file(
...
+    description: str | None = None,
+    default_ref_name: str | None = None,
+) -> ChecklistBuilder:
+    builder = ChecklistBuilder(
+        name=name,
+        assignee_email=assignee_email,
+        description=description,
+        default_ref_name=default_ref_name,
+    )
+    builder._client = get_default_client()
```
ritwikdixit, line 365: pydantic would not actually end up setting this field if I tried to set the client in the constructor

nominal/nominal.py
```diff
@@ -339,3 +348,33 @@ def _get_start_end_timestamp_csv_file(
         IntegralNanosecondsUTC(start.to_datetime64().astype(int)),
         IntegralNanosecondsUTC(end.to_datetime64().astype(int)),
     )
+
+
+def draft_checklist(
```
ritwikdixit, line 353: Open to other naming options .. hoping to not overload the term "draft", which is used on the Front-end of the app, but I'm not actually sure how that maps to the concept of branches/VCS on the backend - we don't use the term draft in the APIs

nominal/core.py
```diff
@@ -863,3 +1122,118 @@ def _verify_csv_path(path: Path | str) -> tuple[Path, FileType]:
...
+) -> Mapping[str, scout_checks_api.UnresolvedNumRangesConditionV3]:
+    
+    response_dict = client.batch_expression_to_compute(
+        auth_header, scout_compute_representation_api.BatchExpressionToComputeRequest(
+            expressions=expressions,
+            default_ref_name=default_ref_name
+        )
+    )
+
+    def _get_compute_condition_for_compiled_node(compiledNode: scout_compute_representation_api.CompiledNode):
```
ritwikdixit, line 1141: Is inlining helper functions frowned upon?

nominal/core.py
```diff
@@ -802,6 +1032,35 @@ def get_attachments(self, rids: Iterable[str]) -> Sequence[Attachment]:
...
+
+        if "variables" in data:
+            for variable_dict in data["variables"]:
+                variable_dict["expression"] = _remove_newlines(variable_dict["expression"])
+                checklist_builder.add_checklist_variable(**variable_dict)
+
+        if "checks" in data:
+            for check_dict in data["checks"]:
+                check_dict["expression"] = _remove_newlines(check_dict["expression"])
+                checklist_builder.add_check(**check_dict)
```
ritwikdixit, line 1059: I just inlined this and went straight from dict into the args of the method - the CreateVar / CreateCheck objects themselves extend the pydantic object

nominal/core.py
```diff
@@ -311,6 +328,188 @@ def _from_conjure(cls, client: NominalClient, dataset: scout_catalog.EnrichedDat
...
+        # # TODO(ritwikdixit): remove the need for these extraneous network requests
+        variable_names_to_expressions = client._compute_representation_client.batch_compute_to_expression(
+            client._auth_header,
+            variable_name_to_graph_map
+        )
+        check_rids_to_expressions = client._compute_representation_client.batch_compute_to_expression(
+            client._auth_header,
+            {check_rid : graph for check_rid, (_, graph) in check_rid_to_graph_and_def_map.items()}
+        )
+        check_rids_to_definitions = {check_rid: check_def for check_rid, (check_def, _) in check_rid_to_graph_and_def_map.items()}
```
ritwikdixit, line 465: I don't love the style here, but the goals were:

1. Batch the requests for converting between compute graph to expression in such a way that I can always keep track of the graph, expression, and metadata for a rid

2. Remove the concept of the graph from Check and Variable signature for the user

ritwikdixit: Let me know if I should change the dict values being tuples as well

nominal/core.py
```diff
@@ -311,6 +328,188 @@ def _from_conjure(cls, client: NominalClient, dataset: scout_catalog.EnrichedDat
...
+    name: str
+    assignee_email: str
+    checklist_variables: Sequence[CreateChecklistVariable] = Field(default_factory=list)
+    checks: Sequence[CreateCheck] = Field(default_factory=list)
+    default_ref_name: str | None = None
+    commit_message: str | None = None
+    description: str | None = None
+    properties: Mapping[str, str] | None = None
+    labels: Sequence[str] = Field(default_factory=list)
+    _client: NominalClient = PrivateAttr()
```
ritwikdixit, line 355: I had to change the signature quite a bit to make sure pydantic wasn't angry - namely with defaults for lists/dicts and the client not being serializable/deep-copy-able

nominal/core.py
```diff
@@ -311,6 +328,188 @@ def _from_conjure(cls, client: NominalClient, dataset: scout_catalog.EnrichedDat
...
+    checklist_variables: Sequence[ChecklistVariable]
+    checks: Sequence[Check]
+    _client: NominalClient = field(repr=False)
+
+    @staticmethod
+    def _from_conjure(client: NominalClient, checklist: scout_checks_api.VersionedChecklist) -> Checklist:
+
+        # TODO(ritwikdixit): support draft checklists with VCS
+        if not checklist.metadata.is_published:
+            raise ValueError("cannot get a checklist that has not been published")
```
ritwikdixit, line 447: throw for checklists that aren't published for now, good call

ritwikdixit:
Examples of invoking this:

```
# get a checklist
checklist = nm.get_checklist(checklist_rid)

# build a checklist
checklist_builder = nm.draft_checklist(
    name="Python Client Checklist",
    assignee_email="ritwik@nominal.io",
    description="Python client checklist builder",
    default_ref_name="Reduced_"
)
checklist_builder.add_checklist_variable(
    name="b",
    expression="derivative(numericChannel(channelName = 'Cycle_Time'))"
)
checklist_builder.add_check(
    name="derivative of cycle time is too high",
    priority="P2",
    expression="numericSeriesRef('b') > 0.05"
)
finished_checklist = checklist_builder.build_and_publish()
```

ritwikdixit:
```
# From yaml
checklist_draft = nm.draft_checklist_from_yaml(yaml_path)
checklist = checklist_draft.build_and_publish()
```

example yaml file:
```
checklist:
  name: "Python Client YAML Checklist"
  description: "Basic checklist"
  assignee_email: "ritwik@nominal.io"
  default_ref_name: "Reduced_"
  variables:
    - name: "a"
      expression: |
        rolling(
            numericChannel(channelName = 'Temperature_1', refName = 'Reduced_'),
            AVERAGE,
            Duration(min = 2, s = 30)
        )
    - name: "b"
      expression: |
        derivative(
            numericChannel(channelName = 'Cycle_Time')
        )
  checks:
    - name: "Power consumption change or battery low"
      priority: "P0"
      description: "Change in power consumption or battery low"
      expression: |
        onChange(
            numericChannel(channelName = 'Power_Consumption', refName = 'Reduced_')
        ) or
        numericChannel(channelName = 'Battery_Level', refName = 'Reduced_') < 70
    - name: "Absolute value of the derivative of the cycle time limit"
      priority: "P1"
      expression: "numericSeriesRef('b') < 0.05"
    - name: "Rolling average of temperature limit"
      priority: "P3"
      expression: "numericSeriesRef('a') > 40"

```

nominal/core.py
```diff
@@ -976,6 +1225,122 @@ def _verify_csv_path(path: Path | str) -> tuple[Path, FileType]:
...
+    )
+
+    def _get_compute_condition_for_compiled_node(
+        compiledNode: scout_compute_representation_api.CompiledNode,
+    ) -> scout_checks_api.UnresolvedCheckCondition:
+        if compiledNode.node.type == "rangeSeries" and compiledNode.node.range_series is not None:
+            return scout_checks_api.UnresolvedCheckCondition(
+                num_ranges_v3=scout_checks_api.UnresolvedNumRangesConditionV3(
+                    function_spec={},
+                    operator=scout_compute_api.ThresholdOperator.GREATER_THAN,
```
alkasm, line 1248: is it correct that this just assigns an operator?

alkasm: same with the 0 value under this line

alkasm:
closing out in favor of: https://github.com/nominal-io/nominal-client/pull/61

PR #52, opened 2024-09-20
test: .csv.gz ingest

alkasm:


markgrex (approved):
Approving but ideally we'd check the series from the file were created and have data

alkasm:
> Approving but ideally we'd check the series from the file were created and have data

agreed! this is blocked on downloading series data - but that will come soon :)
this at least tests that it didn't fail ingesting. doesn't validate the ingestion tho.

tests/e2e/conftest.py
```diff
@@ -47,6 +48,7 @@ def csv_data():
 """
 
 
+@pytest.fixture(scope="session")
 def csv_data2():
```
alkasm, line 52: unclear to me why it didn't matter that this wasn't marked as a fixture

PR #53, opened 2024-09-21
feat: support log sets

ritwikdixit:
Porting functionality from https://github.com/nominal-io/nominal-client/pull/18 to v1 client

Supports
- Creating a log set
- Adding logs to set and finalizing
- Getting that set, and logs within that set in a paginated manner
- Adding log set to a run


nominal/_utils.py
```diff
@@ -49,6 +49,19 @@ class CustomTimestampFormat:
...
+    seconds, nanos = _flexible_time_to_seconds_nanos(timestamp)
+    return api.Timestamp(seconds=seconds, nanos=nanos)
+
+
+def _global_conjure_api_to_integral_nanoseconds(
+    timestamp: api.Timestamp,
+) -> IntegralNanosecondsUTC:
+    return timestamp.seconds * 1_000_000_000 + timestamp.nanos
+
+
```
ritwikdixit, line 64: Probably a bit of consolidation I can do with the other helper methods here

nominal/core.py
```diff
@@ -72,6 +80,26 @@ def add_dataset(self, ref_name: str, dataset: Dataset | str) -> None:
         """
         self.add_datasets({ref_name: dataset})
 
+    def add_log_set(self, ref_name: str, log_set: LogSet | str) -> None:
+        """
+        Add a log set to this run.
+        """
+        self.add_log_sets({ref_name: log_set})
+
+    def add_log_sets(self, log_sets: Mapping[str, LogSet | str]) -> None:
```
ritwikdixit, line 89: Separated into its own method rather than trying to combine with datasets

nominal/core.py
```diff
@@ -311,6 +339,132 @@ def _from_conjure(cls, client: NominalClient, dataset: scout_catalog.EnrichedDat
...
+            self._client._auth_header,
+            log_set_rid=self.rid,
+            request=datasource_logset_api.SearchLogsRequest(page_size=page_size),
+        )
+        return PaginatedLogs(
+            logs=[Log._from_conjure(log) for log in response.logs],
+            page_size=page_size,
+            next_page_token=response.next_page_token,
+            client=self._client,
+        )
```
ritwikdixit, line 387: Don't love the workflow here, of like `get_log_set().get_logs_paginated().get_logs()`

nominal/core.py
```diff
@@ -311,6 +339,132 @@ def _from_conjure(cls, client: NominalClient, dataset: scout_catalog.EnrichedDat
...
+    name: str
+    created_at: datetime
+    updated_at: datetime
+    log_count: int
+    timestamp_type: TimestampType
+    origin_metadata: Mapping[str, str]
+    description: str | None
+    _client: NominalClient = field(repr=False)
+
+    def add_logs_and_finalize(self, logs: list[Log]) -> Self:
```
ritwikdixit, line 355: there's maybe an opportunity to have a builder pattern here where the user can keep adding logs before we send out a network request and finalize the log set, but I'm not sure whether it's worth the effort given that it might get ripped out. Key API weakness here is log sets are immutable once you attach a set of logs once

nominal/core.py
```diff
@@ -311,6 +339,132 @@ def _from_conjure(cls, client: NominalClient, dataset: scout_catalog.EnrichedDat
...
+            self._client._auth_header,
+            log_set_rid=self.rid,
+            request=request,
+        )
+        return PaginatedLogs(
+            logs=[Log._from_conjure(log) for log in response.logs],
+            page_size=self.page_size,
+            next_page_token=response.next_page_token,
+            client=self._client,
+        )
```
ritwikdixit, line 465: don't love that the user would have to call `load_next_page()` and then `get_logs()`, open to ideas of how to deal with pagination cleanly.

nominal/core.py
```diff
@@ -709,6 +866,35 @@ def create_video_from_io(
...
+        not searchable until the logset is finalized.
+        Timestamp type must be a string equal to either 'ABSOLUTE' or 'RELATIVE'.
+        """
+        timestamp_type_enum = None
+        if timestamp_type == "absolute":
+            timestamp_type_enum = datasource.TimestampType.ABSOLUTE
+        elif timestamp_type_enum == "relative":
+            timestamp_type_enum = datasource.TimestampType.RELATIVE
+        else:
+            raise TypeError(f"timestamp type {timestamp_type} must be one of [RELATIVE, ABSOLUTE]")
```
ritwikdixit, line 887: I just inlined this for now

nominal/core.py
alkasm: should document what the key to the map is

nominal/core.py
```diff
@@ -72,6 +80,26 @@ def add_dataset(self, ref_name: str, dataset: Dataset | str) -> None:
...
+        """
+        self.add_log_sets({ref_name: log_set})
+
+    def add_log_sets(self, log_sets: Mapping[str, LogSet | str]) -> None:
+        """
+        Add multiple log sets to this run.
+        """
+        data_sources = {
+            ref_name: scout_run_api.CreateRunDataSource(
+                data_source=scout_run_api.DataSource(log_set == _rid_from_instance_or_string(log_set)),
```
alkasm, line 95: double equal

nominal/core.py
```diff
@@ -311,6 +339,132 @@ def _from_conjure(cls, client: NominalClient, dataset: scout_catalog.EnrichedDat
         )
 
 
+@dataclass(frozen=True)
+class LogSet:
+    rid: str
+    created_by: str
+    name: str
+    created_at: datetime
+    updated_at: datetime
```
alkasm, line 348: since the backend uses nanosecond precision, you'll see that all the types here (annoyingly) have to use IntegralNanosecondsUTC (i.e., int) when it's coming from the backend rather than a nice datetime

nominal/core.py
alkasm: allow for more flexible / less restrictive types on inputs. We don't need a `list` specifically, just something we can iterate through (i.e. an iterable)

alkasm: also agree that a builder may make sense for this, for the same reasons it makes sense for checklists

nominal/core.py
```diff
@@ -311,6 +339,132 @@ def _from_conjure(cls, client: NominalClient, dataset: scout_catalog.EnrichedDat
...
+            description=log_set_metadata.description,
+            _client=client,
+        )
+
+
+@dataclass(frozen=True)
+class Log:
+    time: datetime | IntegralNanosecondsUTC
+    body: str
+    properties: Mapping[str, str] | None = None
```
alkasm, line 417: I don't think this needs to be optional nor have a default value

nominal/core.py
```diff
@@ -311,6 +339,132 @@ def _from_conjure(cls, client: NominalClient, dataset: scout_catalog.EnrichedDat
...
+            timestamp_type=timestamp_type,
+            origin_metadata=MappingProxyType(log_set_metadata.origin_metadata),
+            description=log_set_metadata.description,
+            _client=client,
+        )
+
+
+@dataclass(frozen=True)
+class Log:
+    time: datetime | IntegralNanosecondsUTC
```
alkasm, line 415: lets standardize on "timestamp" as the name of every time field for the client lib

nominal/core.py
alkasm, line 887: suggestion: if you can ever get into the situation where you have to initialize a variable because branches may risk it not being defined, break it out into a function. This ensures you'll never end up in a branch with an undefined variable and makes type checking easy and explicit. E.g.:

```py
def convert_timestamp(t: Literal["absolute", "relative"]) -> datasource.TimestampType:
    if t == "absolute":
        return datasource.TimestampType.ABSOLUTE
    if t == "relative":
        return = datasource.TimestampType.RELATIVE
    raise ValueError(f"timestamp type {t} must be one of [RELATIVE, ABSOLUTE]")
```

now the caller will _always_ have a value of type `datasource.TimestampType`.

alkasm: also unrelated nit, but ValueError is more semantic here - the value is of the correct type (str), just an incorrect value.

ritwikdixit:
Note to self to include usage example

nominal/core.py
```diff
@@ -311,6 +339,68 @@ def _from_conjure(cls, client: NominalClient, dataset: scout_catalog.EnrichedDat
...
+            yield from response.logs
+            if response.next_page_token is None:
+                break
+            request = datasource_logset_api.SearchLogsRequest(page_token=response.next_page_token)
+
+    def list_logs(self) -> Iterable[Log]:
+        """
+        Get logs within this log set.
+        """
+        # TODO: different name?
```
alkasm, line 367: maybe `stream_logs()` is most semantic? Or maybe we just define an `__iter__()` so you can `for log in logset`? (or both)?

nominal/core.py
alkasm: is adding multiple log sets a common workflow?

nominal/core.py
ritwikdixit: resolved during pairing, consolidating into one method with create.

nominal/core.py
ritwikdixit: good catch

nominal/core.py
ritwikdixit: I like doing both!

alkasm: lets start with `stream_logs()` and if it becomes a common workflow, we can `__iter__` it up

ritwikdixit: Ah ok I added iter but I'll remove

nominal/core.py
ritwikdixit: renamed to timestamp

nominal/core.py
```diff
@@ -311,6 +343,69 @@ def _from_conjure(cls, client: NominalClient, dataset: scout_catalog.EnrichedDat
...
+            time=_flexible_time_to_global_conjure_api(self.timestamp),
+            body=datasource_logset_api.LogBody(
+                basic=datasource_logset_api.BasicLogBody(message=self.body, properties={}),
+            ),
+        )
+
+    @classmethod
+    def _from_conjure(cls, log: datasource_logset_api.Log) -> Self:
+        if log.body.basic is None:
+            raise ValueError("log body is empty")
```
alkasm, line 402: this is only coming from the backend - users can't change anything here, so it shouldn't be a value error. If this is a real possibility, the type should reflect it with `body: str | None`. If this shouldn't be possible then I think we should raise a `NominalError` or something to indicate that something is actually wrong with the backend, or the client-lib's understanding of the backend, if that makes sense.

nominal/core.py
```diff
@@ -709,6 +807,35 @@ def create_video_from_io(
         response = self._ingest_client.ingest_video(self._auth_header, request)
         return self.get_video(response.video_rid)
 
+    def create_log_set(
+        self,
+        name: str,
+        logs: Iterable[Log] | Iterable[tuple[datetime | IntegralNanosecondsUTC, str]],
+        timestamp_type: LogTimestampType = "absolute",  # TODO: ideate on this a bit
```
alkasm, line 814: we can remove this TODO. I think the default seems fine?

nominal/core.py
```diff
@@ -820,7 +954,13 @@ def _get_dataset(
...
+def _get_log_set(
+    auth_header: str, client: datasource_logset.LogSetService, log_set_rid: str
+) -> datasource_logset_api.LogSetMetadata:
+    return client.get_log_set_metadata(auth_header, log_set_rid)
+
+
+def _rid_from_instance_or_string(value: Attachment | Run | Dataset | Video | LogSet | str) -> str:
     if isinstance(value, str):
         return value
     elif isinstance(value, (Attachment, Run, Dataset, Video)):
```
alkasm, line 966: this should also have LogSet added

nominal/__init__.py
alkasm, line 1: we should export the Log and LogSet types here as well (and hide them in the docs as the comment states)

nominal/core.py
ritwikdixit: This is one of those weird situations where in Conjure we define a union type (Body) with just one option (Basic), and the generated Python stubs end up being of type Optional[Basic] - I'll change body to str | None

ritwikdixit:
Examples:

```
# these are just tuples that look like (timestamp, log message)
logs = [
    (datetime.now(timezone.utc) - timedelta(seconds=5-i), f"Log message {i}")
    for i in range(5)
]
client = nm.get_default_client()
log_set = client.create_log_set(
    name="log testing", 
    logs=logs, 
    description="description"
)

# add to run
run_rid = ...
run.add_log_set("log tests", log_set)


# Can get a log set by rid
rid = ...
new_log_set = nm.get_log_set(rid)

# view logs in a log set
for new_log in new_log_set.stream_logs():
    print(new_log)

```

nominal/core.py
```diff
@@ -8,7 +8,7 @@
 from io import TextIOBase
 from pathlib import Path
 from types import MappingProxyType
-from typing import BinaryIO, Iterable, Mapping, Sequence, cast
+from typing import BinaryIO, Iterable, Iterator, Mapping, Sequence, cast
```
alkasm, line 11: Iterator is unused

nominal/core.py
```diff
@@ -311,6 +343,66 @@ def _from_conjure(cls, client: NominalClient, dataset: scout_catalog.EnrichedDat
...
+            time=_flexible_time_to_global_conjure_api(self.timestamp),
+            body=datasource_logset_api.LogBody(
+                basic=datasource_logset_api.BasicLogBody(message=self.body, properties={}),
+            ),
+        )
+
+    @classmethod
+    def _from_conjure(cls, log: datasource_logset_api.Log) -> Self:
+        if log.body.basic is None:
+            raise RuntimeError("log body is empty")
```
alkasm, line 402: error message should be descriptive on why this is actually an error

nominal/core.py
```diff
@@ -709,6 +804,35 @@ def create_video_from_io(
...
+        not searchable until the logset is finalized.
+        Timestamp type must be a string equal to either 'absolute' or 'relative'.
+        """
+        request = datasource_logset_api.CreateLogSetRequest(
+            name=name,
+            description=description,
+            origin_metadata={},
+            timestamp_type=_log_timestamp_type_to_conjure(timestamp_type),
+        )
+        # we need to create the logset, then attach logs and finalize it
```
alkasm, line 825: i know i wrote this comment, but kinda useless given that the next two lines are "create the log set, then attach and finalize"

nominal/core.py
```diff
@@ -728,6 +852,13 @@ def get_dataset(self, rid: str) -> Dataset:
         response = _get_dataset(self._auth_header, self._catalog_client, rid)
         return Dataset._from_conjure(self, response)
 
+    def get_log_set(self, log_set_rid: str) -> LogSet:
+        """
+        Retrieve a LogSet along with its metadata given its RID.
```
alkasm, line 857: style nit: all of the docstrings in this repo follow the convention:
```py
def f():
    """The first line of the docstrings are on the same line as the quotes.
    The closing quotes follow on an empty line if the docstring is multiline.
    """
```

nominal/core.py
```diff
@@ -863,3 +1000,30 @@ def _verify_csv_path(path: Path | str) -> tuple[Path, FileType]:
...
+        return "absolute"
+    elif log_timestamp_type == datasource.TimestampType.RELATIVE:
+        return "relative"
+    raise ValueError(f"unhandled timestamp type {log_timestamp_type}")
+
+
+def _logs_to_conjure(
+    logs: Iterable[Log] | Iterable[tuple[datetime | IntegralNanosecondsUTC, str]],
+) -> Iterable[datasource_logset_api.Log]:
+    for log in logs:
```
alkasm, line 1024: since logs might be high frequency...may be worth it to map over the iterable depending on the type (rather than checking each element)

nominal/core.py
```diff
@@ -709,6 +804,35 @@ def create_video_from_io(
...
 
+    def create_log_set(
+        self,
+        name: str,
+        logs: Iterable[Log] | Iterable[tuple[datetime | IntegralNanosecondsUTC, str]],
+        timestamp_type: LogTimestampType = "absolute",
+        description: str | None = None,
+    ) -> LogSet:
+        """
+        Creates a log set, to which logs can be attached using `attach-and-finalize`. The logs within a logset are
```
alkasm, line 815: nit: use imperative mood / command language for docstrings:

> The docstring is a phrase ending in a period. It prescribes the function or method’s effect as a command (“Do this”, “Return that”), not as a description; e.g. don’t write “Returns the pathname …”.


https://peps.python.org/pep-0257/

alkasm: also this docstring is now inaccurate

ritwikdixit: updating

nominal/core.py
alkasm: meh, its too ugly to worry about for now lol

alkasm (approved):

nominal/__init__.py
```diff
@@ -39,4 +41,6 @@
     "Attachment",
     "NominalClient",
     "Video",
+    "LogSet",
+    "Log",
```
alkasm, line 45: need to add these to the filter in docs/reference/toplevel.md

alkasm (changes requested):

tests/e2e/test_toplevel.py
```diff
@@ -253,3 +253,28 @@ def test_get_video(mp4_data):
...
+
+def test_get_log_set():
+    name = f"logset-{uuid4()}"
+    desc = f"top-level test to create & get a log set {uuid4()}"
+    base_time = datetime.now(timezone.utc)
+    logs = [(base_time - timedelta(seconds=i), f"Log message {i}") for i in range(5)]
+
+    client = nm.get_default_client()
+
+    logset = client.create_log_set(name, logs, "absolute", desc)
```
alkasm, line 266: hmm this is not getting patched for me, and is trying to hit prod..

alkasm (changes requested):

tests/e2e/test_toplevel.py
alkasm: I fixed this by having the client be a fixture, and provided the client to the default connection patch, as well as an arg to this test

alkasm (approved):
There is a `get_log_set()` function but not a top-level function to create them yet. We would like to get a feel for what type of log files people are generally uploading so that we can create utilities for that. Unlike `upload_csv()`, `upload_polars()`, etc--there's less of an obvious standard for us to upload from a known text format. Could potentially see something like `upload_journald()` or similar.

nominal/__init__.py
ritwikdixit: ah sorry, I thought this _was_ the filter

PR #54, opened 2024-09-23
feat: cleanup & simplify timestamp handling

alkasm:
I went through a few different iterations to find a specification that worked well. What I ended up with wasn't too far from where we started, so I opted for backcompat with warnings for changes to how we're handling relative timestamps.

## `_AnyTimeDomain`: all time domains that we allow
This is value that should be used in the high-level client. It allows for basic literals + strongly typed values.

### `_LiteralAbsolute` values
These are the values we expect many users will use - simple, literal strings.
```py
"iso_8601"
"epoch_nanoseconds"
"epoch_microseconds"
"epoch_milliseconds"
"epoch_seconds"
"epoch_minutes"
"epoch_hours"
```

### `TypedTimeDomain` values
These are the strongly typed time domains. These include additional parameterized types, like `Relative` and `Custom`.
```py
nm.ts.Iso8601()
nm.ts.Epoch("nanoseconds")
nm.ts.Epoch("microseconds")
nm.ts.Epoch("milliseconds")
nm.ts.Epoch("seconds")
nm.ts.Epoch("minutes")
nm.ts.Epoch("hours")
nm.ts.Relative("nanoseconds", start=datetime(2024, 9, 23, 18, 12, 0))
nm.ts.Relative("microseconds", start=datetime(2024, 9, 23, 18, 12, 0))
nm.ts.Relative("milliseconds", start=datetime(2024, 9, 23, 18, 12, 0))
nm.ts.Relative("seconds", start=datetime(2024, 9, 23, 18, 12, 0))
nm.ts.Relative("minutes", start=datetime(2024, 9, 23, 18, 12, 0))
nm.ts.Relative("hours", start=datetime(2024, 9, 23, 18, 12, 0))
nm.ts.Custom(r"yyyy-MM-dd[T]hh:mm:ss")
nm.ts.Custom(r"MM-dd[T]hh:mm:ss", default_year=2024)
```

### `TypedTimeDomain` constants
These are equivalent to the string representations.
```py
nm.ts.ISO_8601
nm.ts.EPOCH_NANOSECONDS
nm.ts.EPOCH_MICROSECONDS
nm.ts.EPOCH_MILLISECONDS
nm.ts.EPOCH_SECONDS
nm.ts.EPOCH_MINUTES
nm.ts.EPOCH_HOURS
```

### `_LiteralRelativeDeprecated` values
Usage of these values is deprecated and will send a `UserWarning` to use `nm.ts.Relative()`. These are deprecated since we want the users to explicitly add the start time. However these are currently in-use, so we're emitting a warning but still allowing them for now.
```py
"relative_nanoseconds"
"relative_microseconds"
"relative_milliseconds"
"relative_seconds"
"relative_minutes"
"relative_hours"
```

The warning raised will look like:
```py
/Users/alex/nominal-io/nominal-client/nominal/timedomain.py:87: UserWarning: specifying 'relative_{unit}' as a string is deprecated and will be removed in a future version: use `nm.timedomain.Relative` instead. for example: instead of 'relative_seconds', use `nm.timedomain.Relative('seconds', start=datetime.now())`. 
```

### winner: mix of strings + types, but advertise with constants
```py
dataset = upload_csv("path/to/file.csv", "dataset", "timestamp", nm.timedomain.ISO_8601)
dataset = upload_csv("path/to/file.csv", "dataset", "timestamp", nm.timedomain.ABSOLUTE_NANOSECONDS)
dataset = upload_csv("path/to/file.csv", "dataset", "timestamp", nm.timedomain.Relative("nanoseconds", offset=15))
dataset = upload_csv("path/to/file.csv", "dataset", "timestamp", nm.timedomain.Custom(r"yyyy-MM-dd[T]hh:mm:ss"))
```

### current: mix of strings + types: relative offsets not supported yet, would need to create a new type for it
```py
dataset = upload_csv("path/to/file.csv", "dataset", "timestamp", "iso_8601")
dataset = upload_csv("path/to/file.csv", "dataset", "timestamp", "epoch_nanoseconds")
dataset = upload_csv("path/to/file.csv", "dataset", "timestamp", "relative_nanoseconds")
dataset = upload_csv("path/to/file.csv", "dataset", "timestamp", nm.CustomTimestampFormat(r"yyyy-MM-dd[T]hh:mm:ss"))
dataset = upload_csv("path/to/file.csv", "dataset", "timestamp", nm.RelativeTimestampFormat("nanoseconds", offset=15))
```

### strongly typed: types and docstrings are extremely clear, allows more flexibility per-type
```py
dataset = upload_csv("path/to/file.csv", "dataset", "timestamp", nm.timedomain.Iso8601())
dataset = upload_csv("path/to/file.csv", "dataset", "timestamp", nm.timedomain.Absolute("ns"))
dataset = upload_csv("path/to/file.csv", "dataset", "timestamp", nm.timedomain.Relative("nanoseconds", offset=15))
dataset = upload_csv("path/to/file.csv", "dataset", "timestamp", nm.timedomain.Custom(r"yyyy-MM-dd[T]hh:mm:ss"))
```

### flexible factory function: multiple behaviors depending on how you call it; difficult to document but easy to read
```py
dataset = upload_csv("path/to/file.csv", "dataset", "timestamp", nm.timedomain("iso_8601"))
dataset = upload_csv("path/to/file.csv", "dataset", "timestamp", nm.timedomain("absolute", "ns"))
dataset = upload_csv("path/to/file.csv", "dataset", "timestamp", nm.timedomain("relative", "ns", offset=15))
dataset = upload_csv("path/to/file.csv", "dataset", "timestamp", nm.timedomain("custom", r"yyyy-MM-dd[T]hh:mm:ss"))
```

### mix of strings + types, but advertise with singletons
```py
dataset = upload_csv("path/to/file.csv", "dataset", "timestamp", nm.timedomain.iso_8601)
dataset = upload_csv("path/to/file.csv", "dataset", "timestamp", nm.timedomain.absolute_nanoseconds)
dataset = upload_csv("path/to/file.csv", "dataset", "timestamp", nm.timedomain.Relative("nanoseconds", offset=15))
dataset = upload_csv("path/to/file.csv", "dataset", "timestamp", nm.timedomain.Custom(r"yyyy-MM-dd[T]hh:mm:ss"))
```

### factory functions for each type - pretty chill but wordier than strings
```py
dataset = upload_csv("path/to/file.csv", "dataset", "timestamp", nm.timedomain.iso_8601())
dataset = upload_csv("path/to/file.csv", "dataset", "timestamp", nm.timedomain.absolute_nanoseconds())
dataset = upload_csv("path/to/file.csv", "dataset", "timestamp", nm.timedomain.relative_nanoseconds(offset=15))
dataset = upload_csv("path/to/file.csv", "dataset", "timestamp", nm.timedomain.custom(r"yyyy-MM-dd[T]hh:mm:ss"))
```

### singletons (with transforming class methods) - too magic, custom() behaves differently
```py
dataset = upload_csv("path/to/file.csv", "dataset", "timestamp", nm.timedomain.iso_8601)
dataset = upload_csv("path/to/file.csv", "dataset", "timestamp", nm.timedomain.absolute_nanoseconds)
dataset = upload_csv("path/to/file.csv", "dataset", "timestamp", nm.timedomain.relative_nanoseconds)
dataset = upload_csv("path/to/file.csv", "dataset", "timestamp", nm.timedomain.relative_nanoseconds.offset(15))
dataset = upload_csv("path/to/file.csv", "dataset", "timestamp", nm.timedomain.custom(r"yyyy-MM-dd[T]hh:mm:ss"))
```


nominal/_timeutils.py
alkasm, line 1: more than half of _utils.py was for timestamps, so split out into a specific util file

nominal/core.py
alkasm, line 1: In general I opted for the methods in core to use the strongly typed values (since this is a programmer-focused SDK)...but I don't feel too strongly about it, so we can keep the literals (e.g. "epoch_timestamps") if we think that's best?

nominal/timedomain.py
```diff
@@ -0,0 +1,151 @@
+""" """
```
alkasm, line 1: oops

nominal/timedomain.py
alkasm, line 1: reminder: export the timestamp types

nominal/core.py
```diff
@@ -624,7 +612,7 @@ def create_dataset_from_io(
         dataset: BinaryIO,
         name: str,
         timestamp_column: str,
-        timestamp_type: TimestampColumnType,
+        timestamp_type: TypedTimeDomain,
```
alkasm, line 615: the docs are now out of date. Also, need to describe the functionality here in multiple places.

nominal/timedomain.py
```diff
@@ -0,0 +1,151 @@
...
+    "relative_microseconds",
+    "relative_milliseconds",
+    "relative_seconds",
+    "relative_minutes",
+    "relative_hours",
+]
+
+TypedTimeDomain: TypeAlias = Iso8601 | Epoch | Relative | Custom
+_AbsoluteTimeDomain: TypeAlias = Iso8601 | Epoch | Custom | _LiteralAbsolute
+_AnyTimeDomain: TypeAlias = TypedTimeDomain | _LiteralAbsolute | _LiteralRelativeDeprecated
```
alkasm, line 79: Python 3.9 doesn't like the mixture of types and literals with a `|` - need to explicitly use `typing.Union`

see https://github.com/python/mypy/issues/11665 - this was already an issue previously in the client lib, just in a different spot.

nominal/timedomain.py
```diff
@@ -0,0 +1,149 @@
...
+        return ingest_api.TimestampType(absolute=ingest_api.AbsoluteTimestamp(custom_format=fmt))
+    if isinstance(domain, Relative):
+        relative = ingest_api.RelativeTimestamp(
+            time_unit=_time_unit_to_conjure(domain.unit), offset=_flexible_to_iso8601(domain.start)
+        )
+        return ingest_api.TimestampType(relative=relative)
+    raise TypeError(f"invalid time domain type: {type(domain)}")
+
+
+def _flexible_to_iso8601(ts: datetime | IntegralNanosecondsUTC) -> str:
```
alkasm, line 119: probably belongs in _timeutils

alkasm:
I liked `timedomain` for the module name, but it's long / feels verbose. Can't use `time` since it's a built-in, and `domain` feels too generic. `chrono` or `clock` could work, but a bit loaded due to cpp. @alxhill suggested `ts` which is probably sufficient, so going with that.

alkasm:
Revamped all of the internal conversion functions to pass through an internal 'universal' type (_SecondsNanos) and moved all the timestamp conversion logic from _utils.py into ts.py

nominal/core.py
alkasm: I think we should keep the flexibility. It's not any more difficult to support.

nominal/core.py
```diff
@@ -250,7 +237,7 @@ def update(
         update_dataclass(self, dataset, fields=self.__dataclass_fields__)
         return self
 
-    def add_csv_to_dataset(self, path: Path | str, timestamp_column: str, timestamp_type: TimestampColumnType) -> None:
+    def add_csv_to_dataset(self, path: Path | str, timestamp_column: str, timestamp_type: TypedTimeDomain) -> None:
```
alxhill, line 240: do we still want the Domain naming here / will this be exposed to users? would lean towards something api-like (e.g TimestampType) for user-facing functions for consistency

nominal/ts.py
```diff
@@ -0,0 +1,238 @@
...
+    if domain not in _str_to_type:
+        raise ValueError(f"string time domains must be one of: {_str_to_type.keys()}")
+    return _str_to_type[domain]
+
+
+def _time_unit_to_conjure(unit: _LiteralTimeUnit) -> ingest_api.TimeUnit:
+    return ingest_api.TimeUnit[unit.upper()]
+
+
+_str_to_type: Mapping[_LiteralAbsolute | _LiteralRelativeDeprecated, Iso8601 | Epoch | Relative] = MappingProxyType(
```
alxhill, line 171: didn't know this existed, cool

nominal/ts.py
```diff
@@ -0,0 +1,238 @@
...
+        "relative_microseconds": Relative("microseconds", start=0),
+        "relative_milliseconds": Relative("milliseconds", start=0),
+        "relative_seconds": Relative("seconds", start=0),
+        "relative_minutes": Relative("minutes", start=0),
+        "relative_hours": Relative("hours", start=0),
+    }
+)
+
+
+class _SecondsNanos(NamedTuple):
```
alxhill, line 190: naming nit: _SecondsNanos is a v confusing name... would `_TimestampNanos` be fine? kinda doesn't matter how it's stored right? alternatively, `_SecondsNanoOffset` would be clearer

alkasm: this is only intended for internal use, but it's a tuple holding (seconds, nanoseconds) - didn't feel that weird to me

alkasm: maybe im just too used to api layers where this is the norm lol

nominal/core.py
alkasm: sure, not opposed to that!

alkasm: well, I am opposed to it in the sense that I think "type" is the wrong word to denote these as it's not the datatype but the format...but we're already there so idk

alkasm: I do want to expand these signatures to include the string types.

nominal/ts.py
alxhill: yeah the equivalent type in our conjure is `Timestamp`, hence the confusion to me I suppose - doesn't matter too much for internal types, I just generally like names that represent what something "is" vs describing what the fields are. trust you if this is a common enough pattern to be clear to others

alxhill (approved):
looks good overall, happy to take a closer a look at any specifics if you want a line-by-line

nominal/ts.py
alkasm: I get what you're saying and don't necessarily disagree, but `SecondsNanoOffset` sounds even more confusing to me and idk how to disambiguate better when there's tons of other timestamp types. Maybe just `_InternalTimestamp` or something

alxhill: cool, happy to leave as is

justfile
```diff
@@ -12,7 +12,10 @@ test-e2e token:
 
 # check static typing
 check-types:
-    poetry run mypy
+    poetry run mypy --python-version 3.12
+    poetry run mypy --python-version 3.11
+    poetry run mypy --python-version 3.10
+    poetry run mypy --python-version 3.9
```
alkasm, line 18: I found this useful for local dev but is less useful for CI since we already run the full gamut across 4 python versions in CI. maybe this should be a separate just command?

PR #58, opened 2024-09-27
feat: add function to poll ingestion status for multiple datasets

alkasm:
The `upload_csv()`, `upload_pandas()`, `upload_polars()` functions have a `wait_until_complete=True` argument which will wait until the dataset is completely ingested before moving forward. You can instead set this flag to `False` which will allow the program to continue without waiting for the ingestion to complete, which will allow you to upload multiple datasets while ingestion is running in the background.

However, if you use that flag, you won't be notified if ingestion fails, or know when it completes unless you check each dataset manually. This PR adds a function `nm.wait_until_ingestions_complete()` to allow you to wait on ingestion after uploading _all_ of the datasets, so ingestion can happen in parallel.

You can use this function like so:

```py
datasets = []
for file in files:
    dataset = nm.upload_csv(..., wait_until_complete=False)
    datasets.append(dataset)

nm.wait_until_ingestions_complete(datasets)
```

### DIFF
--- a/nominal/core.py
+++ b/nominal/core.py
@@ -8,7 +8,7 @@ from datetime import datetime, timedelta
 from io import TextIOBase
 from pathlib import Path
 from types import MappingProxyType
-from typing import BinaryIO, Iterable, Iterator, Mapping, Sequence, cast
+from typing import BinaryIO, Iterable, Mapping, Sequence, cast
 
 import certifi
 from conjure_python_client import RequestsClient, ServiceConfiguration, SslConfiguration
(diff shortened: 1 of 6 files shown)


nominal/exceptions.py
```diff
@@ -1,3 +1,6 @@
+from typing import Mapping, Sequence
```
alkasm: can get rid of sequence

nominal/core.py
```diff
@@ -1025,3 +1025,22 @@ def _logs_to_conjure(
         elif isinstance(log, tuple):
             ts, body = log
             yield Log(timestamp=_flexible_time_to_integral_nanoseconds(ts), body=body)._to_conjure()
+
+
+def poll_until_ingestion_completed(datasets: Iterable[Dataset], interval: timedelta = timedelta(seconds=1)) -> None:
```
alxhill, line 1030: is this intended as a public method or only for internal use?

alxhill (approved):

nominal/core.py
alkasm: i think both; the top level version is just a passthrough, but figured if you're in `nominal.core` you probably won't be calling into the top-level functions much? idk, not a biggie either way of course

PR #59, opened 2024-09-27
fix: search_runs parameter rename: exact_name -> name_substring

alkasm:
This change renames the parameter `exact_name` to `name_substring` to communicate the intent. This performs an exact (as opposed to fuzzy) substring match on the name.

You should update your code accordingly:

```py
# in nominal-client<1.3
nm.search_runs(exact_name="some_substring")
```

should become

```py
# in nominal-client>=1.3
nm.search_runs(name_substring="some_substring")
```

This change is backwards compatible; the old argument "exact_name" is soft-deprecated - it warns, but still works, and will be removed later. Functionality with the new and old argument name is tested, as is the warning.

### DIFF
--- a/nominal/_utils.py
+++ b/nominal/_utils.py
@@ -7,10 +7,10 @@ from contextlib import contextmanager
 from dataclasses import dataclass
 from datetime import datetime, timezone
 from pathlib import Path
-from typing import BinaryIO, Callable, Iterable, Iterator, Literal, NamedTuple, ParamSpec, TypeVar, Union
+from typing import BinaryIO, Callable, Iterable, Iterator, Literal, NamedTuple, TypeVar, Union
 
 import dateutil.parser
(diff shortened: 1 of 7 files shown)


nominal/_utils.py
```diff
@@ -212,3 +212,25 @@ def reader_writer() -> Iterator[tuple[BinaryIO, BinaryIO]]:
...
+                    f"The '{old_name}' keyword argument is deprecated and will be removed in a future version, use '{new_name}' instead.",
+                    UserWarning,
+                    stacklevel=2,
+                )
+                kwargs[new_name] = kwargs.pop(old_name)
+            return f(*args, **kwargs)
+
+        return wrapper
+
+    return _deprecate_keyword_argument_decorator
```
alxhill, line 236: fancy!

alxhill (approved):

PR #62, opened 2024-10-02
feat: get current user

ross-barrett:


nominal/core.py
```diff
@@ -44,6 +45,13 @@
 ]
 
 
+@dataclass(frozen=True)
+class NominalUser:
```
alkasm, line 49:
```suggestion
class User:
```

nominal/nominal.py
```diff
@@ -10,7 +10,7 @@
 
 from . import ts
 from ._utils import FileType, FileTypes, deprecate_keyword_argument, reader_writer
-from .core import Attachment, Dataset, LogSet, NominalClient, Run, Video, poll_until_ingestion_completed
+from .core import Attachment, Dataset, LogSet, NominalClient, NominalUser, Run, Video, poll_until_ingestion_completed
```
alkasm, line 13:
```suggestion
from .core import Attachment, Dataset, LogSet, NominalClient, User, Run, Video, poll_until_ingestion_completed
```

nominal/nominal.py
```diff
@@ -48,6 +48,12 @@ def get_default_client() -> NominalClient:
     return _get_or_create_connection(_global_base_url, token)
 
 
+def get_user() -> NominalUser:
```
alkasm, line 51:
```suggestion
def get_user() -> User:
```

nominal/nominal.py
alkasm: we should also add User to __init__.py and filter it from the top-level classes

nominal/core.py
```diff
@@ -576,8 +586,13 @@ def create(cls, base_url: str, token: str | None, trust_store_path: str | None =
             _attachment_client=attachment_client,
             _video_client=video_client,
             _logset_client=logset_client,
+            _authentication_client=authentication_client,
         )
 
+    def get_user(self) -> NominalUser:
```
alkasm, line 592:
```suggestion
    def get_user(self) -> User:
```

nominal/core.py
```diff
@@ -576,8 +586,13 @@ def create(cls, base_url: str, token: str | None, trust_store_path: str | None =
...
+        response = self._authentication_client.get_my_profile(self._auth_header)
+        return NominalUser(rid=response.rid, display_name=response.display_name, email=response.email)
```
alkasm, line 594:
```suggestion
        return User(rid=response.rid, display_name=response.display_name, email=response.email)
```

alkasm (approved):

PR #65, opened 2024-10-03
feat: add access to the units endpoint to the python client

drake-nominal:
I am currently working on an ETL pipeline that, amongst other things, needs to take a user mapping of column name -> units and pre-validate that all of the units are valid before uploading _rather than_ failing mid / post upload much later.

To do this, I expose the `get_all_units`, `get_units`, and `get_commensurable_units` endpoints from our units service to users of the python client.

### DIFF
--- a/nominal/core.py
+++ b/nominal/core.py
@@ -8,7 +8,7 @@ from datetime import datetime, timedelta
 from io import TextIOBase
 from pathlib import Path
 from types import MappingProxyType
-from typing import BinaryIO, Iterable, Mapping, Optional, Sequence, cast
+from typing import BinaryIO, Iterable, Mapping, Sequence, cast
 
 import certifi
 from conjure_python_client import RequestsClient, ServiceConfiguration, SslConfiguration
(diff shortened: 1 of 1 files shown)


nominal/core.py
```diff
@@ -928,6 +932,19 @@ def get_attachments(self, rids: Iterable[str]) -> Sequence[Attachment]:
         """Retrive attachments by their RIDs."""
         return list(self._iter_get_attachments(rids))
 
+    def get_all_units(self) -> Sequence[scout_units_api.Unit]:
+        """Retrieve list of all allowable units"""
+        response = self._units_client.get_all_units(self._auth_header)
+        return [unit for units in response.units_by_property.values() for unit in units]
+
+    def get_unit(self, unit_symbol: str) -> Optional[scout_units_api.Unit]:
```
alkasm, line 940:
```suggestion
    def get_unit(self, unit_symbol: str) -> scout_units_api.Unit | None:
```
nit: this library prefers `T | None` to `Optional[T]` as per the [python stdlib recommendation](https://docs.python.org/3/library/typing.html#typing.Union)
> ... the shorthand `int | str`. Using that shorthand is recommended.

alkasm (approved):

PR #67, opened 2024-10-04
chore: refactor core from module into subpackage

alkasm:
The core.py module was getting unwieldy. This PR moves each of the independent parts of the data model into their own files. This should not affect any lib users; existing imports for publicly exposed functions/types should remain the same. Hence, this is marked as a chore.

Terminology: "resource" in this description refers to the main types of our data model that we have modeled in the python client: Runs, Datasets, Attachments, LogSets, etc.

One minor annoyance with a naive refactor is the circular dependence nature of the client itself. The client is the main entrypoint to create/get resources. But those same resources also store a reference to the client so they can make API calls. So there's a circular reference of "client depends on resource which depends on client." This can cause an issue in importing modules in Python generally. That is solved in this PR by introducing a `ClientBunch` class which is just a holder for all the conjure request clients and the auth token. That holder doesn't have any dependence on the data types in this lib, so it can be referenced by the client and the resources without any circular dependence.

Along the same lines, all internal modules (i.e. those prefixed with _) don't depend on any of the resource types---just stdlib or third party libs, so they can also be included in any file without risk of circular dependence. And any helper functions that return a resource (e.g. `_get_dataset() -> Dataset`) should be able to live in the same module where the type itself is defined (e.g. nominal/core/dataset.py).


nominal/core/_utils.py
```diff
@@ -0,0 +1,47 @@
+from typing import Iterable, Protocol, TypeVar, runtime_checkable
+
+from .._utils import logger
+
+T = TypeVar("T")
+
+
+@runtime_checkable
+class HasRid(Protocol):
+    rid: str
```
alkasm, line 12: Previously I was just using a large union, but then I have to import all the resource types. This is a simple way to tag which types can have their RIDs extracted without actual hierarchies.

nominal/core/_utils.py
```diff
@@ -0,0 +1,49 @@
...
+
+
+def rid_from_instance_or_string(value: HasRid | str) -> str:
+    if isinstance(value, str):
+        return value
+    elif isinstance(value, HasRid):
+        return value.rid
+    raise TypeError("{value!r} is not a string nor an instance with a 'rid' attribute")
```
eidukas-drake, line 20: Missing f for format string?

nominal/core/_utils.py
```diff
@@ -0,0 +1,49 @@
...
+    """
+    for field in fields:
+        self.__dict__[field] = getattr(other, field)
+
+
+def construct_user_agent_string() -> str:
+    """Constructs a user-agent string with system & Python metadata.
+    E.g.: nominal-python/1.0.0b0 (macOS-14.4-arm64-arm-64bit) cpython/3.12.4
+    """
+    import importlib.metadata
```
eidukas-drake, line 36: Move to top of file? Or is there a specific reason to import within the function?

nominal/core/attachment.py
```diff
@@ -0,0 +1,90 @@
...
+            rid=attachment.rid,
+            name=attachment.title,
+            description=attachment.description,
+            properties=MappingProxyType(attachment.properties),
+            labels=tuple(attachment.labels),
+            _clients=clients,
+        )
+
+
+def _iter_get_attachments(clients: ClientsBunch, rids: Iterable[str]) -> Iterable[Attachment]:
```
eidukas-drake, line 86: Doesn't seem used in this file-- should we name it without the underscore (public) or move it to the file where it's used?

nominal/core/client.py
```diff
@@ -0,0 +1,497 @@
...
+        name, value = property
+        q = scout_run_api.SearchQuery(property=scout_run_api.Property(name=name, value=value))
+        queries.append(q)
+    return scout_run_api.SearchQuery(and_=queries)
+
+
+def _verify_csv_path(path: Path | str) -> tuple[Path, FileType]:
+    path = Path(path)
+    file_type = FileType.from_path_dataset(path)
+    if file_type.extension not in (".csv", ".csv.gz"):
```
eidukas-drake, line 476: Should this super me a class constant in FileType? (Assuming we wrote that)... some ALLOWED_EXTENSIONS?

eidukas-drake (commented):
Overall good change, just some small style comments / nits / questions

nominal/core/_utils.py
alkasm: good catch! should get a linter running at some point soon.

nominal/core/attachment.py
alkasm: Definitely intentional to keep it internal with the underscore. I kept these types of helper functions with their resource, since it's used in multiple files (I think this one is in run.py and client.py). Not sure the best place to put them otherwise, we could have a module just for some basic conjure helpers tho?

nominal/core/client.py
alkasm: Hmm, there is a diff between what's allowed by the API vs what's allowed by certain functions. In this case the function is validating that the path is an allowed CSV explicitly

nominal/core/_utils.py
alkasm, line 20:
```suggestion
    raise TypeError(f"{value!r} is not a string nor an instance with a 'rid' attribute")
```

nominal/core/_utils.py
alkasm: Ya I should probs move these since pretty much any nominal script will end up calling this func

nominal/core/attachment.py
alkasm: for sure, but I think there's some risk of circular dependence with that. In general the utilities don't depend on the other files to avoid that. Maybe we just say these helper functions should return/use conjure types, not `nominal.core` types and then it'll be chill

PR #68, opened 2024-10-05
feat: allow users to get channel metadata and set channel types from SDK

drake-nominal:
This is needed for numerous applications. Ideally these methods are supported more natively by our service endpoints, but until then, we can expose them via some _moderately_ hacky implementations.


nominal/core.py
```diff
@@ -331,6 +334,106 @@ def add_to_dataset_from_io(
         )
         self._client._ingest_client.trigger_file_ingest(self._client._auth_header, request)
 
+    def get_channel_metadata(
+        self,
+        exact_match: list[str] | None = None,
+        fuzzy_search_text: str = "",
+        channel_names: Sequence[str] | None = None,
+    ) -> Iterable[datasource_api.ChannelMetadata]:
```
alkasm, line 342: users shouldn't be exposed to any conjure types

nominal/core.py
```diff
@@ -331,6 +334,106 @@ def add_to_dataset_from_io(
         )
         self._client._ingest_client.trigger_file_ingest(self._client._auth_header, request)
 
+    def get_channel_metadata(
+        self,
+        exact_match: list[str] | None = None,
+        fuzzy_search_text: str = "",
+        channel_names: Sequence[str] | None = None,
```
alkasm, line 341: You can use `Sequence[str] = ()` and not have to deal with the optional

alkasm: similar on `exact_match`

nominal/core.py
```diff
@@ -331,6 +334,106 @@ def add_to_dataset_from_io(
...
+                    continue
+
+                yield channel_metadata
+
+            if response.next_page_token is None:
+                break
+            else:
+                next_page_token = response.next_page_token
+
+    def set_channel_units(self, channel_to_unit_map: Mapping[str, str]) -> None:
```
alkasm, line 383: nit: remove `_map` suffix, it's duplicative with the type

nominal/core.py
```diff
@@ -331,6 +334,106 @@ def add_to_dataset_from_io(
...
+        """Set units for channels based on a provided mapping of channel names to units.
+
+        Args:
+            channel_to_unit_map: A mapping of channel names to unit symbols.
+                NOTE: any existing units may be cleared from a channel by providing None as a symbol.
```
alkasm, line 388: the type should be `str | None` then ya?

nominal/core/client.py
```diff
@@ -442,6 +443,50 @@ def get_commensurable_units(self, unit_symbol: str) -> Sequence[scout_units_api.
...
+            rid: Identifier for the logical series to look up
+        Returns:
+            Resolved metadata for the requested logical series
+        Raises:
+            conjure_python_client.ConjureHTTPError: An error occurred while looking up the logical series.
+                This typically occurs when there is no such logical series for the given RID.
+        """
+        return self._clients.logical_series.get_logical_series(self._clients.auth_header, rid)
+
+    def set_series_units(
```
alkasm, line 458: The APIs use both "series" and "channel" for the same thing but we should use "channel" everywhere in the client (this is the decided/externally-facing name)

nominal/core/dataset.py
```diff
@@ -135,6 +137,104 @@ def add_to_dataset_from_io(
         )
         self._clients.ingest.trigger_file_ingest(self._clients.auth_header, request)
 
+    def get_channel_metadata(
```
alkasm, line 140: get_channel is probably sufficient? in general we'll probably keep to all of these resource types being more metadata-y and anything that needs to be downloaded will do so through an additional method & API call on the resource (e.g. the `Attachment.get_contents()` or `Attachment.write()` methods).

nominal/core/dataset.py
```diff
@@ -135,6 +137,104 @@ def add_to_dataset_from_io(
...
+
+        # Validate that all user provided unit symbols are valid
+        for channel_name, unit_symbol in channels_to_units.items():
+            if unit_symbol is None:
+                continue
+
+            if unit_symbol not in available_units:
+                raise ValueError(
+                    f"User provided unit {unit_symbol} for channel {channel_name} does not resolve to a unit recognized by nominal."
+                )
```
alkasm, line 201: it will be super ugly, but maybe worth it to print out the available units? e.g. unit <blah> not in recognized units <all units...>?

potentially, could just log it instead, idk.

alkasm: eh i think this might just be too long / annoying

nominal/core/channel.py
```diff
@@ -0,0 +1,80 @@
...
+        else:
+            raise TypeError(f"_from_conjure not supported for input type {type(conjure_obj)}!")
+
+    @classmethod
+    def _from_channel_metadata(cls, channel: datasource_api.ChannelMetadata) -> Self:
+        channel_rid = (
+            channel.series_rid.series_archetype
+            if channel.series_rid.series_archetype
+            else channel.series_rid.logical_series
+        )
```
alkasm, line 55: q, do you know if these are actually interchangeable? I couldn't tell from the apis and was wondering about it the other day

nominal/core/client.py
```diff
@@ -442,6 +444,48 @@ def get_commensurable_units(self, unit_symbol: str) -> Sequence[scout_units_api.
         """Get the list of units that are commensurable (convertible to/from) the given unit symbol"""
         return self._clients.units.get_commensurable_units(self._clients.auth_header, unit_symbol)
 
+    def get_channel(self, rid: str) -> Channel:
+        """Get metadata for a given logical series by looking up its rid
```
alkasm, line 448: docstring nit: logical series -> channel

nominal/core/client.py
```diff
@@ -442,6 +444,48 @@ def get_commensurable_units(self, unit_symbol: str) -> Sequence[scout_units_api.
...
+        Args:
+            rid: Identifier for the logical series to look up
+        Returns:
+            Resolved metadata for the requested logical series
+        Raises:
+            conjure_python_client.ConjureHTTPError: An error occurred while looking up the logical series.
```
alkasm, line 454: not for now, just food for thought for later - we should probably consider wrapping all these conjure errors

nominal/core/client.py
```diff
@@ -442,6 +444,48 @@ def get_commensurable_units(self, unit_symbol: str) -> Sequence[scout_units_api.
...
+            conjure_python_client.ConjureHTTPError: An error occurred while setting metadata on the channel.
+                This typically occurs when either the units are invalid, or there are no
+                channels with the given RIDs present.
+        """
+        series_updates = []
+        for rid, series_type in rids_to_types.items():
+            if series_type is None:
+                unit_update = timeseries_logicalseries_api.UnitUpdate(clear_unit=timeseries_logicalseries_api.Empty())
+            else:
+                unit_update = timeseries_logicalseries_api.UnitUpdate(unit=series_type)
```
alkasm, line 476: lets move this to a function - generally want to avoid conditionally declaring variables, and a function which just returns helps remove that

nominal/core/dataset.py
```diff
@@ -135,6 +137,104 @@ def add_to_dataset_from_io(
...
+            channel.name: channel for channel in self.get_channel_metadata(channel_names=list(channels_to_units.keys()))
+        }
+
+        # For each channel / unit combination, create an update request to set the series's unit
+        # to that symbol
+        update_requests = []
+        for channel_name, unit in channels_to_units.items():
+            # No data uploaded to channel yet ...
+            if channel_name not in found_channels:
+                continue
```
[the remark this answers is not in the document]
alkasm: logging <-> click is different from what the client library should do tho.

the cli is an application, and applications decide how to handle errors. but libraries should not be opinionated on the matter, the caller should decide whether or not this is an error.

nominal/core/channel.py
```diff
@@ -0,0 +1,80 @@
...
+        if channel_rid is None:
+            raise ValueError(f"Cannot create ChannelMetadata for channel {channel.name}-- no defined RID!")
```
alkasm, line 57: nit, but the Sentence casing, exclamation, and "-- " are inconsistent with the rest of the codebase. 

Separately, is this just working around the conjure union weirdness? we should probably write a utility function for that at some point, this is annoying to handle (and we don't do it consistently)

nominal/core/channel.py
```diff
@@ -0,0 +1,80 @@
...
+    rid: str
+    name: str
+    data_source: str
+    data_type: ChannelDataType | None
+    unit: str | None
+    description: str | None
+
+    @classmethod
+    def _from_conjure(
+        cls, conjure_obj: datasource_api.ChannelMetadata | timeseries_logicalseries_api.LogicalSeries
```
alkasm, line 40: nit: I'd probably just remove this function since IIUC you always know which of the two types you have at the call-site. is there any context in which you might have either or?

nominal/core/dataset.py
alkasm: yea thats what I meant by ugly - not the code, just the rendering since there's so many units. lets remove it for now?

nominal/core/dataset.py
```diff
@@ -135,6 +137,104 @@ def add_to_dataset_from_io(
...
+                fuzzy_search_text=fuzzy_search_text,
+                previously_selected_channels={},
+                next_page_token=next_page_token,
+                page_size=None,
+                prefix=None,
+            )
+            response = self._clients.datasource.search_channels(self._clients.auth_header, query)
+            for channel_metadata in response.results:
+                if channel_names is not None and not channel_metadata.name in channel_names:
+                    continue
```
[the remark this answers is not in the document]
alkasm: do we have to expose it though, can't the user just filter after the fact?

nominal/core/channel.py
```diff
@@ -0,0 +1,80 @@
...
+
+
+@dataclasses.dataclass
+class Channel(HasRid):
+    """Metadata for working with channels."""
+
+    rid: str
+    name: str
+    data_source: str
+    data_type: ChannelDataType | None
```
alkasm, line 34: when does a channel not have a data type btw?

nominal/core/channel.py
alkasm: Ya I think we can probably just assume its logical series since its on the Dataset class (but I don't think we need to gate on it, it's still correct)

nominal/core/channel.py
alkasm: I would like to investigate this more (in the future)...seems kinda incorrect? maybe its for before ingestion has completed?

nominal/core/client.py
alkasm: think you used this in one place but not all :)

alkasm (approved):

nominal/core/dataset.py
alkasm: ah thats a great idea

PR #70, opened 2024-10-07
feat: add CLI endpoint to summarize an existing dataset

drake-nominal:
Several times we have needed to generate a CSV file containing:
- Dataset RID
- Dataset Name
- Channel Name
- Channel Unit
- Channel RID

For every single channel corresponding to a set of dataset RIDs. 

This PR adds in a simple CLI to expose this information to either the commandline, or to a CSV.

To do this, I pull in the legendary `tabulate` library to make formatting easy. 

Sample output from a test dataset:

```
poetry run python -m nominal dataset summarize --rid "ri.catalog.cerulean-staging.dataset.33af9f79-9307-4759-8394-bf84dbb1645b" 
+----------------------------+--------------+------------------------------------------------------------------------------------+--------------+--------------------------------------------------------------------------+
|        channel name        | channel unit |                                    channel rid                                     | dataset name |                               dataset rid                                |
+----------------------------+--------------+------------------------------------------------------------------------------------+--------------+--------------------------------------------------------------------------+
|        Error_Codes         |              | ri.timeseries.cerulean-staging.logical-series.00341b50-91f3-3493-a2b7-ad8a5d532dfa |  arm_3.csv   | ri.catalog.cerulean-staging.dataset.33af9f79-9307-4759-8394-bf84dbb1645b |
|       Joint_Torque_1       |      N       | ri.timeseries.cerulean-staging.logical-series.00f7c214-68df-3c64-9f21-147cb79e00f8 |  arm_3.csv   | ri.catalog.cerulean-staging.dataset.33af9f79-9307-4759-8394-bf84dbb1645b |
|      External_Force_3      |      N       | ri.timeseries.cerulean-staging.logical-series.051d78ff-b9fd-3f97-ae80-7e5029b1aeac |  arm_3.csv   | ri.catalog.cerulean-staging.dataset.33af9f79-9307-4759-8394-bf84dbb1645b |
|        Control_Mode        |              | ri.timeseries.cerulean-staging.logical-series.0d409053-cb2a-3d38-8ced-e1109cc156ed |  arm_3.csv   | ri.catalog.cerulean-staging.dataset.33af9f79-9307-4759-8394-bf84dbb1645b |
|       Temperature_2        |     Cel      | ri.timeseries.cerulean-staging.logical-series.130f2d80-d3ac-3057-93f4-dd5d796d1f97 |  arm_3.csv   | ri.catalog.cerulean-staging.dataset.33af9f79-9307-4759-8394-bf84dbb1645b |
[...]

PR #71, opened 2024-10-08
feat: implement global options / decorators for client + debug + logging within CLI

drake-nominal:
## Overview

This PR is ultimately doing 2 high level things:

- Firstly, we make it so that when the client `create_dataset` (either from io, or csv, etc.), it has a `wait_for_complete` flag. This allows us to simplify how we create datasets by removing a helper function, and just directly using the client across the board. In turn, this cleans up our includes tree which stops errors which were otherwise popping up with circular imports in the rest of this PR. **fair game to ask for this in its own pre-requisite PR**
- Adds new decorators `@global_options` and `@client_options`
  - `@global_options` adds in three new click options to a given `click.Command`: `--debug` (hidden), `-v[v]`, and `--no-color`. 
    - `--debug` spawns a pdb post-mortem debugger any time an exception would otherwise terminate the script. I don't see this being useful externally (hence, I stop it from reaching `--help` texts), but is immensely useful for developers of other CLI commands.
    - `-v` and `--no-color` control the `logging` log verbosity and style respectively with the new `ClickLogHandler` class that I've added. The tl;dr is that `-v` adds in `INFO` logs and `-vv` adds `DEBUG` logs, and without `--no-color`, all emitted logs are logged with some different color to add visual *pop* to warnings, errors, and fatals. This also provides better support for some of the worse windows terminal emulators,  though, poses few other improvements to unix users.
  - `@client_options` replaces the formerly used `@BASE_URL_OPTION` and `@TOKEN_OPTION` decorators by combining them into one, creating a `NominalClient` object, and directly passing it to the original function using the `client` key-word. This serves to remove some boilerplate within all of our CLI subcommands that utilize a `NominalClient` and ensures that we'll have a consistent look & feel to our CLI subcommands.
    - Key difference from the status quo: we now add an additional `--token-path` to allow selecting between other files than `~/.nominal.yml` which _defaults_ to `~/.nominal.yml`. `--token` now is explicitly _only_ for passing a raw access token.

## Usage Example

Imagine I have the following CLI subcommand:

``` python
@dataset_cmd.command()
@client_options
@global_options
def test(client: NominalClient):
    """ test command to demonstrate logging and exceptions """
    
    logger.log(logging.NOTSET, "test notset")
    logger.debug("test debug")
    logger.info("test info")
    logger.warning("test warning")
    logger.error("test error")
    logger.critical("test critical")
    logger.fatal("test fatal")
    
    raise RuntimeError("oof")
    click.echo("about to print winning lotter numbers!")
```

By default, we'd see this upon running the command:
image

We can increase the verbosity to see some of the other log messages:
image

If you really don't like the colors (perhaps they mess up your CI/CD?)... just tell the CLI that you don't want them
image

and pop open a debugger on that pesky "oof" error message to see what _really_ went wrong
image

NOTE: Blocked by #70

### DIFF
--- a/nominal/cli/util/click_log_handler.py
+++ b/nominal/cli/util/click_log_handler.py
@@ -38,15 +38,14 @@ class ClickLogHandler(logging.StreamHandler[typing.TextIO]):
         super().__init__(stream)
 
         self._no_color = no_color
-        
+
     def format(self, record: logging.LogRecord) -> str:
-        """ Add colors when formatting log records
-        """
-        
+        """Add colors when formatting log records"""
+
         msg = super().format(record)
         if not self._no_color:
             msg = click.style(msg, fg=self.LEVEL_TO_COLOR_MAP.get(record.levelno, "white"))
-            
+
         return msg
 
 


nominal/cli/util/global_decorators.py
```diff
@@ -0,0 +1,125 @@
...
+    site of any raised exceptions that would have otherwise crashed the program.
+
+    This is primarily directed at developers, and as such, the --debug flag that is added to click commands
+    is witheld from the emitted --help text.
+    """
+    debug_option = click.option(
+        "--debug",
+        is_flag=True,
+        default=False,
+        show_default=True,
```
alkasm, line 67: I think you can just remove these two lines - the default for a flag is false and showing the default isn't helpful if it's hidden right?

nominal/cli/util/global_decorators.py
```diff
@@ -0,0 +1,125 @@
...
+        hidden=True,
+        help="If true, spawn a python debugger upon any exception being raised",
```
alkasm, line 69: super small nit, but would just remove "If true", I think that's implied

nominal/cli/util/__init__.py
```diff
@@ -0,0 +1,6 @@
+from .global_decorators import client_options, global_options
+
+__all__ = [
+    "client_options",
+    "global_options",
+]
```
alkasm, line 6: I would either define these functions in this init, or just leave it empty - it's not to be imported by any users so the additional flexibility of being able to do `from . import client_options` as opposed to e.g. `from ._options import client_options` is kinda w/e. we don't need to "export" them since it's all the same package iykwim

nominal/cli/util/global_decorators.py
```diff
@@ -0,0 +1,125 @@
...
+
+
+def global_options(func: typing.Callable[..., None]) -> typing.Callable[..., None]:
+    """Helper decorator that combines together debug_switch and verbosity_switch to provide commonly and globally
+    used utility options across all CLI endpoints within the nominal python client.
+    """
+    return functools.wraps(func)(debug_switch(verbosity_switch(func)))
+
+
+def client_options(func: typing.Callable[..., None]) -> typing.Callable[..., None]:
```
alkasm, line 92: do you know if the click decorators work with the ParamSpec thing for callables? See `deprecate_keyword_argument` in `nominal/_utils.py` for an example.

nominal/cli/util/__init__.py
alkasm: ah ok. hm..will think about this a bit

nominal/cli/util/global_decorators.py
alkasm: oof, sorry for the runaround! hopefully it was at least fun 😅

nominal/_config.py
```diff
@@ -44,8 +44,8 @@ def get_token(self, url: str) -> str:
         raise NominalConfigError(f"url {url!r} not found in config: set a token with `nom auth set-token`")
 
 
-def get_token(url: str) -> str:
-    return NominalConfig.from_yaml().get_token(_strip_scheme(url))
+def get_token(url: str, config_path: Path = _DEFAULT_NOMINAL_CONFIG_PATH) -> str:
```
alkasm, line 47: 👍🏼

nominal/cli/attachment.py
```diff
@@ -15,47 +16,35 @@ def attachment_cmd() -> None:
 
 @attachment_cmd.command()
 @click.option("-r", "--rid", required=True)
-@BASE_URL_OPTION
-@TOKEN_OPTION
-def get(rid: str, base_url: str, token: str | None) -> None:
+@client_options
+@global_options
+def get(rid: str, client: NominalClient) -> None:
```
alkasm, line 21: nice! this is a slick change :)

nominal/cli/util/global_decorators.py
```diff
@@ -0,0 +1,122 @@
...
+        show_default=True,
+        help="Base URL of the Nominal API to hit. Useful for hitting other clusters, e.g., staging for internal users.",
+    )
+    token_option = click.option(
+        "--token",
+        default=pathlib.Path("~/.nominal.yml").expanduser().resolve(),
+        type=click.Path(dir_okay=False, exists=True, resolve_path=True, path_type=pathlib.Path),
+        show_default=True,
+        help="Path to the yaml file containing the Nominal access token for authenticating with the API",
+    )
```
alkasm, line 113: this is semantically different than what existed previously right? I'd rather allow explicitly passing tokens in - this is 50% for debugging, so prefer the parameterization to be explicit rather than indirect

nominal/core/client.py
```diff
@@ -194,6 +196,7 @@ def create_dataset_from_io(
         timestamp_type: _AnyTimestampType,
         file_type: tuple[str, str] | FileType = FileTypes.CSV,
         description: str | None = None,
+        poll_until_completed: bool = False,
```
alkasm, line 199: I'd rather callers just do this themselves

alkasm (approved):

alkasm:
Huge QoL improvement, well done. I've never seen the `--debug` to spawn pdb pattern before but I love it.

nominal/_config.py
```diff
@@ -8,7 +8,7 @@
 
 from nominal.exceptions import NominalConfigError
 
-_DEFAULT_NOMINAL_CONFIG_PATH = Path("~/.nominal.yml").expanduser()
+_DEFAULT_NOMINAL_CONFIG_PATH = Path("~/.nominal.yml").expanduser().resolve()
```
alkasm, line 11: for my own curiousity, why is/was the resolve necessary?

nominal/cli/util/click_log_handler.py
```diff
@@ -0,0 +1,74 @@
...
+
+    See: https://click.palletsprojects.com/en/8.1.x/utils/#printing-to-stdout
+    """
+
+    LEVEL_TO_COLOR_MAP = {
+        logging.DEBUG: "blue",
+        logging.INFO: "cyan",
+        logging.WARNING: "yellow",
+        logging.ERROR: "red",
+        logging.FATAL: "bright_red",
```
alkasm, line 25: unimportant information: logging.WARN and logging.CRITICAL are equivalent but preferred over WARNING and FATAL respectively

nominal/cli/util/click_log_handler.py
```diff
@@ -0,0 +1,74 @@
...
+        try:
+            msg = self.format(record)
+            if not self._no_color:
+                msg = click.style(msg, fg=self.LEVEL_TO_COLOR_MAP.get(record.levelno, "white"))
+
+            stream = self.stream
+            # issue 35046: merged two stream.writes into one.
+            stream.write(msg + self.terminator)
+            self.flush()
+        except RecursionError:  # See issue 36272
```
alkasm, line 55: can you link the issues in the comments?

alkasm (approved):

nominal/cli/util/click_log_handler.py
alkasm: oops, I thought it was the other way around!

nominal/cli/util/click_log_handler.py
alkasm: maybe this should be handled in `format()` rather than `emit()`? then you can just super().format() and optionally change the color

PR #72, opened 2024-10-08
feat: archive run

ross-barrett:
This allows a run to be archived from the client

### DIFF
--- a/nominal/core/run.py
+++ b/nominal/core/run.py
@@ -120,6 +120,12 @@ class Run(HasRid):
 
     def list_attachments(self) -> Sequence[Attachment]:
         return list(self._iter_list_attachments())
+    
+    def archive(self) -> None:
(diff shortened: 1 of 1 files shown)


nominal/core/run.py
```diff
@@ -121,6 +121,12 @@ def _iter_list_attachments(self) -> Iterable[Attachment]:
     def list_attachments(self) -> Sequence[Attachment]:
         return list(self._iter_list_attachments())
 
+    def archive(self) -> None:
```
alkasm, line 124: should we also add unarchive?

alkasm (approved):

nominal/core/run.py
ross-barrett: That's a great question except that an unarchive endpoint is not exposed

PR #73, opened 2024-10-08
fix: update release please config

sidpalas:
- Add release please manifest and config
- Configure release please GHA workflow to use ^


alkasm (approved):
I feel like when I set this up the manifest was not mandatory! but oh well, 🤞🏼 this fixes it

PR #75, opened 2024-10-09
fix: remove component from release please tag

sidpalas:
- set `include-component-in-tag: false` ([defaults to true](https://github.com/googleapis/release-please/blob/64ffba96734bf158818121c067870077d903611b/schemas/config.json#L78)) to prevent release-please from searching for `nominal-vX.Y.Z` formatted tags


alkasm (approved):
Sweet, thanks!

PR #78, opened 2024-10-09
fix: revert to prior release please config

sidpalas:
Running into issues with the release please manifest config. PRs are being created properly, but the tagging is failing.It likely has to do with the `"include-component-in-tag": false,` setting which is the only one different from our other repos.

This reverts to the older approach (using the GHA with no config json file)


alkasm (approved):

alkasm:
:spiderman-pointing:

PR #80, opened 2024-10-11
feat: native tdms support

ross-barrett:


nominal/nominal.py
```diff
@@ -62,6 +64,27 @@ def get_user() -> User:
     return conn.get_user()
 
 
+def upload_tdms(file: Path | str, wait_until_complete: bool = True) -> Dataset:
+    """Create a dataset in the Nominal platform from a tsdms file."""
+    path = Path(file)
+    with TdmsFile.open(path) as tdms_file:
+        df = tdms_file.as_dataframe(time_index=True,absolute_time=True,scaled_data=True)
+        # from r"/'group'/'group'/'channel'" to "group.group.channel"
+        df.columns = [".".join(re.findall(r"'(.*?)'", col)) for col in df.columns if len(re.findall(r"'(.*?)'", col)) > 0]
```
alkasm, line 73: why do we need to rename them?

nominal/nominal.py
```diff
@@ -62,6 +64,27 @@ def get_user() -> User:
...
+
+        df.index.set_names("time", level=None, inplace=True)
+        df.reset_index(inplace=True)
```
alkasm, line 76: prefer reassignment over `inplace=True`

nominal/nominal.py
```diff
@@ -62,6 +63,44 @@ def get_user() -> User:
     return conn.get_user()
 
 
+def upload_tdms(file: Path | str, wait_until_complete: bool = True) -> Dataset:
+    """Create a dataset in the Nominal platform from a tdms file."""
+    import pandas as pd
+    from nptdms import TdmsFile, TdmsChannel, TdmsGroup
```
ross-barrett, line 69: I noticed inline imports were done elsewhere

nominal/nominal.py
```diff
@@ -62,6 +63,61 @@ def get_user() -> User:
...
+        `wait_until_ingestions_complete()` after uploading all datasets to allow for parallel ingestion.
+    """
+    import numpy as np
+    import pandas as pd
+    from nptdms import TdmsChannel, TdmsFile, TdmsGroup
+
+    path = Path(file)
+    with TdmsFile.open(path) as tdms_file:
+        # identify channels to extract
+        channels_to_export: OrderedDict[str, TdmsChannel] = OrderedDict()
```
alkasm, line 88: ordered dict seems unnecessary here

nominal/nominal.py
```diff
@@ -62,6 +63,61 @@ def get_user() -> User:
...
+            channel: TdmsChannel
+            for channel in group.channels():
+                # some channels may not have the required properties to construct a time track
+                if ("wf_increment" in channel.properties) and ("wf_start_time" in channel.properties):
+                    channel_name = f"{channel.group_name.replace(' ', '_')}.{channel.name.replace(' ', '_')}"
+                    channels_to_export[channel_name] = channel
+
+        df = pd.DataFrame.from_dict(
+            {
+                channel_name: pd.Series(data=channel[:], index=channel.time_track(absolute_time=True, accuracy="ns"))
```
alkasm, line 100: is `channel[:]` necessary? this makes a copy of the data IIUC?

nominal/nominal.py
```diff
@@ -62,6 +63,61 @@ def get_user() -> User:
...
+                for channel_name, channel in channels_to_export.items()
+            }
+        )
+
+        # format for nominal upload
+        time_column = "time_ns"
+        df.index = df.index.set_names(time_column, level=None, inplace=False)
+        df = df.reset_index(inplace=False)
```
alkasm, line 108: nit but `inplace=False` is the default, so unnecessary to specify

alkasm (approved):
nits, but looks good in general!

nominal/nominal.py
ross-barrett: will remove

nominal/nominal.py
ross-barrett: They have overwritten the `__things` to make this actually read the data - I should be able to call the `.read_data` with the right args if you think that's more clear

nominal/nominal.py
ross-barrett: Fair

alkasm (approved):

nominal/nominal.py
alkasm: ah gotcha! that probably is a bit clearer, a comment would suffice too

PR #81, opened 2024-10-11
feat: get channel data as a pandas series

alkasm:
Example:

```python
ds = nm.upload_csv("temps.csv", name, "timestamp", "iso_8601")
channel = ds.get_channel("temperature")
series = channel.to_pandas()
series.plot()  # do pandas things
```


nominal/core/_clientsbunch.py
```diff
@@ -25,52 +28,40 @@
...
-    ingest: ingest_api.IngestService
     catalog: scout_catalog.CatalogService
-    attachment: attachments_api.AttachmentService
-    compute_representation: scout_compute_representation_api.ComputeRepresentationService
     checklist: scout_checks_api.ChecklistService
-    video: scout_video.VideoService
-    logset: datasource_logset.LogSetService
-    units: scout_units_api.UnitsService
+    compute_representation: scout_compute_representation_api.ComputeRepresentationService
+    dataexport: scout_dataexport_api.DataExportService
```
alkasm, line 37: dataexport was added in this PR, the other changes are due to alphabetizing this long list

nominal/core/channel.py
```diff
@@ -34,20 +49,15 @@ class Channel(HasRid):
     data_type: ChannelDataType | None
     unit: str | None
     description: str | None
+    _clients: ClientsBunch = field(repr=False)
```
alkasm, line 52: clients needed now to download the data

nominal/core/channel.py
```diff
@@ -34,20 +49,15 @@ class Channel(HasRid):
...
-        if isinstance(conjure_obj, timeseries_logicalseries_api.LogicalSeries):
-            return cls._from_logicalseries(conjure_obj)
-        elif isinstance(conjure_obj, datasource_api.ChannelMetadata):
-            return cls._from_channel_metadata(conjure_obj)
-        else:
-            raise TypeError(f"_from_conjure not supported for input type {type(conjure_obj)}!")
+    def to_pandas(self) -> pd.Series[Any]:
+        body = _get_series_values_csv(self._clients.auth_header, self._clients.dataexport, self.rid, self.name)
+        df = pd.read_csv(body, parse_dates=["timestamp"], index_col="timestamp")
+        return df[self.name]
```
alkasm, line 57: this returns a series where the timestamps are the indexes. add docstring!

nominal/core/_clientsbunch.py
```diff
@@ -1,10 +1,13 @@
 from __future__ import annotations
 
 from dataclasses import dataclass
+from functools import partial
 
 from conjure_python_client import RequestsClient, ServiceConfiguration
 from typing_extensions import Self
 
+from nominal._api.combined import scout_dataexport_api
```
alkasm, line 9: should be using relative import

nominal/core/dataset.py
```diff
@@ -140,6 +140,12 @@ def add_to_dataset_from_io(
         )
         self._clients.ingest.trigger_file_ingest(self._clients.auth_header, request)
 
+    def get_channel(self, name: str) -> Channel:
```
alkasm, line 143: added helper function for getting a single channel by name

alkasm: kinda want to rename the other function to `search_channels` since it is a search function, not "get by name" or "get by rid" etc

alkasm: since this is "name" maybe this should be `get_channel_by_name()`?

alkasm: `get_channel_by_rid` doesn't really make sense on a single dataset, but it does on the client. so we could have:

```python
client.get_channel()  # by rid
dataset.get_channel()  # by name
```

but might be kinda whack to have these with different semantics?

alkasm: the other `get_blah()` methods take a `rid`, so... maybe 

```python
client.get_channel()  # by rid
dataset.get_channel_by_name()  # by name
```

is the best

alkasm: alternatively we could force a keyword arg e.g. `get_channel(name="name")` and `get_channel(rid="rid")`

bauerm97: Probably fine to just keep this function scoped to a dataset for now

bauerm97 (approved):

PR #82, opened 2024-10-12
feat: validate login flow for new users

Ben-Epstein:
As a new user, i ran into issues where i put in the wrong api for nominal, but since the auth flow didnt throw any issues, I didn't realize. Then, working through the docs examples, i ran into erroneous errors since my auth was misconfigured.

This eagerly checks the auth and url provided so users don't go down the path of and get frustrated.

invalid base_url
image

invalid token
image

everything correct
image

### DIFF
--- a/nominal/cli/auth.py
+++ b/nominal/cli/auth.py
@@ -1,6 +1,9 @@
 from __future__ import annotations
 
+import warnings
+
 import click
+import requests
 
 from .. import _config
 from .util.global_decorators import global_options
@@ -9,6 +12,22 @@ from .util.global_decorators import global_options
 @click.group(name="auth")
 def auth_cmd() -> None:
     pass
+
+
+def _validate_token_url(token: str, base_url: str) -> None:
+    """Ensure the user sets a valid configuration before letting them import the client."""
+    token_link = "https://app.gov.nominal.io/settings/user?tab=tokens"
+    r = requests.get(f"{base_url}/openapi", headers={"Authorization": f"Bearer {token}"})
+    if r.status_code == 401:
+        raise ValueError(
+            f"Your authorization token seems to be incorrect. Please recreate one here: {token_link}"
+        ) from None
+    if r.status_code == 404:
+        raise ValueError(f"Your base_url is not correct. Ensure it points to the API and not the app.") from None
+    if r.status_code != 200:
+        raise ValueError(
+            f"There is a misconfiguration between your base_url and token. Ensure you use the API url, and create a new token: {token_link}"
+        ) from None
 
 
 @auth_cmd.command()
@@ -20,5 +39,6 @@ def auth_cmd() -> None:
 def set_token(token: str, base_url: str) -> None:
     """update the token for a given URL in the Nominal config file"""
     path = _config._DEFAULT_NOMINAL_CONFIG_PATH
+    _validate_token_url(token, base_url)
     _config.set_token(base_url, token)
     click.secho(f"Successfully set token for '{base_url}' in {path}", fg="green")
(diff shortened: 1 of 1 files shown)


nominal/cli/auth.py
```diff
@@ -11,6 +14,22 @@ def auth_cmd() -> None:
     pass
 
 
+def _validate_token_url(token: str, base_url: str) -> None:
+    """Ensure the user sets a valid configuration before letting them import the client."""
+    token_link = "https://app.gov.nominal.io/settings/user?tab=tokens"
+    r = requests.get(f"{base_url}/openapi", headers={"Authorization": f"Bearer {token}"})
```
alkasm, line 20: We have `NominalClient.get_user()` which is being used in some scripts to do this eager check--let's use that instead of `/openapi` which I don't think is particularly concrete comparatively

nominal/cli/auth.py
```diff
@@ -11,6 +14,22 @@ def auth_cmd() -> None:
...
+    if r.status_code == 401:
+        raise ValueError(
+            f"Your authorization token seems to be incorrect. Please recreate one here: {token_link}"
+        ) from None
```
alkasm, line 24: why `raise ... from None`?

Ben-Epstein: its just a shorter stacktrace for users to parse through, in case of a nested stacktrace

nominal/cli/auth.py
Ben-Epstein: nice thanks. i couldn't find that but it's much better

nominal/cli/auth.py
alkasm: ah gotcha, I ran in a repl and didn't see a difference but that's probably cuz there was no context to suppress!

alkasm: might be better overall to just wrap with a `click.secho(..., error=True)`? only mentioning since we just merged a PR that moves most of the cli printing to use `click.secho`

alkasm: (I don't have a strong opinion though either way here)

alkasm:
PR nit: would say that this is a "feat" not a "chore" - chore would be more like a no-op refactor or changing the readme, etc

nominal/cli/auth.py
Ben-Epstein: @drake-nominal should I close this in favor of #88 or should I still update this?

nominal/cli/auth.py
Ben-Epstein: @alkasm we can't call `get_user` here because the url and token haven't been set yet. That assumes a correctly setup auth, which requires this flow to complete.

nominal/cli/auth.py
Ben-Epstein: updated this

nominal/cli/auth.py
alkasm: `NominalClient.create(base_url, token).get_user()`?

Ben-Epstein: thanks, way cleaner

nominal/cli/auth.py
```diff
@@ -1,8 +1,12 @@
 from __future__ import annotations
 
+import warnings
```
alkasm, line 3: unused

nominal/cli/auth.py
```diff
@@ -1,8 +1,12 @@
...
+
 import click
+import requests
```
alkasm, line 6: unused

alkasm (approved):

nominal/cli/auth.py
Ben-Epstein: will be fixed with https://github.com/nominal-io/nominal-client/pull/96 !

PR #83, opened 2024-10-12
feat: get_channel for dataset

Ben-Epstein:
Showcasing a clean interface to get channel data for a dataset.

This is a draft because it's not clear to me how to structure the api interface in the nominal client. So hoping we can use this as an example so I can get equipped with the layout, and then repeat that for future apis

tested it on the 2 datasets i have in my account
image

image


alkasm:
Hey @Ben-Epstein, I already have a PR out for this! See #81

PR #84, opened 2024-10-15
feat: expose set_token

ross-barrett:
Clients needs a runtime accessible way of easily setting a token


alkasm (approved):
I kept this private because it will be changing soon, but ack on the need today--doesn't really matter if it's exposed or not, if we need it now, we need it!

PR #85, opened 2024-10-16
feat: wrap various unit types within the Conjure spec within a single Unit abstraction

drake-nominal:
Fitting with the theme of the rest of the SDK's API, wrap the various "Unit" types found throughout the conjure spec within a `Unit` class which is then used in the various methods that used to expose these conjure types to end users.

### DIFF
--- a/nominal/core/unit.py
+++ b/nominal/core/unit.py
@@ -1,15 +1,15 @@
 from __future__ import annotations
 
-import dataclasses
+from dataclasses import dataclass
 
 from typing_extensions import Self
 
 from .._api.combined import datasource_api, scout_asset_api, scout_run_api, scout_units_api
 
 
-@dataclasses.dataclass(frozen=True)
+@dataclass(frozen=True)
 class Unit:
-    """Plain python wrapper to encapsulate the unit data coming from various places in the Conjure Spec.
+    """Combination of the name and symbol of a unit within the supported systems of measurement.
 
     This is primarily used when setting or retrieving the units of a channel within a dataset.
     """


nominal/core/unit.py
```diff
@@ -0,0 +1,66 @@
...
+
+@dataclasses.dataclass(frozen=True)
+class Unit:
+    """Plain python wrapper to encapsulate the data from scout_units_api.Unit.
+
+    This is primarily used when setting or retrieving the units of a channel within a dataset.
+    """
+
+    # Plain english name of the unit (e.g. 'coulomb')
+    name: str
```
alkasm, line 16: you should actually turn these into docstrings on the attribute, they'll show up in the docs!

nominal/core/unit.py
```diff
@@ -0,0 +1,66 @@
...
+        return cls(
+            name=api_unit.name,
+            symbol=api_unit.symbol,
+            property=api_unit.property,
+            dimension=base_dimensions,
+            system=api_unit.system,
+        )
+
+    @classmethod
+    def _from_simple_api_unit(cls, api_unit: scout_run_api.Unit | scout_asset_api.Unit | datasource_api.Unit) -> Self:
```
alkasm, line 53: hmm a little unsure about having these properties be exposed if their semantic meaning can be lost via being constructed differently

alkasm: maybe lets only keep the basic attributes exposed for now, until we can get the backend to always return the enriched unit types?

alkasm: alternatively, if everyone is agreed on making the backend change, we can always make the extra network round trip to get all the unit information and stitch it together here, and then in the future when we do change the backend impl, the client impl can change without any difference to users

alkasm (approved):

PR #86, opened 2024-10-16
fix: remove subscript type specification for inheritance, not supported in <=3.10

drake-nominal:
With our current mypy settings, it is not allowed to inherit from a genericly type-hinted class. HOWEVER, this fails in python <=3.10 with a runtime error that is _not_ found by mypy. The fix is simple-- tell mypy to ignore this line and write code in a version agnostic way


alkasm (approved):

alkasm:
TY for the fix!

PR #88, opened 2024-10-16
feat: add convenience methods for checking if auth header is valid, and when it expires

drake-nominal:
When creating scripts to run long-running tasks, it is preferable to be able to inspect the auth token to prevent making expensive API calls that have no chance at success. This allows users to build on top of the existing SDK to add token authentication flows with ease.

### DIFF
--- a/nominal/core/_clientsbunch.py
+++ b/nominal/core/_clientsbunch.py
@@ -86,7 +86,7 @@ class ClientsBunch:
         if expiry_s is None:
             raise ValueError("Expected expiry to be present in auth token, but not present")
 
-        return datetime.datetime.fromtimestamp(expiry_s, datetime.UTC)
+        return datetime.datetime.fromtimestamp(expiry_s, datetime.timezone.utc)
 
     def auth_valid(self) -> bool:
         """Validates that authentication tokens are valid and we can successfully hit the nominal API."""
--- a/nominal/core/_clientsbunch.py
+++ b/nominal/core/_clientsbunch.py
@@ -1,9 +1,12 @@
 from __future__ import annotations
 
+import datetime
+import logging
 from dataclasses import dataclass
 from functools import partial
 
-from conjure_python_client import RequestsClient, ServiceConfiguration
+import jwt
+from conjure_python_client import ConjureHTTPError, RequestsClient, ServiceConfiguration
 from typing_extensions import Self
 
 from .._api.combined import (
@@ -22,6 +25,8 @@ from .._api.combined import (
     timeseries_logicalseries_api,
     upload_api,
 )
+
+logger = logging.getLogger(__name__)
 
 
 @dataclass(frozen=True)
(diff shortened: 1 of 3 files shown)


nominal/core/_clientsbunch.py
```diff
@@ -64,3 +69,39 @@ def from_config(cls, cfg: ServiceConfiguration, agent: str, token: str) -> Self:
             upload=client_factory(upload_api.UploadService),
             video=client_factory(scout_video.VideoService),
         )
+
+    def auth_expiration(self) -> datetime.datetime:
```
alkasm, line 73: I don't think I want this exposed as a primary method, because I don't believe all tokens that people will handle will be jwts. For automated scripts, we'll likely be moving towards _API keys_ instead which I'm not clear yet if they will be jwts

nominal/core/_clientsbunch.py
```diff
@@ -64,3 +69,39 @@ def from_config(cls, cfg: ServiceConfiguration, agent: str, token: str) -> Self:
...
+
+        return datetime.datetime.fromtimestamp(expiry_s, datetime.timezone.utc)
+
+    def auth_valid(self) -> bool:
+        """Validates that authentication tokens are valid and we can successfully hit the nominal API."""
+        try:
+            # simplest possible API call with minimal overhead to validate conectivity and
+            # proper authentication
+            self.authentication.get_my_profile(self.auth_header)
+            return True
```
alkasm, line 97: this already exists on the client via `get_user()`: https://github.com/nominal-io/nominal-client/blob/ff6f015bfba7a95617985763df1df4871eb8baee/nominal/core/client.py#L73-L75

pyproject.toml
```diff
@@ -31,6 +31,8 @@ pyyaml = "*"
 tabulate = "^0.9.0"
 types-tabulate = "^0.9.0.20240106"
 nptdms = "^1.9.0"
+pyjwt = "^2.9.0"
+cryptography = "^43.0.1"
```
alkasm, line 35: this library can kinda be a PITA to have as a dependency for a number of reasons from my experience, mostly relating to different OpenSSL versions on various OSs, would prefer for it to not be a primary dep for the core library if we can avoid it

nominal/core/_clientsbunch.py
```diff
@@ -64,3 +69,39 @@ def from_config(cls, cfg: ServiceConfiguration, agent: str, token: str) -> Self:
             upload=client_factory(upload_api.UploadService),
             video=client_factory(scout_video.VideoService),
         )
+
+    def auth_expiration(self) -> datetime.datetime:
+        """Returns the timestamp at which the current auth token becomes invalid
+
+        Raises:
+            ValueError: Malformed auth token present
```
Ben-Epstein, line 77: I get why the implementation here is more correct on a technical level, but from a simplicity perspective, maybe we should follow the same path as #82 

In this case, we shouldn't really need to check if it has expired, right? We can just catch the response error for a 401, and we'll "know" with sufficient confidence that the token has expired. especially because we validate that the original token is valid in the auth flow now

alkasm:
Closing in favor of #82, which has been merged

PR #90, opened 2024-10-17
feat: move unit logic for channels into channel classes, remove from client

drake-nominal:
Logical series and archetype series need to set the type of the channel in a different way-- logical series may (and should) be updated in a single batch, whereas archetypical series must be updated one at a time, and both using different, discrete service endpoints. This change both abstracts this away from the user (assuming they continue using `Channel` as they have been) while also providing fine-grained control to callers (by having separate classes / implementations for Logical vs Archetype series). 

As part of this change, I've realized that the current implementations for doing these activities in the client:
* make a lot more sense to perform on dataset objects
* are unsafe in the presence of series archetypes due to using logical series endpoints

and simply removed said methods.

Recently, the backend provided support for setting the channel units _at the time of data upload_: this should be considered the preferred way to set this metadata (if known ahead of time), support for which will come in a future PR.

### DIFF
--- a/nominal/core/channel.py
+++ b/nominal/core/channel.py
@@ -80,12 +80,12 @@ class Channel(HasRid, abc.ABC):
         """Instantiate a concrete channel given a channel's metadata"""
 
         if channel.series_rid.logical_series is None:
-            return ArchetypeChannel._from_conjure(clients, channel)
+            return _ArchetypeChannel._from_conjure(clients, channel)
         else:
-            return LogicalChannel._from_conjure(clients, channel)
+            return _LogicalChannel._from_conjure(clients, channel)
 
 
-class LogicalChannel(Channel):
+class _LogicalChannel(Channel):
     def set_unit(self, unit_symbol: str | Unit | None) -> None:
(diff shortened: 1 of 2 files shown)


nominal/core/channel.py
```diff
@@ -50,6 +53,14 @@ class Channel(HasRid):
...
+    @abc.abstractmethod
+    def set_unit(self, unit_symbol: str | Unit | None) -> None:
+        """Set the unit for the data contained within the channel
+
+        Args:
+            unit_symbol: Symbol of the unit to set for the channel's data
+        """
+
     def to_pandas(self) -> pd.Series[Any]:
         """Retrieve the channel data as a pandas.Series.
```
alkasm, line 66: this method doesn't work for series archetypes, fyi

alkasm (commented):
My understanding is that we're in the process of migrating off of these being distinct channel types, so I'm a bit hesitant to support these types right now

nominal/core/channel.py
alkasm: That sounds okay for now, I expect to be able to actually impl it for archetypes soon

PR #92, opened 2024-10-18
feat: add connections

alkasm:
- add `NominalClient.get_connection()`
- add `Connection.get_channel()`
- add start/end parameters to `Channel.to_pandas()`

### DIFF
--- a/nominal/core/connection.py
+++ b/nominal/core/connection.py
@@ -1,17 +1,16 @@
 from __future__ import annotations
 
-from dataclasses import dataclass, field
 import itertools
 import logging
+from dataclasses import dataclass, field
 from typing import Iterable, Mapping, Sequence
 
 from nominal._api.combined import datasource_api, timeseries_logicalseries_api, timeseries_seriescache_api
 
-from .channel import Channel
-
 from .._api.combined import scout_datasource_connection_api
 from ._clientsbunch import ClientsBunch
 from ._utils import HasRid
+from .channel import Channel
 
 
 @dataclass(frozen=True)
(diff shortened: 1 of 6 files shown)


nominal/core/connection.py
```diff
@@ -0,0 +1,113 @@
...
+        )
+        resp = self._clients.logical_series.resolve_batch(self._clients.auth_header, req)
+        for resolved_series in resp.series:
+            if resolved_series.type == "error" and resolved_series.error is not None:
+                raise RuntimeError(f"error resolving series: {resolved_series.error}")
+            # TODO(alkasm): is there a batch get_logical_series ?
+            series = self._clients.logical_series.get_logical_series(self._clients.auth_header, resolved_series.rid)
+            yield Channel._from_conjure_logicalseries_api(self._clients, series)
+
+    def _get_channels(self) -> Iterable[Channel]:
```
alkasm, line 88: this is currently "private" and unused; I'm not 100% sure if this implementation is correct, and making a distinct `get_logical_series` call per channel is not what we want to do. Still I wanted this code to stick around for now so we know how to model it in the future, the tag resolution was particularly confusing so don't want to erase the context.

nominal/ts.py
```diff
@@ -433,3 +433,11 @@ def from_flexible(cls, ts: str | datetime | IntegralNanosecondsUTC) -> Self:
         if isinstance(ts, str):
             ts = dateutil.parser.parse(ts)
         return cls.from_datetime(ts)
+
+
+_MIN_TIMESTAMP = _SecondsNanos(seconds=0, nanos=0)
+_MAX_TIMESTAMP = _SecondsNanos(seconds=9223372036, nanos=854775807)
```
alkasm, line 439: Maybe we should make `_SecondsNanos` more first class: make it public, named appropriately as e.g. `Timestamp` or similar, and put it in the function signatures.

bauerm97 (approved):

PR #93, opened 2024-10-19
feat: download dataset

Ben-Epstein:
testing with e2e + against my account

```py
import nominal as nm
import pandas as pd
from uuid import uuid4
from io import BytesIO

csv_data = b"""\
timestamp,relative_minutes,temperature,humidity
2024-09-05T18:00:00Z,0,20,50
2024-09-05T18:01:00Z,1,21,49
2024-09-05T18:02:00Z,2,22,48
2024-09-05T18:03:00Z,3,23,47
2024-09-05T18:04:00Z,4,24,46
2024-09-05T18:05:00Z,5,25,45
2024-09-05T18:06:00Z,6,26,44
2024-09-05T18:07:00Z,7,27,43
2024-09-05T18:08:00Z,8,28,42
2024-09-05T18:09:00Z,9,29,41
"""

pd.read_csv(BytesIO(csv_data), index_col="timestamp").to_csv("data.csv")

name = f"dataset-{uuid4()}"
desc = f"core test to get the dataset {uuid4()}"
ds = nm.upload_csv("data.csv", name, "timestamp", "iso_8601", desc)
df = ds.to_pandas()
df2 = ds.to_pandas(channel_exact_match=["relative", "minutes"])
```

image

### DIFF
--- a/nominal/core/dataset.py
+++ b/nominal/core/dataset.py
@@ -20,7 +20,7 @@ from nominal.core._clientsbunch import ClientsBunch
 from nominal.core._conjure_utils import _available_units, _build_unit_update
 from nominal.core._multipart import put_multipart_upload
 from nominal.core._utils import HasRid, update_dataclass
-from nominal.core.channel import Channel
+from nominal.core.channel import Channel, _get_series_values_csv
 from nominal.exceptions import NominalIngestError, NominalIngestFailed, NominalIngestMultiError
 from nominal.ts import _AnyTimestampType, _to_typed_timestamp_type
 
@@ -207,17 +207,10 @@ class Dataset(HasRid):
         print("index:", s.index, "index mean:", s.index.mean())
         ```
         """
-        # We concurrently download all channels for faster performance
-        futures = []
-        seriess = []
-        with ThreadPoolExecutor(max_workers=5) as pool:
-            for channel in self.get_channels(channel_exact_match, channel_fuzzy_search_text):
-                futures.append(pool.submit(channel.to_pandas))
-            if not futures:
-                raise ValueError("No channels were found given the provided filters")
-            for future in as_completed(futures):
-                seriess.append(future.result())
-        return pd.concat(seriess, axis=1)
+        rid_name = {ch.rid: ch.name for ch in self.get_channels(channel_exact_match, channel_fuzzy_search_text)}
+        body = _get_series_values_csv(self._clients.auth_header, self._clients.dataexport, rid_name)
+        df = pd.read_csv(body, parse_dates=["timestamp"], index_col="timestamp")
+        return df
 
     def set_channel_units(self, channels_to_units: Mapping[str, str | None], validate_schema: bool = False) -> None:
         """Set units for channels based on a provided mapping of channel names to units.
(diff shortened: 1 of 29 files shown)


nominal/core/dataset.py
```diff
@@ -185,6 +188,40 @@ def get_channels(
...
+        """
+        # We concurrently download all channels for faster performance
+        futures = []
+        seriess = []
+        with ThreadPoolExecutor(max_workers=5) as pool:
+            for channel in self.get_channels(channel_exact_match, channel_fuzzy_search_text):
+                futures.append(pool.submit(channel.to_pandas))
+            if not futures:
+                raise ValueError("No channels were found given the provided filters")
+            with tqdm(total=len(futures)) as progress:
```
alkasm, line 219: Understood on the want here but to me progress bars are generally an application concern and not a library concern--maybe we can provide a progress callback or similar so we can still provide the functionality without being opinionated?

nominal/core/dataset.py
```diff
@@ -185,6 +188,40 @@ def get_channels(
...
+        ```
+        import nominal as nm
+
+        rid = "..." # Taken from the UI or via the SDK
+        dataset = nm.get_dataset(rid)
+        s = dataset.to_pandas()
+        print("index:", s.index, "index mean:", s.index.mean())
+        ```
+        """
+        # We concurrently download all channels for faster performance
```
alkasm, line 211: The code path to download a single channel is actually parameterized for a list of channels as well. The backend stitches it all together into a single csv, so we don't actually need to invoke this in parallel per channel and concat them client side

nominal/core/dataset.py
Ben-Epstein: i don't want to overcomplicate it. If you'd prefer no progress bar I'll drop it, i was just thinking about some datasets with hundreds of channels

nominal/core/dataset.py
Ben-Epstein: but I think performance wise, we'd rather download n channels at the same time than ask the backend for a giant file, because we can get each channel concurrently since the network request is non-blocking.

alkasm: The backend is already creating the CSV for each channel as well, so I don't think it's likely it will be faster client-side. we can perf test tho?

Ben-Epstein: nice, thanks for that info. You are definitely right here, we don't need to test. The only thing to note here is that we are losing the "time to first response" UX since we have to wait for the entire dataset to download, but we can handle that separately.

updated to leverage that logic

nominal/core/dataset.py
```diff
@@ -185,6 +187,31 @@ def get_channels(
...
+
+        rid = "..." # Taken from the UI or via the SDK
+        dataset = nm.get_dataset(rid)
+        s = dataset.to_pandas()
+        print("index:", s.index, "index mean:", s.index.mean())
+        ```
+        """
+        rid_name = {ch.rid: ch.name for ch in self.get_channels(channel_exact_match, channel_fuzzy_search_text)}
+        body = _get_series_values_csv(self._clients.auth_header, self._clients.dataexport, rid_name)
+        df = pd.read_csv(body, parse_dates=["timestamp"], index_col="timestamp")
```
alkasm, line 212: I wonder what the backend does here if you also upload a dataset with a column named "timestamp"...

PR #95, opened 2024-10-19
chore: absolute imports

Ben-Epstein:
see slack discussion :)

### DIFF
--- a/tests/e2e/test_toplevel.py
+++ b/tests/e2e/test_toplevel.py
@@ -9,8 +9,7 @@ import pytest
 
 import nominal as nm
 from nominal import _utils
-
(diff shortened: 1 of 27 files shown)


tests/e2e/test_toplevel.py
```diff
@@ -9,8 +9,7 @@
 
 import nominal as nm
 from nominal import _utils
-
-from . import _create_random_start_end
+from tests.e2e import _create_random_start_end
```
alkasm, line 12: is this change right? tests is not a package

tests/__init__.py
alkasm, line 1: oh i see

alkasm (approved):

PR #96, opened 2024-10-20
chore: clean unused imports

Ben-Epstein:
we might want to consider moving this inside the pyproject.toml for some more organization, and adding many of the other supported ruff rules

### DIFF
--- a/justfile
+++ b/justfile
@@ -27,14 +27,14 @@ check-format:
 
 # check import ordering | fix with `just fix-imports`
 check-imports:
-    poetry run ruff check --select I
+    poetry run ruff check --select I --select F401
 
 # run all static analysis checks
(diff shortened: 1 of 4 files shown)


alkasm (approved):
surprisingly fewer than I would have guessed!

alkasm:
>we might want to consider moving this inside the pyproject.toml for some more organization, and adding many of the other supported ruff rules

+1, goes along with the ethos expressed in https://github.com/nominal-io/nominal-client/blob/main/CONTRIBUTING.md:

> As a rule, all tools should be configured via pyproject.toml, and should prefer configuration over parameters for project information. For example, `poetry run mypy` should work without having to run `poetry run mypy nominal`.

PR #97, opened 2024-10-20
fix: early return in set_units, add coverage + unit tests

Ben-Epstein:
* coveragerc
* test structure
* some testing in cli and core
* caught a bug :)

### DIFF
--- a/justfile
+++ b/justfile
@@ -4,7 +4,7 @@ install:
 
 # run unit tests
 test:
-    poetry run pytest --cov=nominal --cov-branch --cov-report=html --cov-report=term --cov-config=.coveragerc
+    poetry run pytest
 
 # run e2e tests | token: auth token for api.nominal.test
 test-e2e token:
--- a/justfile
+++ b/justfile
@@ -4,7 +4,7 @@ install:
 
 # run unit tests
 test:
-    poetry run pytest
+    poetry run pytest --cov=nominal --cov-branch --cov-report=html --cov-report=term --cov-config=.coveragerc
 
 # run e2e tests | token: auth token for api.nominal.test
 test-e2e token:
--- a/nominal/core/dataset.py
+++ b/nominal/core/dataset.py
@@ -233,7 +233,7 @@ class Dataset(HasRid):
 
             if unit_symbol not in supported_symbols:
                 raise ValueError(
-                    f"Provided unit '{unit_symbol}' for channel '{channel_name}' does not resolve to a unit recognized by nominal."
+                    f"Provided unit '{unit_symbol}' for channel '{channel_name}' does not resolve to a unit recognized by nominal. "
                     "For more information on valid symbols, see https://ucum.org/ucum"
                 )
 
@@ -261,6 +261,8 @@ class Dataset(HasRid):
             )
             update_requests.append(channel_request)
 
+        if not update_requests:
+            return
         # Set units in database
         request = timeseries_logicalseries_api.BatchUpdateLogicalSeriesRequest(update_requests)
         self._clients.logical_series.batch_update_logical_series(self._clients.auth_header, request)
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -60,6 +60,7 @@ build-backend = "poetry.core.masonry.api"
 [tool.ruff]
 line-length = 120
 exclude = ["nominal/_api/*"]
+include = ["nominal/**/*.py", "tests/**/*.py"]
 
 [tool.mypy]
 strict = true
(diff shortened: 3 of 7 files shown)


pyproject.toml
```diff
@@ -59,6 +60,7 @@ build-backend = "poetry.core.masonry.api"
 [tool.ruff]
 line-length = 120
 exclude = ["nominal/_api/*"]
+include = ["nominal/**/*.py", "tests/**/*.py"]
```
alkasm, line 63: is the wildcard in the path here going to match any level of subdir? is it sufficient to use something like `nominal/*`?

tests/cli/test_auth.py
```diff
@@ -0,0 +1,65 @@
+from unittest.mock import MagicMock, patch
+
+from click.testing import CliRunner
+from conjure_python_client import ConjureHTTPError
+from requests import HTTPError
+
+from nominal.cli.auth import set_token
+
+runner = CliRunner()
```
alkasm, line 9: probs slightly better to use a fixture for this?

tests/cli/test_auth.py
```diff
@@ -0,0 +1,65 @@
...
+    result = runner.invoke(set_token, ["-t", "invalid-token", "-u", "https://api.gov.nominal.io/api"])
+
+    assert "Your authorization token seems to be incorrect" in result.output
+    assert result.exit_code == 1
+
+
+@patch("nominal.NominalClient.create")
+def test_invalid_url(mock_create: MagicMock):
+    mock_client = MagicMock()
+    mock_create.return_value = mock_client
```
alkasm, line 35: I think you should be able to do this repetitive part inside a fixture with function scope?

tests/core/test_dataset.py
```diff
@@ -0,0 +1,180 @@
...
+    mock_dataset._clients.catalog.get_ingest_progress_v2.return_value = MagicMock(
+        ingest_status=MagicMock(type="success")
+    )
+
+    mock_dataset.poll_until_ingestion_completed(interval=timedelta(seconds=1))
+
+    mock_dataset._clients.catalog.get_ingest_progress_v2.assert_called()
+
+
+@patch("time.sleep", return_value=None)
```
alkasm, line 130: you can just pass in 0 for the time delta rather than patching time.sleep right?

tests/core/test_dataset.py
```diff
@@ -0,0 +1,180 @@
...
+    mock_dataset._clients.catalog.get_ingest_progress_v2.side_effect = [
+        MagicMock(ingest_status=MagicMock(type="inProgress")),
+        MagicMock(ingest_status=MagicMock(type="inProgress")),
+        MagicMock(ingest_status=MagicMock(type="success")),
+    ]
+
+    mock_dataset.poll_until_ingestion_completed(interval=timedelta(seconds=1))
+
+    assert mock_dataset._clients.catalog.get_ingest_progress_v2.call_count == 3
+    assert mock_sleep.call_count == 2
```
alkasm, line 141: ah I see. I dunno if I'd test if sleep is called in particular since it's kinda an impl detail (for example could use a condition variable or event instead)

alkasm: eh but there's no other way to test the polling frequency and it's trivial to change the test if it becomes an event or something instead

justfile
```diff
@@ -4,7 +4,7 @@ install:
 
 # run unit tests
 test:
-    poetry run pytest
+    poetry run pytest --cov=nominal --cov-branch --cov-report=html --cov-report=term --cov-config=.coveragerc
```
alkasm, line 7: is this config possible in pyproject.toml or nah?

nominal/core/dataset.py
```diff
@@ -261,6 +261,8 @@ def set_channel_units(self, channels_to_units: Mapping[str, str | None], validat
             )
             update_requests.append(channel_request)
 
+        if not update_requests:
+            return
```
alkasm, line 265: if this fixes a bug, the PR should be a fix: instead. Alternatively you can split out this tiny fix into a tiny PR to merge first

justfile
Ben-Epstein: yes it is! i was gonna do a bigger pyproject setup PR, but can add this in here

pyproject.toml
Ben-Epstein: no we need the .py, otherwise ruff tries to check non py files which we don't want to do right now

Ben-Epstein: image

tests/cli/test_auth.py
Ben-Epstein: true!

tests/core/test_dataset.py
Ben-Epstein: yea will resolve because of the point below, wanted to test the calls

tests/cli/test_auth.py
Ben-Epstein: sure thing

alkasm (approved):

PR #105, opened 2024-10-24
feat: ingest video from MCAP file

mariobuikhuizen:

### DIFF
--- a/nominal/core/client.py
+++ b/nominal/core/client.py
@@ -13,6 +13,7 @@ from typing_extensions import Self
 
 from nominal import _config
 from nominal._api.combined import (
+    api,
     attachments_api,
     datasource,
     datasource_logset_api,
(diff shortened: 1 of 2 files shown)


nominal/core/client.py
```diff
@@ -511,6 +512,51 @@ def get_connection(self, rid: str) -> Connection:
         response = self._clients.connection.get_connection(self._clients.auth_header, rid)
         return Connection._from_conjure(self._clients, response)
 
+    def create_video_from_mcap_io(
+        self,
+        mcap: BinaryIO,
+        topic: str,
+        name: str,
+        description: str | None = None,
+        file_type: tuple[str, str] | FileType = FileTypes.BINARY,
```
alkasm, line 521: let's add `FileTypes.MCAP`

nominal/core/client.py
```diff
@@ -511,6 +512,51 @@ def get_connection(self, rid: str) -> Connection:
...
+            ],
+            labels=list(labels),
+            properties={} if properties is None else dict(properties),
+            sources=[ingest_api.IngestSource(s3=ingest_api.S3IngestSource(path=s3_path))],
+            description=description,
+            title=name,
+        )
+        response = self._clients.ingest.ingest_mcap(self._clients.auth_header, request)
+        video_rid = response.outputs[0].target.video_rid
+        assert video_rid is not None, "No RID returned"
```
alkasm, line 557: let's handle this a bit more explicitly:
- check the length of outputs and ensure it is exactly 1 (so users don't see an IndexError, and in case there's somehow >1)
- check and raise an exception rather than assert (so that users can handle the exception reliably) - this one is mostly annoying because handling the conjure unions is kinda silly, but we don't have a great pattern for handling this across the codebase yet. it won't ever be none if the `target.type == "video"` (the constructor itself raises if that happens)

alkasm (approved):

PR #107, opened 2024-10-27
chore: cli cleanup

Ben-Epstein:
this is the proper setup for the CLI and enforces using the actual script, which we weren't using.

`poetry run nom auth set-token`

We were doing `poetry run python nominal auth set-token` which is why the `__main__` was needed

### DIFF
--- a/nominal/__main__.py
+++ b/nominal/__main__.py
@@ -1,4 +1,0 @@
-from nominal.cli import nom
-
-if __name__ == "__main__":
-    nom()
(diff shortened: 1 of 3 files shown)


nominal/__main__.py
alkasm, line 1: Having `__main__.py` was intentional, I wanted the ability to use `python -m nominal`

nominal/cli/__init__.py
```diff
@@ -12,3 +12,6 @@ def nom() -> None:
 nom.add_command(auth.auth_cmd)
 nom.add_command(dataset.dataset_cmd)
 nom.add_command(run.run_cmd)
+
+if __name__ == "__main__":
+    nom()
```
alkasm, line 17: this shouldn't be necessary - `__init__.py` shouldn't be used as a script

PR #110, opened 2024-10-29
feat: add bounds to dataset class

drake-nominal:
When sending customers code snippets for creating runs from uploaded datasets, the typical snippet shared is this:

``` python
import nominal
from nominal.core.dataset import _get_dataset
from nominal.ts import _SecondsNanos
...

client = nominal.get_default_client()
dataset = client.get_dataset("rid")
enriched_dataset = _get_dataset(client._clients.auth_header, clients._clients.catalog, dataset.rid)
dataset_bounds = enriched_dataset.bounds

start = datetime.datetime.fromtimestamp(_SecondsNanos.from_api(dataset_bounds.start).to_nanoseconds() / 1e9, datetime.timezone.utc)
end = datetime.datetime.fromtimestamp(_SecondsNanos.from_api(dataset_bounds.end).to_nanoseconds() / 1e9, datetime.timezone.utc)

run = client.create_run(
    name="run name",
    description="run description",
    start=start,
    end=end,
)
run.add_dataset(dataset=dataset.rid, ref_name="refname")
```

Instead, we should only be exposing our public api in scripts we send to users:

``` python 
import nominal

client = nominal.get_default_client()
dataset = client.get_dataset("rid")
run = client.create_run(
    name="run name",
    description = "run description",
    start = dataset.bounds.start,
    end = dataset.bounds.end,
)
run.add_dataset(dataset = dataset.rid, ref_name="refname")
```

### DIFF
--- a/nominal/core/dataset.py
+++ b/nominal/core/dataset.py
@@ -4,7 +4,7 @@ import logging
 import time
 import urllib.parse
 from dataclasses import dataclass, field
-from datetime import datetime, timedelta
+from datetime import timedelta
 from io import TextIOBase
 from pathlib import Path
 from types import MappingProxyType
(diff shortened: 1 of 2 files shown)


nominal/core/dataset.py
```diff
@@ -31,18 +31,32 @@
...
-from nominal.ts import _MAX_TIMESTAMP, _MIN_TIMESTAMP, _AnyTimestampType, _to_typed_timestamp_type
+from nominal.ts import _MAX_TIMESTAMP, _MIN_TIMESTAMP, _AnyTimestampType, _SecondsNanos, _to_typed_timestamp_type
 
 logger = logging.getLogger(__name__)
 
 
+@dataclass(frozen=True)
+class DatasetBounds:
+    start: datetime
+    end: datetime
```
alkasm, line 42: we should use `IntegralNanosecondsUTC` here instead as we do for `Run` and others; `datetime` loses precision

alkasm (changes requested):

alkasm (approved):

PR #111, opened 2024-10-29
feat: add nominal_url property to dataset

drake-nominal:
Within customers' scripts, some find it useful to print a link to the created dataset after it has been ingested for them to ctrl+click and view on the app. This exposes a property that may be used to quickly get that URL. The alternative is customer scripts hackily making the assumption that our URL scheme won't change-- this allows us to change the scheme of our URLs without breaking customer scripts.

### DIFF
--- a/nominal/core/dataset.py
+++ b/nominal/core/dataset.py
@@ -64,6 +64,7 @@ class Dataset(HasRid):
     @property
     def nominal_url(self) -> str:
         """Returns a URL to the page in the nominal app containing this dataset"""
+        # TODO (drake): move logic into _from_conjure() factory function to accomodate different URL schemes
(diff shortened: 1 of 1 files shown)


alkasm (approved):
Thanks!

nominal/core/dataset.py
```diff
@@ -61,6 +61,11 @@ def upload(self) -> upload_api.UploadService: ...
         @property
         def units(self) -> scout.UnitsService: ...
 
+    @property
+    def nominal_url(self) -> str:
+        """Returns a URL to the page in the nominal app containing this dataset"""
+        return f"https://app.gov.nominal.io/data-sources/{self.rid}"
```
alkasm, line 67: same comment as https://github.com/nominal-io/nominal-client/pull/109#discussion_r1821587175

PR #112, opened 2024-10-29
feat: expose resetting run boundaries

drake-nominal:
When creating a run, a user is required to enter in the start / end timestamps _before_:

* All datasets are added to the run
* All data is present in the datasets being added to the run (e.g. a dataset is mid ingest via multi-file or may otherwise be updated in the future)

As a result, customers can end up in a situation where the start / end bounds of a run aren't particularly accurate as data continues to be ingested into the platform, and having a simple way to just "reset" the bounds turns out to be powerful.


alkasm:
>All data is present in the datasets being added to the run (e.g. a dataset is mid ingest via multi-file or may otherwise be updated in the future)

IIUC this PR doesn't fix this issue? The datasets that are still mid-ingest are filtered out

PR #113, opened 2024-10-29
feat: allow removing datasets from runs

drake-nominal:

### DIFF
--- a/nominal/core/run.py
+++ b/nominal/core/run.py
@@ -176,7 +176,10 @@ class Run(HasRid):
             if dataset.rid == dataset_rid:
                 ref_names_to_remove.add(ref_name)
                 
-        return self.remove_ref_names(ref_names_to_remove)
+        if ref_names_to_remove:    
+            return self.remove_ref_names(ref_names_to_remove)
+        else:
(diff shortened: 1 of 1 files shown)


nominal/core/run.py
```diff
@@ -144,6 +144,43 @@ def archive(self) -> None:
         """
         self._clients.run.archive_run(self._clients.auth_header, self.rid)
 
+    def remove_ref_names(self, ref_names_to_remove: Iterable[str]) -> Self:
```
alkasm, line 147: nit: don't love this func name, but I get what you were going for.

what if we put it all in one method and had:

```py
def remove_dataset(self, *, rids: Iterable[str] = (), ref_names: Iterable[str] = ())
```
and you could call it with `rids=[r1, r2]` or `ref_names=[ref1, ref2]`?

or maybe just `remove_datasets_by_rid()` and `remove_datasets_by_refname()`?

alkasm: And I think either way I probably would just remove the singular function, doesn't seem like an "important" enough task to have a ton of functions for it

PR #118, opened 2024-11-04
feat: add method to get decimated data to Channel

mariobuikhuizen:

### DIFF
--- a/nominal/core/channel.py
+++ b/nominal/core/channel.py
@@ -16,6 +16,7 @@ from nominal._api.combined import (
     timeseries_logicalseries,
     timeseries_logicalseries_api,
 )
+from nominal._api.combined._impl import api_Timestamp
 from nominal.core._clientsbunch import HasAuthHeader
 from nominal.core._utils import HasRid
 from nominal.ts import _MAX_TIMESTAMP, _MIN_TIMESTAMP, IntegralNanosecondsUTC, _SecondsNanos
@@ -53,6 +54,8 @@ class Channel(HasRid):
         def dataexport(self) -> scout_dataexport_api.DataExportService: ...
         @property
         def logical_series(self) -> timeseries_logicalseries.LogicalSeriesService: ...
+        @property
+        def compute(self) -> scout_compute_api.ComputeService: ...
 
     def to_pandas(
         self,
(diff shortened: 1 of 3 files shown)


nominal/core/channel.py
```diff
@@ -114,6 +117,80 @@ def _from_conjure_logicalseries_api(
...
+            df = pd.DataFrame(
+                result.numeric.values,
+                columns=["value"],
+                index=[to_ts(timestamp) for timestamp in result.numeric.timestamps],
+            )
+            df.index.name = "timestamp"
+            return df
+
+        if result.bucketed_numeric is None:
+            raise ValueError("Unexpected response from compute service")
```
alkasm, line 153: can you be more explicit in the error message on what is unexpected?

nominal/core/channel.py
```diff
@@ -114,6 +117,80 @@ def _from_conjure_logicalseries_api(
...
+        Resolution in picoseconds for picosecond-granularity dataset, nanoseconds otherwise.
+        """
+        if buckets is not None:
+            # Somehow the number of points returned is buckets / 1000
+            buckets = buckets * 1000
+
+        result = self._decimate_request(start, end, buckets, resolution)
+
+        def to_ts(timestamp: api_Timestamp) -> pd.Timestamp:
+            return pd.Timestamp(timestamp.seconds, unit="s", tz="UTC") + pd.Timedelta(timestamp.nanos, unit="ns")
```
alkasm, line 140: this doesn't create/utilize a closure so I'd just define this outside of this function

nominal/core/channel.py
```diff
@@ -114,6 +117,80 @@ def _from_conjure_logicalseries_api(
...
 
+    def get_decimated(
+        self,
+        start: str | datetime | IntegralNanosecondsUTC,
+        end: str | datetime | IntegralNanosecondsUTC,
+        *,
+        buckets: int | None = None,
+        resolution: int | None = None,
+    ) -> pd.DataFrame:
+        """Retrieve the channel data as a pandas.Dataframe, decimated to the given buckets or resolution.
```
alkasm, line 128:
```suggestion
        """Retrieve the channel data as a pandas.DataFrame, decimated to the given buckets or resolution.
```

nominal/core/channel.py
```diff
@@ -114,6 +117,80 @@ def _from_conjure_logicalseries_api(
...
+
+        Enter either the number of buckets or the resolution for the output.
```
alkasm, line 130: is it valid to provide both? if not, lets check before we make the API call?

nominal/core/channel.py
```diff
@@ -114,6 +117,80 @@ def _from_conjure_logicalseries_api(
...
+        start: str | datetime | IntegralNanosecondsUTC,
+        end: str | datetime | IntegralNanosecondsUTC,
+        *,
+        buckets: int | None = None,
+        resolution: int | None = None,
+    ) -> pd.DataFrame:
+        """Retrieve the channel data as a pandas.DataFrame, decimated to the given buckets or resolution.
+
+        Enter either the number of buckets or the resolution for the output.
+        Resolution in picoseconds for picosecond-granularity dataset, nanoseconds otherwise.
```
alkasm, line 131: Oof, this just silently switches behavior depending on the dataset precision? That's unfortunate if true

nominal/core/channel.py
mariobuikhuizen: Providing both results in an API error.

nominal/core/channel.py
mariobuikhuizen: It is stated like this in the docstring of SummarizeSeriesNode.resolution.

nominal/core/dataset.py
```diff
@@ -81,6 +82,8 @@ def logical_series(self) -> timeseries_logicalseries.LogicalSeriesService: ...
         def upload(self) -> upload_api.UploadService: ...
         @property
         def units(self) -> scout.UnitsService: ...
+        @property
+        def compute(self) -> scout_compute_api.ComputeService: ...
```
alkasm, line 86: ah, this type should inherit from `Channel._Clients` and then it would propagate automatically

alkasm: Runs does this as well like so:

```py
class _Clients(Dataset._Clients, Attachment._Clients, HasAuthHeader, Protocol):
```

note that you need to put the inheritance _before_ `HasAuthHeader`

alkasm (approved):

PR #120, opened 2024-11-05
feat: add create/get asset

alkasm:

### DIFF
--- a/nominal/__init__.py
+++ b/nominal/__init__.py
@@ -1,5 +1,6 @@
 from nominal import ts
 from nominal.core import (
+    Asset,
     Attachment,
     Channel,
     Check,
@@ -11,6 +12,7 @@ from nominal.core import (
     LogSet,
     NominalClient,
     Run,
+    User,
     Video,
 )
 from nominal.nominal import (
@@ -59,6 +61,7 @@ __all__ = [
     "upload_tdms",
     "upload_video",
     # classes: when adding a new class, also add a filter to "hide" it in docs/reference/toplevel.md
+    "Asset",
     "Attachment",
     "Channel",
     "Check",
--- a/nominal/core/asset.py
+++ b/nominal/core/asset.py
@@ -1,0 +1,72 @@
+from __future__ import annotations
+
+from dataclasses import dataclass, field
+from datetime import datetime
+from types import MappingProxyType
+from typing import Mapping, Protocol, Sequence
+
+from typing_extensions import Self
+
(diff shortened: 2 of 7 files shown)


nominal/core/asset.py
```diff
@@ -0,0 +1,72 @@
...
+        self,
+        *,
+        name: str | None = None,
+        start: datetime | IntegralNanosecondsUTC | None = None,
+        end: datetime | IntegralNanosecondsUTC | None = None,
+        description: str | None = None,
+        properties: Mapping[str, str] | None = None,
+        labels: Sequence[str] | None = None,
+    ) -> Self:
+        """Replace run metadata.
```
alkasm, line 40: this docstring references run in a couple places (copied from there)

nominal/__init__.py
```diff
@@ -11,6 +12,7 @@
     LogSet,
     NominalClient,
     Run,
+    User,
```
alkasm, line 15: this was left out previously on accident!

nominal/core/asset.py
```diff
@@ -0,0 +1,72 @@
...
+    _clients: _Clients = field(repr=False)
+
+    class _Clients(HasAuthHeader, Protocol):
+        @property
+        def assets(self) -> scout_assets.AssetService: ...
+
+    def update(
+        self,
+        *,
+        name: str | None = None,
```
ad-nmnl, line 33: Should this mirror the underlying API and be called `title`?

nominal/core/client.py
```diff
@@ -563,6 +566,36 @@ def create_video_from_mcap_io(
...
+        labels: Sequence[str] = (),
+    ) -> Asset:
+        """Create an asset."""
+        request = scout_asset_api.CreateAssetRequest(
+            description=description,
+            labels=list(labels),
+            properties={} if properties is None else dict(properties),
+            title=name,
+            attachments=[],
+            data_scopes=[],
```
ad-nmnl, line 584: This is the most important part of creating an asset as it defines the underlying data. Would it be possible to implement this functionality?

nominal/core/asset.py
alkasm: We use name and title interchangeably at the API layer, but made the decision to unify this into "name" in the client lib regardless of what the API is using.

nominal/core/client.py
jackparmer: @ad-nmnl how do you imagine `data_scopes` usage here? Is this like `ref_name`?

PR #121, opened 2024-11-08
feat: adds create_workbook_from_template

ad-nmnl:
## Changes
- Adds a Workbook and Template classes
- Adds create_workbook_from_template

## Remaining work
- [ ] Add appropriate typings
- [ ] Determine if publishing should be exposed as a param or enabled by default
- [ ] Add tests

### DIFF
--- a/nominal/core/client.py
+++ b/nominal/core/client.py
@@ -20,6 +20,7 @@ from nominal._api.combined import (
     ingest_api,
     scout_catalog,
     scout_datasource_connection_api,
+    scout_notebook_api,
     scout_run_api,
     scout_video_api,
     storage_datasource_api,
@@ -41,9 +42,11 @@ from nominal.core.connection import Connection
 from nominal.core.dataset import Dataset, _get_dataset, _get_datasets
 from nominal.core.log import Log, LogSet, _get_log_set
 from nominal.core.run import Run
+from nominal.core.template import Template
 from nominal.core.unit import Unit
 from nominal.core.user import User, _get_user, _get_user_with_fallback
 from nominal.core.video import Video
+from nominal.core.workbook import Workbook
 from nominal.exceptions import NominalIngestError
 from nominal.ts import (
     IntegralNanosecondsUTC,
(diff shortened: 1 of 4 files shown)


nickpeng95:
Additional ask from Pacific Fusion: Can they also add a custom workbook title when creating from template?

nominal/core/client.py
```diff
@@ -602,6 +634,52 @@ def create_streaming_connection(
...
+            charts=charts,
+            layout=layout,
+            content=content,
+        )
+        response = self._clients.notebook.create_notebook(
+            self._clients.auth_header, request)
+        return Workbook._from_conjure(self._clients, response)
+
+    # TODO: This should probably allow for passing either a run_rid or asset_rid.
+    def create_workbook_from_template(self, run_rid: str, template_rid: str) -> Workbook:
```
alkasm, line 668: put `template_rid` in the argument list first then. that way in the future when we add support for assets here we can change the signature to 

```py
..., run_rid: str | None = None, asset_rid: str | None = None
```

nominal/core/client.py
```diff
@@ -612,10 +690,12 @@ def _create_search_runs_query(
 ) -> scout_run_api.SearchQuery:
     queries = []
     if start is not None:
-        q = scout_run_api.SearchQuery(start_time_inclusive=_SecondsNanos.from_flexible(start).to_scout_run_api())
+        q = scout_run_api.SearchQuery(
```
alkasm, line 693: it looks like you're using a different set of formatting rules. you can run `just fix` to correct the formatting. Set up `ruff` as your formatter in your IDE.

nominal/core/workbook.py
```diff
@@ -0,0 +1,34 @@
...
+    scout_notebook_api
+)
+from nominal.core._clientsbunch import HasAuthHeader
+from nominal.core._utils import HasRid
+
+
+@dataclass(frozen=True)
+class Workbook(HasRid):
+    rid: str
+    charts: any
```
alkasm, line 17: `any` is a built-in function, not the type `typing.Any` - but also, we shouldn't be using `Any`

ad-nmnl: Sorry, the intent was just that another developer could pick it up and type correctly! (seems like we went a different direction anyway)

PR #123, opened 2024-11-12
feat: add connections to run

mariobuikhuizen:


nominal/core/run.py
```diff
@@ -180,6 +181,33 @@ def update(
         update_dataclass(self, run, fields=self.__dataclass_fields__)
         return self
 
+    def add_connection(self, ref_name: str, connection: Connection | str) -> None:
```
alkasm, line 184: in general we should just do singular or plural where it makes sense and not both, unless it really improves quality of life. This just expands the client library surface area otherwise, we don't need both forms everywhere. Probably it will be uncommon to create multiple connections, so I'd opt for just `add_connection()` I think?

mariobuikhuizen: Ok, I copied it from `add_dataset(s)`

PR #128, opened 2024-11-15
feat: expose listing connections and logsets from a run object

drake-nominal:
As title-- pretty self explanatory. Allow:

``` python
client = nominal.get_default_client()
run = client.get_run("my-rid")
run.list_connections() # list connections
run.list_log_sets() # list logsets
```

### DIFF
--- a/nominal/core/connection.py
+++ b/nominal/core/connection.py
@@ -212,8 +212,8 @@ class Connection(HasRid):
 
 
 def _get_connections(
-    clients: Connection._Clients, connection_rids: Iterable[str]
-) -> Iterable[scout_datasource_connection_api.Connection]:
+    clients: Connection._Clients, connection_rids: Sequence[str]
+) -> Sequence[scout_datasource_connection_api.Connection]:
     return [clients.connection.get_connection(clients.auth_header, rid) for rid in connection_rids]
 
 
--- a/nominal/core/run.py
+++ b/nominal/core/run.py
@@ -277,7 +277,7 @@ class Run(HasRid):
         conn_rids_by_ref_name = self._list_datasource_rids("connection")
         connections_by_rids = {
             conn.rid: Connection._from_conjure(self._clients, conn)
-            for conn in _get_connections(self._clients, conn_rids_by_ref_name.values())
+            for conn in _get_connections(self._clients, list(conn_rids_by_ref_name.values()))
         }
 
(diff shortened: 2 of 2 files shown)


nominal/core/connection.py
```diff
@@ -25,11 +30,25 @@ class Connection(HasRid):
     name: str
     description: str | None
     _tags: Mapping[str, Sequence[str]]
-    _clients: ClientsBunch = field(repr=False)
+    _clients: _Clients = field(repr=False)
     _nominal_data_source_rid: str | None = None
 
+    class _Clients(HasAuthHeader, Protocol):
```
alkasm, line 36: You can subclass from e.g. `Dataset._Clients` here so that you don't need to explicitly re-list all of the conjure clients for your dependencies. Check out how it's done for `Run._Clients` here: https://github.com/nominal-io/nominal-client/blob/3093c72bd78c9382cf6b94942e7afd50baa2ea2b/nominal/core/run.py#L34-L40

nominal/core/run.py
```diff
@@ -38,20 +46,34 @@ def attachment(self) -> attachments_api.AttachmentService: ...
...
+        self, datasource_type: str | None = None, property_name: str | None = None
+    ) -> Mapping[str, str]:
+        enriched_run = self._clients.run.get_run(self._clients.auth_header, self.rid)
+        datasource_rids_by_ref_name = {}
+        for ref_name, source in enriched_run.data_sources.items():
+            if datasource_type is not None and source.data_source.type != datasource_type:
+                continue
+
+            rid = cast(
+                str, getattr(source.data_source, source.data_source.type if property_name is None else property_name)
```
alkasm, line 72: ugh the unions are so annoying

nominal/core/connection.py
```diff
@@ -198,6 +211,12 @@ def unarchive(self) -> None:
         self._clients.connection.unarchive_connection(self._clients.auth_header, self.rid)
 
 
+def _get_connections(
+    clients: Connection._Clients, connection_rids: Iterable[str]
+) -> Iterable[scout_datasource_connection_api.Connection]:
```
alkasm, line 216: nit: a list is an iterable but I'd probably just either annotate this as a sequence or return a generator expression by swapping the [] for ()

alkasm (approved):

PR #146, opened 2024-11-21
feat: adding ability to specify time column in TDMS groups

sslattery2:
Previous implementation of upload_tdms() only allowed for waveform data where the time interval property had been set. Not all users use waveforms, regular time series data may just have a time column (a la CSV). This PR adds optional args to upload_tdms() for time column name and type.

### DIFF
--- a/nominal/nominal.py
+++ b/nominal/nominal.py
@@ -95,111 +95,122 @@ def upload_tdms(
         returning. If you are uploading many datasets, set `wait_until_complete=False` instead and call
         `wait_until_ingestions_complete()` after uploading all datasets to allow for parallel ingestion.
     """
-    import numpy as np
-    import pandas as pd
+
+    if timestamp_column:
+        return _tdms_upload_with_time_column(file, name, description, timestamp_column, timestamp_type,
+                                             wait_until_complete=wait_until_complete)
+    else:
+        return _tdms_upload_with_waveform_props(file, name, description, wait_until_complete=wait_until_complete)
+
+
+def _tdms_upload_with_time_column(file: Path | str, name: str | None = None, description: str | None = None,
+                                  timestamp_column: str | None = None,
+                                  timestamp_type: ts._AnyTimestampType | None = None, *,
+                                  wait_until_complete: bool = True):
     from nptdms import TdmsChannel, TdmsFile, TdmsGroup
+    import pandas as pd
 
-    def _tdms_upload_with_time_column():
-        path = Path(file)
-        with TdmsFile.open(path) as tdms_file:
-            channels_to_export: dict[str, TdmsChannel] = {}
-            group: TdmsGroup
-            df = None
+    path = Path(file)
+    with TdmsFile.open(path) as tdms_file:
+        channels_to_export: dict[str, TdmsChannel] = {}
+        group: TdmsGroup
+        df = None
 
-            for group in tdms_file.groups():
-                time_channel = None
-                channel: TdmsChannel
+        for group in tdms_file.groups():
+            time_channel = None
+            channel: TdmsChannel
 
-                # pull out timestamp column if expected
-                for channel in group.channels():
-                    if channel.name == timestamp_column:
-                        time_channel = channel
-                if not time_channel:
-                    logger.info(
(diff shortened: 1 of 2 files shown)


nominal/nominal.py
```diff
@@ -74,15 +76,21 @@ def get_user() -> User:
 
 
 def upload_tdms(
-    file: Path | str, name: str | None = None, description: str | None = None, *, wait_until_complete: bool = True
+    file: Path | str, name: str | None = None, timestamp_column: str | None = None,
+        timestamp_type: ts._AnyTimestampType | None = None, description: str | None = None,
```
alkasm, line 80: description should be left as the third arg so that it preserves back compat when called as a positional arg

sslattery2:
Tested locally with representative TDMS; will need to come back and add unit tests.

nominal/nominal.py
```diff
@@ -94,38 +102,64 @@ def upload_tdms(
...
+                        data=channel.read_data(), index=time_channel.read_data() if timestamp_column else
+                        channel.time_track(absolute_time=True, accuracy="ns")
+                    )
+                    for channel_name, channel in channels_to_export.items()
+                })
+            df = group_df if not df else pd.merge(left=df, right=group_df, left_index=True, right_index=True)
 
         # format for nominal upload
-        time_column = "time_ns"
+        time_column = timestamp_column if timestamp_column else "time_ns"
```
alkasm, line 151: there's 5 instances of `if timestamp_column` conditionals in this function. Let's split this into two functions with less conditionals, it's very difficult to grok correctness here

nominal/nominal.py
```diff
@@ -92,42 +99,107 @@ def upload_tdms(
...
-    with TdmsFile.open(path) as tdms_file:
-        # identify channels to extract
-        channels_to_export: dict[str, TdmsChannel] = {}
-        group: TdmsGroup
-        for group in tdms_file.groups():
-            channel: TdmsChannel
-            for channel in group.channels():
-                # some channels may not have the required properties to construct a time track
-                if ("wf_increment" in channel.properties) and ("wf_start_time" in channel.properties):
+    def _tdms_upload_with_time_column():
```
alkasm, line 102: let's unnest these functions and parameterize them

nominal/nominal.py
```diff
@@ -92,42 +99,107 @@ def upload_tdms(
...
+                channel: TdmsChannel
+
+                # pull out timestamp column if expected
+                for channel in group.channels():
+                    if channel.name == timestamp_column:
+                        time_channel = channel
+                if not time_channel:
+                    logger.info(
+                        "Skipping channel group \"%s\" because expected timestamp_column \"%s\" does not exist",
+                        channel.group_name, timestamp_column)
```
alkasm, line 120: I'm not opposed to using the logger interpolation but everywhere else in our python code we're using f-strings, so might as well keep that convention

nominal/nominal.py
```diff
@@ -92,42 +99,107 @@ def upload_tdms(
...
+                    continue
+
+                # select channels for export
+                for channel in group.channels():
+                    # do not export timestamp column
+                    if channel.name == timestamp_column:
+                        continue
+
+                    # skip if unexpected column length
+                    if timestamp_column and len(channel) != len(time_channel):
```
alkasm, line 130: first condition is now unnecessary right?

nominal/nominal.py
```diff
@@ -92,42 +99,107 @@ def upload_tdms(
...
+                        channel_name: pd.Series(
+                            data=channel.read_data(), index=time_channel.read_data()
+                        )
+                        for channel_name, channel in channels_to_export.items()
+                    })
+                df = group_df if not df else pd.merge(left=df, right=group_df, left_index=True, right_index=True,
+                                                      how='outer')
+
+            # format for nominal upload
+            time_column = timestamp_column
```
alkasm, line 148: might as well toss this assignment now

nominal/nominal.py
sslattery2: @alkasm I pulled this logger syntax from elsewhere in the Python client lib

alkasm (approved):

PR #149, opened 2024-11-22
feat: export upload_mcap_video function

alxhill:
allows using `nm.upload_mcap_video` to call the function

### DIFF
--- a/nominal/__init__.py
+++ b/nominal/__init__.py
@@ -62,8 +62,8 @@ __all__ = [
     "upload_pandas",
     "upload_polars",
     "upload_tdms",
-    "upload_video",
     "upload_mcap_video",
+    "upload_video",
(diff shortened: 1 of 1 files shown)


nominal/__init__.py
```diff
@@ -37,6 +37,7 @@
     upload_polars,
     upload_tdms,
     upload_video,
+    upload_mcap_video,
```
alkasm, line 40:
```suggestion
    upload_mcap_video,
    upload_video,
```

nominal/__init__.py
```diff
@@ -62,6 +63,7 @@
     "upload_polars",
     "upload_tdms",
     "upload_video",
+    "upload_mcap_video",
```
alkasm, line 66:
```suggestion
    "upload_mcap_video",
    "upload_video",
```

alkasm (approved):

PR #151, opened 2024-11-25
fix: issues of post merge review of #106

mariobuikhuizen:


nominal/core/stream.py
```diff
@@ -128,7 +128,7 @@ def process_future(fut: concurrent.futures.Future) -> None:  # type: ignore[type
 
     def _process_timeout_batches(self) -> None:
         while self._running:
-            time.sleep(self.max_wait_sec / 10)
+            time.sleep(1)
```
alkasm, line 131: https://github.com/nominal-io/nominal-client/pull/106#discussion_r1846990691

nominal/core/connection.py
```diff
@@ -138,26 +138,40 @@ def get_channel(self, name: str, tags: dict[str, str] | None = None) -> Channel:
...
+        return self.get_write_stream(batch_size, max_wait_sec)
+
+    def get_write_stream(self, batch_size: int = 10, max_wait: IntegralSecondsDuration = 5) -> WriteStream:
+        """Stream to write non-blocking messages to a datasource.
 
         Args:
         ----
             batch_size (int): How big the batch can get before writing to Nominal. Default 10
-            max_wait_sec (int): How long a batch can exist before being flushed to Nominal. Default 5
+            max_wait (IntegralSecondsDuration): How long a batch can exist before being flushed to Nominal. Default 5
```
alkasm, line 160: by a typeful duration, I actually meant `timedelta` rather than aliasing an int

nominal/core/stream.py
```diff
@@ -19,31 +21,38 @@ class BatchItem:
     channel_name: str
     timestamp: str | datetime | IntegralNanosecondsUTC
     value: float
-    tags: Dict[str, str] | None = None
+    tags: dict[str, str] | None = None
+
 
+NominalWriteStream: TypeAlias = "WriteStream"
```
alkasm, line 27: This should be actually equal to the type, not just a type alias

alkasm: alternatively we could also deprecate this via the module's `__getattr__`

nominal/core/stream.py
```diff
@@ -19,31 +21,38 @@ class BatchItem:
...
+        self._max_wait_event = threading.Event()
+        self._pending_jobs = threading.BoundedSemaphore(3)
 
+    def start(self) -> None:
+        self._executor = concurrent.futures.ThreadPoolExecutor(max_workers=self.max_workers)
         self._timeout_thread = threading.Thread(target=self._process_timeout_batches, daemon=True)
         self._timeout_thread.start()
 
-    def __enter__(self) -> "NominalWriteStream":
+    def __enter__(self) -> "WriteStream":
```
alkasm, line 55: we have `from __future__ import annotations` so we don't need to stringify the type here

nominal/core/stream.py
```diff
@@ -99,25 +108,31 @@ def flush(self, wait: bool = False, timeout: float | None = None) -> None:
...
+            logger.debug(f"Starting flush with {len(self._batch)} records")
+            future = self._executor.submit(self._process_batch, self._batch)
+            future.add_done_callback(process_future)
 
-        # Clear metadata
-        self._batch = []
-        self._last_batch_time = time.time()
+            # Clear metadata
+            self._batch = []
+            self._last_batch_time = time.time()
```
alkasm, line 135: its not obvious which variables need to hold the lock. Wonder if this could become more clear using a queue or similar structure

nominal/core/stream.py
mariobuikhuizen: The user should not create an instance directly, but use `connection.get_nominal_write_stream` which already has a deprecation warning.

nominal/core/stream.py
mariobuikhuizen: I've refactored a bit so it's a bit more clear. We also need the lock to coordinate access _last_batch_time, I don't think a queue would solve that.

jackparmer (approved):

PR #153, opened 2024-11-26
feat: remove pydantic for dataclasses

Ben-Epstein:
While it's great in many cases (and we can add it back if the repo requires it), there's no value being had in this particular case in pydantic because we simply store a dict. 

Remove the dependency and use a dataclass.


alkasm (approved):

alkasm:
Changed title to feat as we use "chores" a bit more restrictively.

Thank you!

PR #156, opened 2024-11-27
feat: allow uploading manually timestamped videos

drake-nominal:
When dealing with videos that don't match 1:1 with the playback speed / recording speed (e.g., original video was 200fps, slowed down to 30fps for ingest into nominal), we must provide a timestamp manifest when ingesting. This takes the form of a json file, containing a single array, of all of the per-frame timestamps (in nanos since epoch). 

Annoyingly, the backend requires this to come in the form of an s3 path-- this PR allows users to input a simple array of timestamps, and have it be uploaded and ingested into nominal.

Part of this PR required me to change the pydocstyle to google style-- we were already using the style-- but I was getting a crazy amount of false-positives on documented methods:

``` shell
just fix
poetry run ruff check --fix
nominal/core/_multipart.py:57:5: D417 Missing argument descriptions in the docstring for `put_multipart_upload`: `chunk_size`, `f`, `filename`, `max_workers`, `mimetype`, `upload_client`
nominal/core/_multipart.py:140:5: D417 Missing argument descriptions in the docstring for `upload_multipart_io`: `chunk_size`, `f`, `file_type`, `max_workers`, `name`, `upload_client`
nominal/core/_multipart.py:180:5: D417 Missing argument descriptions in the docstring for `upload_multipart_file`: `chunk_size`, `file`, `file_type`, `max_workers`, `upload_client`
Found 3 errors.
error: Recipe `fix-imports` failed on line 37 with exit code 1
```

### DIFF
--- a/nominal/core/client.py
+++ b/nominal/core/client.py
@@ -364,6 +364,7 @@ class NominalClient:
             text_json_io = TextIOWrapper(json_io)
             json.dump(frame_timestamps, text_json_io)
             text_json_io.flush()
+            json_io.seek(0)
 
             logger.debug("Uploading timestamp manifests to s3")
             manifest_s3_path = upload_multipart_io(


pyproject.toml
```diff
@@ -67,6 +67,9 @@ line-length = 120
 exclude = ["nominal/_api/*"]
 include = ["nominal/**/*.py", "tests/**/*.py"]
 
+[tool.ruff.lint.pydocstyle]
+convention = "google"
```
alkasm, line 71: google uses `:` with no `-----` underline

alkasm: we probs need a separate PR to add this since many functions don't necessarily abide by it. but not opposed!

alkasm: Ah ok I guess it allows the underline. But IIUC numpy uses the underline, google uses the colon: https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html

alkasm: (doesn't really matter for now if it's not complaining, we can be more strict later)

nominal/core/_multipart.py
```diff
@@ -118,6 +124,52 @@ def put_multipart_upload(
...
+    file: pathlib.Path,
+    upload_client: upload_api.UploadService,
+    file_type: FileType | None = None,
+    chunk_size: int = DEFAULT_CHUNK_SIZE,
+    max_workers: int = DEFAULT_NUM_WORKERS,
+) -> str:
+    if file_type is None:
+        file_type = FileType.from_path(file)
+
+    file_name = file.stem.split(".")[0]
```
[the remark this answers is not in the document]
alkasm: oh! TIL. I thought `file.name == file.stem + "".join(file.suffixes)` but indeed it's `file.name == file.stem + file.suffix`. Carry on. (I might have made this assumption elsewhere in the lib...)

alkasm (approved):

PR #164, opened 2024-12-10
fix: handle duplicated channels in tdms

ckniffin-nominal:
When uploading TDMS it was found that channels were being repeated across groups. This was due to a dictionary being declared in the incorrect scope resulting in repeated data. This manifested as two channels showing up _x and _y suffixes. This fix does two things.

- Move the dictionary declaration so it is per group
- Change the suffixes to `_repeated_1` and `_repeated_2` to prevent confusion with x and y coordinate systems


mariobuikhuizen (approved):

nominal/_tdms.py
```diff
@@ -27,7 +27,10 @@ def tdms_with_time_column_to_pandas(path: Path, timestamp_column: str) -> pd.Dat
 
     # format for nominal upload
     df = functools.reduce(
-        lambda left, right: pd.merge(left, right, left_index=True, right_index=True, how="outer"), group_dfs
+        lambda left, right: pd.merge(
+            left, right, left_index=True, right_index=True, how="outer", suffixes=["_repeated_1, _repeated_2"]
```
alkasm, line 31: is this supposed to be `suffixes=["_repeated_1", "_repeated_2"]`?

alkasm (changes requested):

nominal/_tdms.py
CBlumini: Yes that is a good catch. Will update.

ckniffin-nominal: Resolved in latest commit

alkasm:
`_repeated_N` seems like a very weird suffix to me, though I agree `_x` and `_y` are also not very semantic for us. I think the other convention is simply `_left` and `_right` which feels a little less weird. In general though I can't see this being particularly desirable either way. do we have known TDMS files that have this issue, and do we know what the users would want in that scenario?

ckniffin-nominal:
> `_repeated_N` seems like a very weird suffix to me, though I agree `_x` and `_y` are also not very semantic for us. I think the other convention is simply `_left` and `_right` which feels a little less weird. In general though I can't see this being particularly desirable either way. do we have known TDMS files that have this issue, and do we know what the users would want in that scenario?

Based on some discussions I've had I'm going to revert this aspect back to the Pandas default behavior. That keeps behavior consistent across the client. With the fix in place this should be a none issue. It addresses the issue in the files that were originally causing problems so the code never gets to the point of having to add suffixes anyway.

alkasm (approved):

PR #168, opened 2024-12-12
feat: use less restrictive io type for file ingest

alkasm:
Request's data param has type `_Data | None` [as annotated](https://github.com/python/typeshed/blob/bba068c8a1f08a16dee4dab83117804291706553/stubs/requests/requests/api.pyi#L16), where `_Data` is [defined](https://github.com/python/typeshed/blob/bba068c8a1f08a16dee4dab83117804291706553/stubs/requests/requests/sessions.pyi#L62-L80) as a union including `SupportsRead[str | bytes]` from [typeshed](https://github.com/python/typeshed/blob/bba068c8a1f08a16dee4dab83117804291706553/stdlib/_typeshed/__init__.pyi#L254-L255). Since we don't need to parameterize by the string type (we already require binary file), I just copied over the impl from

```py
class SupportsRead(Protocol[_T_co]):
    def read(self, length: int = ..., /) -> _T_co: ...
```

to

```py
class HasBinaryRead(Protocol):
    def read(self, length: int = ..., /) -> bytes: ...
```

(and renamed to be in line with our other protocols)

### DIFF
--- a/nominal/_utils.py
+++ b/nominal/_utils.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 import logging
 import os
 from contextlib import contextmanager
-from typing import Any, BinaryIO, Callable, Iterator, TypeVar
+from typing import Any, BinaryIO, Callable, Iterator, Protocol, TypeVar
 
 from typing_extensions import ParamSpec
 
@@ -64,3 +64,7 @@ def deprecate_keyword_argument(new_name: str, old_name: str) -> Callable[[Callab
         return wrapper
 
     return _deprecate_keyword_argument_decorator
+
+
+class HasBinaryRead(Protocol):
+    def read(self, length: int = ..., /) -> bytes: ...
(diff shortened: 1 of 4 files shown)


mariobuikhuizen (commented):
Shouldn't the docstrings also be updated to reflect the new type?

alkasm:
> Shouldn't the docstrings also be updated to reflect the new type?

the docstrings just give an example of a binary file-like object, so I think it's fine

mariobuikhuizen (approved):

alkasm:
> Should we put this in a publicly-importable file? seems like something I could see users wanting to use in their types for code that interacts with nominal

hmm. This is the double edged sword. If I ask for something more specific (e.g. BinaryIO), it's more restrictive than it really needs to be for consumers. On the other hand, if I minimize it down to the tiniest necessary bit (just `read()`) and then suddenly need one new method (e.g. `seek()`) it's now a breaking change. Blah, why is IO so annoying in python.

alkasm:
nah lets just leave it as-is

PR #180, opened 2024-12-19
feat: add run.add_asset()

najork:
Expose the ability to add an asset to a run.

### DIFF
--- a/nominal/__init__.py
+++ b/nominal/__init__.py
@@ -17,6 +17,7 @@ from nominal.core import (
 )
 from nominal.nominal import (
     checklist_builder,
+    create_asset,
     create_run,
     create_run_csv,
     create_streaming_connection,
@@ -42,6 +43,7 @@ from nominal.nominal import (
 
 __all__ = [
     "ts",
+    "create_asset",
     "create_run",
     "create_run_csv",
     "checklist_builder",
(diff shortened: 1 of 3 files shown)


nominal/core/run.py
```diff
@@ -210,6 +212,25 @@ def list_attachments(self) -> Sequence[Attachment]:
         """List a sequence of Attachments associated with this Run."""
         return list(self._iter_list_attachments())
 
+    def add_assets(self, assets: Iterable[Asset] | Iterable[str]) -> None:
```
Kli16, line 215: does it make sense to keep this a single asset for now and then change it when the backend properly supports multi-asset runs? Just so users don't trip up on this

najork: Yeah, can do. I missed the earlier message about multi-asset run support not being implemented imminently.

nominal/__init__.py
```diff
@@ -17,6 +17,7 @@
 )
 from nominal.nominal import (
     checklist_builder,
+    create_asset,
```
Kli16, line 20: I havent touched the python client before, but out of curiousity is this import for the test? Since it's not being used by the new endpoints

Kli16 (approved):

Kli16:
thanks!

nominal/__init__.py
najork: Yeah, we weren't exposing this through the top-level package beforehand. It's not strictly required for the test, we could have just imported `nominal.nominal.create_asset` there, but from an ergonomics standpoint it seems reasonable to expose asset creation in this manner if we expect users to start adding assets to runs more regularly.

PR #185, opened 2025-01-03
feat: add asset via run creation instead

Kli16:
this PR removes the add_asset endpoint and adds it as an option in create run instead.
The add_asset endpoint is OK to remove because it is currently broken-- it will always fail validation because runs are always associated with an asset by default.

### DIFF
--- a/nominal/core/run.py
+++ b/nominal/core/run.py
@@ -212,13 +212,11 @@ class Run(HasRid):
         """List a sequence of Attachments associated with this Run."""
         return list(self._iter_list_attachments())
 
-    def add_asset(self, asset: Asset | str) -> None:
-        """Add an asset to this run.
+    def set_asset(self, asset: Asset | str) -> None:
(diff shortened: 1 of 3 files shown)


alkasm:
Since this only really makes sense to do when you're creating a new run (and right now it's a forced move since we auto-create the asset which needs to be replaced), let's move this functionality directly into the `Client.create_run` method?

nominal/core/run.py
```diff
@@ -212,18 +212,6 @@ def list_attachments(self) -> Sequence[Attachment]:
         """List a sequence of Attachments associated with this Run."""
         return list(self._iter_list_attachments())
 
-    def add_asset(self, asset: Asset | str) -> None:
-        """Add an asset to this run.
-
-        `asset` can be an `Asset` instance, or asset RID.
-        """
-        if len(self.list_assets()) != 0:
```
Kli16, line 220: this endpoint is ok to kill because this validation always failed currently-- runs are associated with an empty asset by default.

alkasm (approved):

PR #191, opened 2025-01-14
feat: allow archiving and unarchiving all possible types from the python SDK

drake-nominal:
Many types support archiving and unarchiving via the conjure API: this PR provides access to users to be able to archive these objects.

### DIFF
--- a/nominal/core/dataset.py
+++ b/nominal/core/dataset.py
@@ -374,11 +374,13 @@ class Dataset(HasRid):
         self._clients.datasource.index_channel_prefix_tree(self._clients.auth_header, request)
 
     def archive(self) -> None:
-        """Archives the dataset, preventing it from showing up in the 'All Datasets' pane in the UI."""
+        """Archive this dataset.
+        Archived datasets are not deleted, but are hidden from the UI.
+        """
         self._clients.catalog.archive_dataset(self._clients.auth_header, self.rid)
 
     def unarchive(self) -> None:
-        """Unarchives the dataset, allowing it to show up in the 'All Datasets' pane in the UI."""
+        """Unarchives this dataset, allowing it to show up in the 'All Datasets' pane in the UI."""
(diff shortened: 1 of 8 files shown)


nominal/core/dataset.py
```diff
@@ -373,6 +373,14 @@ def set_channel_prefix_tree(self, delimiter: str = ".") -> None:
         request = datasource_api.IndexChannelPrefixTreeRequest(self.rid, delimiter=delimiter)
         self._clients.datasource.index_channel_prefix_tree(self._clients.auth_header, request)
 
+    def archive(self) -> None:
+        """Archives the dataset, preventing it from showing up in the 'All Datasets' pane in the UI."""
```
alkasm, line 377: nit: would prefer to leave most of the "this has X behavior in the app" out since that's subject to change without any relation to the client lib

alkasm (approved):

nominal/core/video.py
```diff
@@ -96,7 +96,7 @@ def archive(self) -> None:
         self._clients.video.archive(self._clients.auth_header, self.rid)
 
     def unarchive(self) -> None:
-        """Unarchives the video, allowing it to show up in the 'All Videos' pane inthe UI."""
+        """Unarchives the video, allowing it to show up in the 'All Videos' pane in the UI."""
```
alkasm, line 99: nice catch :)

nominal/core/run.py
```diff
@@ -225,6 +225,8 @@ def list_assets(self) -> Sequence[Asset]:
     def archive(self) -> None:
         """Archive this run.
         Archived runs are not deleted, but are hidden from the UI.
+
+        NOTE: currently, it is not possible (yet) to unarchive a data review once archived.
```
alkasm, line 229: this says data review

alkasm, line 229:
```suggestion
        NOTE: currently, it is not possible (yet) to unarchive a run once archived.
```

PR #192, opened 2025-01-15
feat: switch from poetry to uv for environment management, and poetry to hatch for build backend

drake-nominal:
UV offers numerous benefits over poetry, first and foremost with its speed. Without using poetry to manage the environment, it then made sense to switch to `hatch` for the build-backend, as it is a much more targeted project and excellently documented when it comes to writing the `pyproject.toml`.

I manually reviewed the generated wheels and sdists (.tar.gz) artifacts produced by `uv build` and `poetry build`, and the only difference I can detect is in the `METADATA` file created in the sdist, which is upgraded to a newer version with slightly different ways of presenting largely the same information.

A lot of the work ultimately came down to converting from the `poetry`-specific version specifier of `^x.y.z` to the pep-compliant`>=x.y.z,<x+1`, as well as `=*` => `>=0.0.0`.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -19,7 +19,6 @@ classifiers = [
     "Programming Language :: Python :: 3.10",
     "Programming Language :: Python :: 3.11",
     "Programming Language :: Python :: 3.12",
-    "Programming Language :: Python :: 3.13",
 ]
 
 requires-python = ">=3.9,<4"


pyproject.toml
```diff
@@ -1,67 +1,105 @@
...
 authors = [
-    "Alexander Reynolds <alex.reynolds@nominal.io>",
-    "Alexander Hill <alex.hill@nominal.io>",
-    "JP <jack@nominal.io>",
-    "Michael Bauer <michael@nominal.io>",
+    { name = "Alexander Reynolds", email = "alex.reynolds@nominal.io" },
+    { name = "Alexander Hill", email = "alex.hill@nominal.io" },
+    { name = "JP", email = "jack@nominal.io" },
+    { name = "Michael Bauer", email = "michael@nominal.io" },
+    { name = "Drake Eidukas", email = "drake@nominal.io" },
```
alkasm, line 10: well deserved!

pyproject.toml
```diff
@@ -1,67 +1,105 @@
...
-[tool.poetry.scripts]
-nom = 'nominal.cli:nom'
+classifiers = [
+    "License :: OSI Approved :: MIT License",
+    "Programming Language :: Python :: 3",
+    "Programming Language :: Python :: 3.9",
+    "Programming Language :: Python :: 3.10",
+    "Programming Language :: Python :: 3.11",
+    "Programming Language :: Python :: 3.12",
+    "Programming Language :: Python :: 3.13",
```
alkasm, line 22: need to add 3.13 to the matrix test strategy and mypy checks in the `justfile` if we're gonna advertise 3.13

alkasm (commented):
This looks great! Appreciate the thoroughness, stoked to bring in `uv`!

najork (approved):

PR #194, opened 2025-01-15
feat: allow searching for multiple properties on an asset

drake-nominal:
For example, my assets may only be machine-identifiable by some combo of, e.g., vehicle platform & tail number. This allows for AND-ing together multiple properties in the search function.

I don't like that we have `property` and `properties`, but not willing to break back-compat without some go-ahead.

### DIFF
--- a/nominal/core/client.py
+++ b/nominal/core/client.py
@@ -889,7 +889,7 @@ class NominalClient:
         if property:
             warnings.warn(
                 "parameter 'property' of search_assets is deprecated, use 'properties' instead",
-                DeprecationWarning,
+                UserWarning,
                 stacklevel=2,
             )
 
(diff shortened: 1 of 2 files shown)


alkasm:
wanna just throw a deprecation warning on `property` so we can eventually remove it?

nominal/core/client.py
```diff
@@ -875,13 +875,20 @@ def search_assets(
...
-        Filters are ANDed together, e.g. `(asset.label == label) AND (asset.property == property)`
+        Filters are ANDed together, e.g. `(asset.label == label) AND (asset.properties == properties)`
         - `search_text`: search case-insensitive for any of the keywords in all string fields.
         - `property` is a key-value pair, e.g. ("name", "value")
+        - `properties` is a mapping of key-value pairs which must ALL be present on the asset to be included.
         """
-        return list(self._iter_search_assets(search_text, label, property))
+        # back-compat for existing code which used property
+        if property and properties:
+            properties = {property[0]: property[1], **properties}
```
alkasm, line 889: also need to handle when `property` is not None and `properties` is None

alkasm (approved):

PR #195, opened 2025-01-16
feat: allow uniformly filtering for multiple labels / properties for runs / assets

drake-nominal:
This cleans up `search_runs` and `search_assets` (and related helper functions) across client.py and nominal.py.
Generally speaking, now all methods:

* support label, property, but mark them as deprecated and warn the user if they are used
* support labels, properties, and use them all in the giant AND query that gets created.

Also cleans up associated documentation.

### DIFF
--- a/nominal/core/client.py
+++ b/nominal/core/client.py
@@ -1047,7 +1047,7 @@ def _handle_deprecated_labels_properties(
         labels = [label]
     elif labels is None:
         labels = []
-        
+
     if all([property, properties]):
         raise ValueError(f"Cannot use both property and propertiess for {function_name}.")
     elif property:


nominal/core/client.py
```diff
@@ -939,29 +961,32 @@ def get_data_review(self, rid: str) -> DataReview:
...
-        q = scout_run_api.SearchQuery(exact_match=name_substring)
-        queries.append(q)
-    if label is not None:
-        q = scout_run_api.SearchQuery(label=label)
-        queries.append(q)
-    if property is not None:
-        name, value = property
-        q = scout_run_api.SearchQuery(property=scout_run_api.Property(name=name, value=value))
-        queries.append(q)
+        queries.append(scout_run_api.SearchQuery(exact_match=name_substring))
```
stefanv, line 980: Double checking that this is not an exact query.

nominal/core/client.py
```diff
@@ -986,19 +1011,53 @@ def _logs_to_conjure(
...
+        labels = []
+
+    # back-compat for existing code which used label
+    if label:
+        warnings.warn(
+            f"parameter 'label' of {function_name} is deprecated, use 'labels' instead",
+            UserWarning,
+            stacklevel=2,
+        )
+        labels = [label, *labels]
```
stefanv, line 1049: May be worth ensuring that `label` and `labels` are not used together:

```
if all((label, labels)):
   raise
```

And same for `properties` below. A user that is aware of the new syntax should have no reason to use the old one, and should at the very least not mix them.

nominal/nominal.py
```diff
@@ -425,27 +434,37 @@ def get_asset(rid: str) -> Asset:
...
+) -> Sequence[Asset]:
     """Search for assets meeting the specified filters.
+    Filters are ANDed together, e.g. `(asset.label == label) AND (asset.search_text =~ field)`
 
-    Filters are ANDed together, e.g. `(asset.label == label) AND (asset.property == property)`
-    - `search_text`: search case-insensitive for any of the keywords in all string fields.
-    - `property` is a key-value pair, e.g. ("name", "value")
+    Args:
+        search_text: case-insensitive search for any of the keywords in all string fields
+        label: Deprecated, use labels instead.
```
stefanv, line 452: Necessary to keep the old arg in the docstring? Old code will warn, new code should not need it. (Perhaps for people reading old code?) I see some tools add a [special section for deprecations and their versions](https://mkdocstrings.github.io/griffe/reference/docstrings/#google-section-deprecated).

stefanv (commented):
Looks good @drake-nominal! I left a few minor questions/comments.

I found some tests in `tests/e2e/test_toplevel.py`, but they don't exercise this functionality. What is the testing policy in general?

stefanv (approved):

PR #196, opened 2025-01-17
feat: allow adding an asset to a run

drake-nominal:
Allow adding an asset to a Run, and document that you cannot un-add an asset from a run, only change which asset is associated with the run.

A follow-up PR will be prepared for once runs can be disassociated with an asset.

### DIFF
--- a/nominal/core/run.py
+++ b/nominal/core/run.py
@@ -214,29 +214,29 @@ class Run(HasRid):
 
     def add_to_asset(self, asset: Asset | str, remove_existing: bool = False) -> Self:
         """Add the run to the given Asset.
-        
+
         NOTE: this will unassociate any datasets from the run, which will instead point to the asset,
               which should itself have datasets associated with it.
-              
+
         NOTE: currently, it is not supported for a run to be associated with more than one asset at
               a time. An error will be raised if a run is added to an asset while already being associated
               with another asset unless remove_existing=True is provided
-              
+
         Args:
             asset: Asset (or asset RID) to associate with the run
             remove_existing: If true, remove any existing assets from the run before adding the new asset.
-            
+
         Returns:
             Updated Run object
         """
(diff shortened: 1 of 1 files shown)


nominal/core/run.py
```diff
@@ -239,27 +266,37 @@ def update(
         description: str | None = None,
         properties: Mapping[str, str] | None = None,
         labels: Sequence[str] | None = None,
+        assets: Sequence[Asset | str] | None = None,
```
Kli16, line 269: can we just take in a single asset for now? Since we don't support multiple assets (and I dont know if we'd want to specify multiple assets as a _list_)

nominal/core/run.py
```diff
@@ -239,27 +266,37 @@ def update(
...
-        Note: This replaces the metadata rather than appending it. To append to labels or properties, merge them before
+        NOTE: This replaces the metadata rather than appending it. To append to labels or properties, merge them before
         calling this method. E.g.:
 
             new_labels = ["new-label-a", "new-label-b"]
             for old_label in run.labels:
                 new_labels.append(old_label)
             run = run.update(labels=new_labels)
+
+        NOTE: currently, it is only possible to add or change an asset associated with a Run,
```
Kli16, line 283: we should fast follow with this update to the python docs

nominal/core/run.py
```diff
@@ -212,6 +212,33 @@ def list_attachments(self) -> Sequence[Attachment]:
         """List a sequence of Attachments associated with this Run."""
         return list(self._iter_list_attachments())
 
+    def add_to_asset(self, asset: Asset | str, remove_existing: bool = False) -> Self:
```
Kli16, line 215: lets take this out for now. I think its important, but lets get the more straightforward thing (exposing more of the update run endpoint) out first, and think carefully about how we want to expose more "workflow-oriented" endpoints

PR #198, opened 2025-01-18
feat: allow --token when default token path doesn't exist in CLI

drake-nominal:
When CLI methods used `@nominal.cli.utils.client_options`, if the user does not yet have a `~/.nominal.yml` and instead provides a `--token` manually, it would error out. Now, we allow the file to not exist at the click level, and instead manually verify that the path exists _if_ the user also did not pass in an auth token.

### DIFF
--- a/nominal/cli/util/global_decorators.py
+++ b/nominal/cli/util/global_decorators.py
@@ -108,7 +108,7 @@ def client_options(func: typing.Callable[Param, T]) -> typing.Callable[..., T]:
     token_path_option = click.option(
         "--token-path",
         default=_DEFAULT_NOMINAL_CONFIG_PATH,
-        type=click.Path(dir_okay=False, exists=True, resolve_path=True, path_type=pathlib.Path),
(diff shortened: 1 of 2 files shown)


nominal/cli/util/global_decorators.py
```diff
@@ -137,10 +137,17 @@ def wrapped_function(
...
+                token = get_token(base_url, token_path)
+            else:
+                raise ValueError(
+                    f"Cannot instantiate client: no token provided and token path {token_path} does not exist."
+                )
+
         client = NominalClient.create(
             base_url,
-            api_token,
+            token=token,
```
wmgroot, line 150: Is it ok for this to become a named parameter?

wmgroot: Yeah I was only worried about issues with positional vs named causing issues. Sounds good to me.

wmgroot (approved):

stefanv (approved):
It makes sense to me that `--token` should taken higher precedence than file-based configuration.

alkasm (approved):

PR #200, opened 2025-01-23
feat: forward client CA cert bundle to requests when pushing artifacts to storage

drake-nominal:
Forward client verify path (ca cert) to requests library when pushing to cloud storage. This is done within conjure, however, needs to similarly be done when we directly use the `requests` library to upload multipart files to storage (s3, minio, etc.)

### DIFF
--- a/nominal/core/_multipart.py
+++ b/nominal/core/_multipart.py
@@ -36,9 +36,9 @@ def _sign_and_upload_part_job(
             extra={"key": key, "part": part, "upload_id": upload_id, "response.url": response.url},
         )
         put_response = requests.put(
-            response.url, 
-            data=data, 
-            headers=response.headers, 
+            response.url,
+            data=data,
+            headers=response.headers,
             verify=upload_client._verify,
         )
         logger.debug(


helenyugithub (approved):

nominal/core/_multipart.py
```diff
@@ -35,7 +35,12 @@ def _sign_and_upload_part_job(
             "successfully signed multipart upload part",
             extra={"key": key, "part": part, "upload_id": upload_id, "response.url": response.url},
         )
-        put_response = requests.put(response.url, data=data, headers=response.headers)
+        put_response = requests.put(
+            response.url,
+            data=data,
+            headers=response.headers,
+            verify=upload_client._verify,
```
stefanv, line 42: Is there a linked PR to add `_verify` to `UploadService`? I just updated to the latest release and don't see it.

stefanv, line 42: Was a bit nervous about the use of an externally defined private variable, but looks like `conjure-python-client` [passes it through as-is to `requests`](https://github.com/palantir/conjure-python-client/blob/develop/conjure_python_client/_http/requests_client.py#L136), so should be OK.

stefanv (approved):

PR #204, opened 2025-01-28
feat: automatically gzip CSVs before uploading

stefanv:
As a novice following along with the quickstart guide, I noticed that data files took a long time to upload (>1M for a 3.5MB file). Turns out, upload is quick and ingestion is the delay, but before I reached that conclusion I adapted the client to automatically compress CSV files before upload. This change could reduce upload time for larger files, especially, but perhaps doing so automatically is considered too magical.

### DIFF
--- a/nominal/core/client.py
+++ b/nominal/core/client.py
@@ -1,7 +1,10 @@
 from __future__ import annotations
 
+import gzip
 import json
 import logging
+import os
+import tempfile
 import warnings
 from dataclasses import dataclass, field
 from datetime import datetime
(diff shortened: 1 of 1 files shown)


alkasm:
As a rule I don't want to "convert" customer data at all; I want to keep the property that you can always redownload your "raw" file from Nominal. With that said we should definitely look into compressing it over-the-wire, for e.g. with `content-encoding: gzip`. The backend might already support this? We should test it!

stefanv:
Ah, yes, good idea! I will try tonight.

stefanv:
@alkasm I explored this route further. Setting `Content-Encoding: gzip` works fine for uploading to S3. Unfortunately, this results in the bucket data also being gzipped, which the ingestion engine does not know how to handle. I also tried `Transfer-Encoding: gzip`, but S3 does not support that. So, probably best for the user to pre-compress their files, or we can add an explicit client option to enable compression as per this PR.

PR #206, opened 2025-01-28
feat: new yaml config to support profile -> (url, token) instead of url -> token

v-tupuri:
current client creation
```python
import nominal as nm
nm.set_token(some_url, some_token)
client = nm.get_default_client()
```

now supported client creation (lets you have multiple tokens for one url)
```python
import nominal as nm
nm.set_profile(p1, url1, some_token_1)
nm.set_profile(p2, url1, some_token_2)

client_one = nm.get_client_from_profile(p1)
client_two = nm.get_client_from_profile(p2)

```

can still use both of the above methods but deprecating `set_token`

updated a bunch of methods in `nominal.py` to not just always use the get_default_client() but have the option to use a profile as well

### DIFF
--- a/nominal/nominal.py
+++ b/nominal/nominal.py
@@ -48,10 +48,22 @@ def set_base_url(base_url: str) -> None:
     For staging environments: "https://api-staging.gov.nominal.io/api".
     For local development: "https://api.nominal.test".
     """
-    _config.get_token(base_url)
+    _config.get_token(base_url) # @vtupuri - why is this needed? error checking?
 
     global _global_base_url
     _global_base_url = base_url
+
+def set_profile(name: str, url: str, token: str) -> None:
+    """Set a named profile configuration.
+    
+    Use in conjunction with `get_client_from_profile()`.
+    
+    Args:
+        name: Name of the profile configuration
+        url: Base URL for the Nominal API (e.g. "https://api.gov.nominal.io/api")
+        token: Authentication token for the API
+    """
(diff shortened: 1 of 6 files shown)


nominal/nominal.py
```diff
@@ -48,11 +48,23 @@ def set_base_url(base_url: str) -> None:
     For staging environments: "https://api-staging.gov.nominal.io/api".
     For local development: "https://api.nominal.test".
     """
-    _config.get_token(base_url)
+    _config.get_token(base_url) # @vtupuri - why is this needed? error checking?
```
alkasm: aside: generally more common to use the format `# TODO(username): comment` for this

alkasm: good question. IIRC the intention is generally error checking, like making sure you do _have_ a token for the URL before you use it? is kinda gross tho

nominal/nominal.py
```diff
@@ -61,18 +73,33 @@ def set_token(base_url: str, token: str) -> None:
...
 def get_default_client() -> NominalClient:
     """Retrieve the default client to the Nominal platform."""
     token = _config.get_token(_global_base_url)
     return _get_or_create_connection(_global_base_url, token)
 
 
-def get_user() -> User:
-    """Retrieve the user associated with the default client."""
-    conn = get_default_client()
+def get_user(*, profile: str | None = None) -> User:
```
alkasm, line 93: oof i did not think about this being necessary. let's chat more

PR #208, opened 2025-02-04
feat: protobuf support in the python client for streaming data

v-tupuri:
wire up our connection.py to use the protobuf endpoint instead


v-tupuri:
doesnt work rn bc the library isnt published yet but works when I use poetry to point to local scout

nominal/core/connection.py
```diff
@@ -191,44 +193,50 @@ def _process_batch(self, batch: Sequence[BatchItem]) -> None:
...
+            if isinstance(sample_value, str):
+                string_points = StringPoints()
+                for item in api_batch:
+                    point = StringPoint()
+                    seconds_nanos = _SecondsNanos.from_flexible(item.timestamp)
+                    point.timestamp.seconds = seconds_nanos.seconds
+                    point.timestamp.nanos = seconds_nanos.nanos
+                    point.value = item.value
+                    string_points.points.append(point)
+                points.string_points.CopyFrom(string_points)
```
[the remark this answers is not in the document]
v-tupuri: hm the Timestamp type is from google.com.protobuf so we would have to import that in the file (not sure if this is fine?). otherwise, I can chase the rest to be nested

nominal/core/connection.py
```diff
@@ -191,44 +193,50 @@ def _process_batch(self, batch: Sequence[BatchItem]) -> None:
...
+                raise ValueError("only float and string are supported types for value")
+            return points
+
+        # Build WriteRequestNominal with properly initialized Series messages
+        request = WriteRequestNominal()
+        for api_batch in api_batches:
+            series_msg = Series()
+            series_msg.channel.name = api_batch[0].channel_name
+
+            points_msg = make_points_proto(api_batch)
```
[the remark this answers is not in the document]
v-tupuri: think the latter. pretty confident and I think its gotten a decent bit of usage already cc @bauerm97

nominal/core/connection.py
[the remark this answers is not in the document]
v-tupuri: see new implementation

pyproject.toml
```diff
@@ -144,6 +144,10 @@ disable_error_code = ["no-any-return", "no-untyped-def"]
 module = "nptdms"
 ignore_missing_imports = true
 
+[[tool.mypy.overrides]]
```
v-tupuri, line 147: should prob fix this but unsure how

nominal/core/_clientsbunch.py
```diff
@@ -27,9 +27,54 @@
...
 from typing_extensions import Self
 
 
+class ProtoWriteService(storage_writer_api.NominalChannelWriterService):
+    def write_nominal_batches(self, auth_header: str, data_source_rid: str, request: Any) -> None:
+        _headers: Dict[str, Any] = {
+            "Accept": "application/json",
+            "Content-Type": "application/x-protobuf",
+            "Authorization": auth_header,
+        }
```
alkasm, line 42: I understand this was typed like this in the generated conjure, but we can/should be more specific here--let's use the correct protobuf type, not use Any, etc. Also I don't think these annotations on the types inside the function are necessary at all.

nominal/core/_clientsbunch.py
```diff
@@ -27,9 +27,54 @@
...
+
+        _params: Dict[str, Any] = {}
+
+        _path_params: Dict[str, Any] = {
+            "dataSourceRid": data_source_rid,
+        }
+
+        _data: Any = request
```
alkasm, line 50: the request needs to be serialized still

tests/test_write_stream.py
```diff
@@ -0,0 +1,290 @@
...
+        description="A connection for testing",
+        _tags={},
+        _clients=mock_clients,
+        _nominal_data_source_rid="test-datasource-rid",
+    )
+
+
+def test_process_batch_double_points(mock_connection):
+    # Create test data
+    timestamp = datetime.now()
```
alkasm, line 47: better to always used fixed values for unit tests

nominal/core/_clientsbunch.py
```diff
@@ -27,9 +27,56 @@
     timeseries_logicalseries,
     upload_api,
 )
+from requests.adapters import (
+    Response,
+)
 from typing_extensions import Self
 
 
+class ProtoWriteService(storage_writer_api.NominalChannelWriterService):
```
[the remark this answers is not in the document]
alkasm: More flexible, yeah. But depending on private methods to be called is also making some assumptions that the private method won't change (and since it's not a public API, we shouldn't be relying on that).

nominal/core/_clientsbunch.py
```diff
@@ -30,6 +30,11 @@
...
+    def _amend_request_kwargs(self, kwargs) -> None:
```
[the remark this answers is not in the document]
v-tupuri: yea makes sense

nominal/core/batch_processor.py
```diff
@@ -0,0 +1,90 @@
...
+from nominal.ts import IntegralNanosecondsUTC, _SecondsNanos
+
+
+def _to_api_batch_key(item: BatchItem) -> tuple[str, Sequence[tuple[str, str]], str]:
+    return item.channel_name, sorted(item.tags.items()) if item.tags is not None else [], type(item.value).__name__
+
+
+def make_points_proto(api_batch: Sequence[BatchItem]) -> Points:
+    # Check first value to determine type
+    sample_value = api_batch[0].value
```
alkasm, line 32: we should def clean up these types in the future; there should be zero need to peek and test as we know the types ahead of time (i.e. at the point of enqueueing)

nominal/core/_clientsbunch.py
```diff
@@ -57,6 +83,7 @@ class ClientsBunch:
     notebook: scout.NotebookService
     checklist_execution: scout_checklistexecution_api.ChecklistExecutionService
     datareview: scout_datareview_api.DataReviewService
+    proto_write_service: ProtoWriteService
```
alkasm, line 86: nit: nix the `Service` suffix; none of the other clients use it

pyproject.toml
```diff
@@ -72,6 +72,7 @@ nominal = { workspace = true }
 
 [project.optional-dependencies]
 hdf5 = ["h5py>=3.0", "tables>=3.7"]
+proto = ["nominal-api-protos>=0.0.0"]
```
alkasm, line 75: also nit but lets use plural `protos` since the library is `nominal-api-protos`

alkasm (approved):

pyproject.toml
```diff
@@ -72,6 +72,7 @@ nominal = { workspace = true }
 
 [project.optional-dependencies]
 hdf5 = ["h5py>=3.0", "tables>=3.7"]
+protos = ["nominal-api-protos>=0.0.0"]
```
alkasm, line 75: oh, I just realized that doing this as an optional dependency means that if an existing user updates their nominal client, they likely won't specify `[protos]` and hence their ability to stream would break. 

we could keep around both implementations and raise a deprecated warning on the old one I guess?

alkasm (changes requested):
oh, I just realized that doing this as an optional dependency means that if an existing user updates their nominal client, they likely won't specify [protos] and hence their ability to stream would break.

we could keep around both implementations and raise a deprecated warning on the old one I guess?

nominal/core/connection.py
```diff
@@ -162,7 +162,7 @@ def get_write_stream(
         ----
             batch_size (int): How big the batch can get before writing to Nominal. Default 10
             max_wait (timedelta): How long a batch can exist before being flushed to Nominal. Default 5 seconds
-            use_protos (bool): Whether to send messages over the proto write service (true) or the storage writer API (false). Default False
+            use_protos (bool): Send data as protobufs (True) or as json (False - legacy). Default False
```
alkasm, line 165: I like this framing better - makes me wonder if we should model this as a bool or a union literal (e.g. `implementation: Literal["json", "protobuf"]`)

nominal/core/connection.py
```diff
@@ -150,13 +155,16 @@ def get_nominal_write_stream(self, batch_size: int = 50_000, max_wait_sec: int =
...
+    def get_write_stream(
+        self, batch_size: int = 50_000, max_wait: timedelta = timedelta(seconds=1), use_protos: bool = False
+    ) -> WriteStream:
         """Stream to write non-blocking messages to a datasource.
 
         Args:
         ----
             batch_size (int): How big the batch can get before writing to Nominal. Default 10
             max_wait (timedelta): How long a batch can exist before being flushed to Nominal. Default 5 seconds
+            use_protos (bool): Whether to use the storage writer API or the batch processor. Default False
```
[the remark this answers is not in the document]
v-tupuri: fixed

nominal/core/connection.py
```diff
@@ -184,10 +187,10 @@ def get_write_stream(
...
         """
-        if not use_protos:
+        if data_format == "json":
             warnings.warn(
                 "Using the batch processor without protos is deprecated and will be removed in a future version, "
-                "use get_write_stream(..., use_protos=True) instead.",
+                "use get_write_stream(..., data_format='protobuf') instead.",
                 UserWarning,
                 stacklevel=2,
             )
```
alkasm, line 196: i think we should just remove the warning, since we're likely going to keep the function around.

once we swap over to default to protos we can issue a warning _if_ they don't have protos installed and fallback

justfile
```diff
@@ -7,6 +7,8 @@ default:
 # development install with dependencies
 install:
     uv sync
+    uv pip install '.[hdf5,protos]'
```
alkasm, line 10: maybe we should add default groups? 

https://docs.astral.sh/uv/concepts/projects/dependencies/#default-groups

alkasm (approved):

nominal/core/connection.py
v-tupuri: yea makes sense

justfile
v-tupuri: yea looks like dev-dependencies is getting deprecated

PR #211, opened 2025-02-08
feat: provide profile-based NominalClient configuration

alkasm:
This PR creates a new "profile-based" config file for Nominal clients. 

### Why are we making this change?
- Uses a more standardized config location at `~/.config/nominal/config.yml` - this dir can be expanded later for other types of config files
- We are now able to have multiple auth configs for a single URL (useful if you have multiple tenants, multiple workspaces, or for staging and dev instances for Nominal devs)
- Profile names allow for simpler parameterization of CLIs or other apps, and are a common choice for authorization configs (see aws, gcp, etc)

### Breaking change considerations
The main downside to changing config versions is that it's a breaking change. To mitigate we will:
- Add the new profile-based config along side the old environment-based config
- Deprecate functions using the old config with a warning
- Deprecate top-level functions which use `get_default_client()` implicitly
- An interactive CLI to migrate: `nom config migrate`

## Docs changes

Our docs need to be updated to remove references to `set_base_url` and `set_token`. Instead we should recommend usage of profiles.

- [x] Create docs page on recommended ways to connect via client lib
- [x] Remove references to `get_default_client`, `set_base_url` and `set_token`

### NominalConfigV1
Example v1 config ("environment-based"), stored at `~/.nominal.yml`

```yaml
environments:
  api.gov.nominal.io/api: eyJ...
  api-staging.gov.nominal.io/api: nominal_api_key_...
  api.nominal.test: nominal_api_key_...
```

### NominalConfigV2
Example v2 config ("profile-based"), stored at `~/.config/nominal/config.yml`

```yaml
version: 2
profiles:
  default:
    base_url: https://api.gov.nominal.io/api
    token: eyJ...
  staging:
    base_url: https://api-staging.gov.nominal.io/api
    token: nominal_api_key_...
    workspace_rid: ri.security...
  dev:
    base_url: https://api.nominal.test
    token: nominal_api_key_...
```


alkasm:
I don't love the solution I came up with here. It's too far into the middle of the old and new config and confusing. Have an alternate proposal that I think is simpler and better: we allow users to set the "current profile" instead of setting the "base url." In either case the user just wants to set the connection that top-level Nominal functions use. The current profile will default to the "default" profile if it's not set, which should be analogous to the default base url being set to "api.gov.nominal.io/api".

Mechanically, this means:
- auto-migrate to v2 config
- deprecate `set_base_url` in favor of `set_current_profile`
  - have `set_base_url` find the relevant profile for that url and then `set_current_profile` with it
- deprecate `set_token` in favor of `nom config set-profile`
  - have `set_token` find the relevant profile for that url and then set the token for it
- deprecate `get_default_client` in favor of `get_client(profile: str | None)`
  - have `get_default_client` call `get_client(None)` which will fetch the current profile
  - function is now confusingly named as it does not retrieve the "default" profile, but instead the "current" profile

tests/e2e/conftest.py
```diff
@@ -24,8 +24,8 @@ def base_url(pytestconfig):
 
 
 @pytest.fixture(scope="session")
-def client(base_url, auth_token) -> Iterator[None]:
-    return NominalClient.create(base_url=base_url, token=auth_token)
+def client(base_url, auth_token) -> NominalClient:
+    return NominalClient.from_url(base_url=base_url, token=auth_token)
```
alkasm, line 28:
```suggestion
    return NominalClient.create(base_url=base_url, token=auth_token)
```

nominal/config.py
```diff
@@ -0,0 +1,149 @@
...
+        url, token = env.popitem()
+        profiles[url] = ConfigProfile(base_url=f"https://{url}", token=token)
+        logger.debug(f"creating profile '{url}' from the {url} environment: {profiles[url]}")
+
+    cfg = NominalConfig(version=2, profiles=profiles)
+    logger.info(f"migrating deprecated config to new config: {cfg}")
+    cfg.to_yaml()
+    logger.warning(
+        f"we recommend deleting deprecated config file {DEPRECATED_NOMINAL_CONFIG_PATH} containing {deprecated_cfg}"
+    )
```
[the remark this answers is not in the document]
alkasm: maybe not, if anyone goes back some versions their config would be gone (also I just dont like the idea of deleting things from users computers)

nominal/cli/auth.py
```diff
@@ -43,7 +20,7 @@ def _validate_token_url(token: str, base_url: str) -> None:
 @global_options
 def set_token(token: str, base_url: str) -> None:
```
alkasm, line 21: should deprecate this CLI

nominal/cli/config.py
```diff
@@ -0,0 +1,64 @@
...
+def config_cmd() -> None:
+    pass
+
+
+@config_cmd.command()
+@click.option("-p", "--profile", prompt=True, help="profile name")
+@click.option("-u", "--base-url", default="https://api.gov.nominal.io/api", prompt=True)
+@click.option("-t", "--token", required=True, prompt=True, help="bearer token or api key")
+@global_options
+def add_profile(profile: str, base_url: str, token: str) -> None:
```
alkasm, line 23: i think i want to do `config profile add <name> [opts...]`

nominal/nominal.py
```diff
@@ -1,13 +1,17 @@
 from __future__ import annotations
 
+import dataclasses
```
alkasm, line 3: ?

nominal/nominal.py
```diff
@@ -349,10 +390,12 @@ def search_runs(
...
-    conn = get_default_client()
-    return conn.search_runs(
-        start=start,
-        end=end,
+    if all([v is None for v in (start, end, name_substring, label, property)]):
+        raise ValueError("must provide one of: start, end, name_substring, label, or property")
+    client = get_client()
+    return client.search_runs(
+        start=None if start is None else ts._SecondsNanos.from_flexible(start).to_nanoseconds(),
+        end=None if end is None else ts._SecondsNanos.from_flexible(end).to_nanoseconds(),
```
alkasm, line 398: wait I thought this came from a main merge??

nominal/nominal.py
```diff
@@ -457,8 +500,10 @@ def search_assets(
     Returns:
         All assets which match all of the provided conditions
     """
-    conn = get_default_client()
-    return conn.search_assets(
+    if all([v is None for v in (search_text, label, property)]):
+        raise ValueError("must provide one of: start, end, search_text, label, or property")
```
alkasm, line 504: same with this - from main or not?

nominal/nominal.py
alkasm: oh it's used for replace nvm

nominal/config.py
```diff
@@ -0,0 +1,149 @@
...
+        logger.debug(f"creating profile 'default' from the single environment: {profiles['default']}")
+    # if there's a prod URL, make that the default
+    elif prod_url in env:
+        profiles["default"] = ConfigProfile(base_url=f"https://{prod_url}", token=env.pop(prod_url))
+        logger.debug(f"creating profile 'default' from the {prod_url} environment: {profiles['default']}")
+    # otherwise, not obvious how to migrate
+    else:
+        raise _NominalConfigMigrationError("unable to automatically migrate v1 config to v2: use `nom config migrate`")
+
+    if staging_url in env:
```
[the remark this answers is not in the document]
alkasm: Yeah this was more "if it's already in your config, then ..." so probably mostly for nominal employees. Also I wasn't sure if any customers had tenants in staging or not, though I suspect no

nominal/config.py
```diff
@@ -0,0 +1,149 @@
...
+    For production environments, the typical URL is: https://api.gov.nominal.io/api
+    For staging environments, the typical URL is: https://api-staging.gov.nominal.io/api
+    For local development, the typical URL is: https://api.nominal.test (note the lack of the /api suffix)
+    """
+
+    profiles: Mapping[str, ConfigProfile]
+    version: Literal[2]
+
+    @classmethod
+    def from_yaml(cls, path: Path = DEFAULT_NOMINAL_CONFIG_PATH) -> Self:
```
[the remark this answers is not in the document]
alkasm: I think I'll leave this for a future PR, but I have seen that lib mentioned before in passing. I just looked into it a bit more though and agree, seems pretty nifty for this!

nominal/_config.py
```diff
@@ -45,12 +45,15 @@ def get_token(self, url: str) -> str:
         raise NominalConfigError(f"url {url!r} not found in config: set a token with `nom auth set-token`")
 
 
+NominalConfig = NominalConfigV1
```
stefanv, line 48: Is this used elsewhere, or in scripts, or why is `NominalConfig` retained and set to V1?

nominal/cli/config.py
```diff
@@ -0,0 +1,69 @@
...
+    """Add or update a profile to your Nominal config"""
+    cfg = config.NominalConfig(profiles={}, version=2)
+    try:
+        cfg = config.NominalConfig.from_yaml()
+    except FileNotFoundError:
+        pass
+    validate_token_url(token, base_url)
+    new_cfg = dataclasses.replace(cfg, profiles={**cfg.profiles, profile: config.ConfigProfile(base_url, token)})
+    new_cfg.to_yaml()
+    click.secho(f"Added profile {profile} to {config.DEFAULT_NOMINAL_CONFIG_PATH}", fg="green")
```
stefanv, line 38:
```suggestion
    click.secho(f"Added profile {profile} to {config.DEFAULT_NOMINAL_CONFIG_PATH}", fg="green")
```

Not a big deal, but the messaging here may be a bit confusing for a user who means to update a profile. We do know whether the profile already exists or not, so could message it accordingly.

nominal/cli/config.py
```diff
@@ -0,0 +1,69 @@
...
+@config_cmd.command()
+def migrate() -> None:
+    """Interactively migrate deprecated config at ~/.nominal.yml
+    to new profile-based config at ~/.config/nominal/config.yml
+    """
+    deprecated_cfg = _deprecated_config.NominalConfig.from_yaml()
+    profiles = {}
+    for url, token in deprecated_cfg.environments.items():
+        if click.prompt(f"Add profile for {url}?", default="y", type=bool):
+            name = click.prompt("Profile name")
```
stefanv, line 61: Should we suggest a default? Many users may want only one profile, and wouldn't want to think about it too much.

nominal/cli/config.py
```diff
@@ -0,0 +1,69 @@
...
+            new_url = click.prompt("Base url", default=f"https://{url}")
+            new_token = click.prompt("Token", default=token)
+            if click.prompt("Validate connection?", default="y", type=bool):
```
stefanv, line 64: perhaps `Validate token` or `Validate authentication`?

nominal/cli/util/verify_connection.py
```diff
@@ -0,0 +1,27 @@
...
+    status_code = 200
+    err_msg = ""
+    try:
+        NominalClient.create(base_url, token).get_user()
+    except ConjureHTTPError as err:
+        status_code = err.response.status_code
+    if status_code == 401:
+        err_msg = f"The authorization token may be invalid. Read the docs on how to get a new token: {docs_link}"
+    elif status_code == 404:
+        err_msg = "The base_url may be incorrect. Ensure the url is using the api subdomain (not the app)."
```
stefanv, line 19: What does that mean? "using the api subdomain (not the app)"

nominal/cli/util/verify_connection.py
```diff
@@ -0,0 +1,27 @@
...
+    elif status_code != 200:
+        err_msg = (
+            f"There is likely a misconfiguration between the base_url and token. Ensure you use the api subdomain, "
```
stefanv, line 22: api subdomain is mentioned here too; is this standard nomenclature the user will know?

nominal/config.py
```diff
@@ -0,0 +1,84 @@
...
+        token: eyJ...
+        workspace_rid: ri.security.gov-staging.workspace.82db1f3a-568e-418e-a2d0-0575396f29a2
+      dev:
+        base_url: https://api.nominal.test
+        token: eyJ...
+
+
+    For production environments, the typical URL is: https://api.gov.nominal.io/api
+    For staging environments, the typical URL is: https://api-staging.gov.nominal.io/api
+    For local development, the typical URL is: https://api.nominal.test (note the lack of the /api suffix)
```
stefanv, line 49: Do customers also host, or have hosted versions? In that case, may provide an example of a hosted URL as well.

nominal/core/client.py
```diff
@@ -78,6 +79,52 @@
...
+    ) -> Self:
+        """Create a connection to the Nominal platform from a named profile in the Nominal config.
+
+        Args:
+            profile: profile name in the Nominal config.
+            trust_store_path: path to a trust store certificate chain to initiate SSL connections. If not provided,
+                certifi's trust store is used.
+            connect_timeout: Request connection timeout.
+        """
+        config = NominalConfig.from_yaml()
```
stefanv, line 98: Just to note: it says in several places the config is stored at ~/.config/nominal "by default", but really the assumption is implicit in statements like these.

nominal/core/client.py
```diff
@@ -78,6 +79,52 @@
 class NominalClient:
```
stefanv, line 79: We should consider a better repr for NominalClient that shows the profile or the URL or *something*. Right now it just prints as `NominalClient()`.

nominal/config.py
```diff
@@ -0,0 +1,84 @@
...
+    """
+
+    profiles: Mapping[str, ConfigProfile]
+    version: Literal[2]
```
stefanv, line 53: Do you want to move this up by one line, so that the config file starts with `version: 2`? Right now, the profiles are written out first.

nominal/core/client.py
```diff
@@ -78,6 +79,52 @@
...
+            profile: profile name in the Nominal config.
+            trust_store_path: path to a trust store certificate chain to initiate SSL connections. If not provided,
+                certifi's trust store is used.
+            connect_timeout: Request connection timeout.
+        """
+        config = NominalConfig.from_yaml()
+        prof = config.get_profile(profile)
+        return cls.from_token(
+            prof.base_url,
+            prof.token,
```
stefanv, line 102: This is a bit problematic, given that `from_token` expects the token first ;)

stefanv: I'm going to push a fix for this to the branch.

stefanv (approved):

nominal/_config.py
alkasm: Yeah I mean it was there before so it would be a breaking change to remove it, not possible to know what exists out in the wild. Though was this originally a private module? if so, I guess no reason to keep it

nominal/cli/util/verify_connection.py
alkasm: If you copy/paste the nominal URL from the web UI for e.g., you get `app.gov.nominal.io` whereas the api url is `api.gov.nominal.io`.

How about "ensure the url begins with `api` and not `app`"?

nominal/config.py
alkasm: Yes, they can self-host, but in that case the domain can be totally arbitrary. Still might be worth mentioning! I don't know what is typical here.

nominal/core/client.py
alkasm: I'm not opposed. I'm very used to reprs that actually eval to the real object (like a typical dataclass repr), do you know of any good ways to signify that it's not that kind in the way it's shown? E.g. `NominalClient(profile="asdf")` is kinda weird to show if you can't actually create it with the code `NominalClient(profile="asdf")`.

nominal/core/client.py
alkasm: good catch! I only decided to flip the function args in the signature later since it's annoying to retype the URL each time.

nominal/cli/config.py
alkasm: the suggestion here is empty/no change - what did you mean to suggest instead? makes sense what you're saying

nominal/cli/config.py
alkasm: I'm a bit hesitant to suggest something like `"default"` because I can see it as a potential footgun, but agree that for the majority of customers, it's totally fine bc there's only ever the one workspace

nominal/core/client.py
stefanv: Same for me, but I don't know if I've really made use of that feature, or if other users do.

Some open source libraries, like scikit-learn, stick to it fairly religiously, but even NumPy recently gave up on it because they wanted a way to share the shape of an array (which is derived during construction):

```ipython
In [9]: X = np.random.random((100, 100))

In [10]: X
Out[10]: 
array([[0.19953225, 0.2357203 , 0.16870384, ..., 0.52077498, 0.71906834,
        0.92459678],
       [0.98253433, 0.78320832, 0.63803374, ..., 0.36126438, 0.70788206,
        0.08869907],
       [0.68975851, 0.1055596 , 0.17298201, ..., 0.87927831, 0.6219271 ,
        0.0062746 ],
       ...,
       [0.88448163, 0.46670468, 0.98500573, ..., 0.168099  , 0.18218392,
        0.75530884],
       [0.60641334, 0.45345147, 0.14428632, ..., 0.18279633, 0.76290031,
        0.21706652],
       [0.31683851, 0.04575519, 0.8478641 , ..., 0.41513162, 0.3055292 ,
        0.51935916]], shape=(100, 100))
```

[The Python docs say](https://docs.python.org/3/reference/datamodel.html) that you may use `<...>` under circumstances where you cannot produce a repr that evaluates to the same object (we should note, `NominalClient()` certainly does *not*!).

> Called by the [repr()](https://docs.python.org/3/library/functions.html#repr) built-in function to compute the “official” string representation of an object. If at all possible, this should look like a valid Python expression that could be used to recreate an object with the same value (given an appropriate environment). If this is not possible, a string of the form <...some useful description...> should be returned.

So, perhaps:

```
<NominalClient profile="asdf">
```

nominal/cli/util/verify_connection.py
stefanv: Yes, that's helpful; should we tell them where to find the API URL, or just say: typically, you get that URL by replacing `app` with `api`?

nominal/cli/config.py
stefanv: Something like `Added profile {profile} to {...}` vs `Updated profile {profile} in {...}`, depending on what has been done. Maybe too nitpicky, and can just always say "Updated profile X in F".

stefanv:
I noticed two missing features while using this branch:

1. No way to list profiles (and the `nom` client does not advertise the location of the config file, so users won't know where to go look)
2. No way to specify workspace for profiles. This should be prompted for, even if it can be null.

stefanv:
Docs PR started at: https://github.com/nominal-io/fern-docs/pull/170

Once we agree on the language around profiles there, I will go ahead and replace all instances of `get_default_client` with `NominalClient.from_profile`.

nominal/core/client.py
alkasm: Yeah let's do the angle brackets! I forgot about that convention.

alkasm:
> I noticed two missing features while using this branch:
> 
> 1. No way to list profiles (and the `nom` client does not advertise the location of the config file, so users won't know where to go look)
> 2. No way to specify workspace for profiles. This should be prompted for, even if it can be null.

Good catches! I wrote the CLI command before I added workspaces functionality and forgot to add them to the CLI part.

> We shifted away from the word "token" to "API Key" on the frontend. Should the config also use this naming?

Hmm I get the point on consistency, but I think token is more general. You can use api keys or potentially other tokens here (e.g. a JWT bearer token) so I'm inclined to leave it.

nominal/config.py
alkasm: 🤷🏼 idk if it matters much either way, i think i went back and forth on defaulting this param in the first place

nominal/cli/config.py
alkasm: I'll say "Wrote profile ..." as it's clear I think in both cases

nominal/core/client.py
```diff
@@ -81,6 +82,56 @@
...
+    def from_token(
+        cls,
+        token: str,
+        base_url: str = "https://api.gov.nominal.io/api",
+        *,
+        workspace_rid: str | None = None,
+        trust_store_path: str | None = None,
+        connect_timeout: timedelta | float = timedelta(seconds=30),
+        _profile: str | None = None,
+    ) -> Self:
```
alkasm, line 125: docstring

nominal/core/client.py
```diff
@@ -104,14 +167,22 @@ def create(
         """
         if token is None:
             token = _config.get_token(base_url)
-        trust_store_path = certifi.where() if trust_store_path is None else trust_store_path
-        cfg = ServiceConfiguration(
-            uris=[base_url],
-            security=SslConfiguration(trust_store_path=trust_store_path),
-            connect_timeout=connect_timeout,
+        timeout = connect_timeout if isinstance(connect_timeout, timedelta) else timedelta(seconds=connect_timeout)
```
v-tupuri, line 170: is this necessary? from_token looks like it handles both types

nominal/core/client.py
```diff
@@ -76,26 +77,88 @@
...
+            prof.base_url,
+            workspace_rid=prof.workspace_rid,
+            trust_store_path=trust_store_path,
+            connect_timeout=connect_timeout,
+            _profile=profile,
+        )
+        return client
+
+    @classmethod
+    def from_token(
```
v-tupuri, line 117: should we call `validate_token_url` if users call this directly? or is it somehow being called and I'm missing it

alkasm: that function is purely in the CLI for when you are adding your token to the config, it's not a library function.

nominal/core/client.py
alkasm: oops good point

v-tupuri (approved):

PR #212, opened 2025-02-10
feat: update client to support multi file mcap datasets via new ingest endp…

aarij-nominal:
…oint

### DIFF
--- a/nominal/core/client.py
+++ b/nominal/core/client.py
@@ -41,7 +41,7 @@ from nominal.core.channel import Channel
 from nominal.core.checklist import Checklist, ChecklistBuilder
 from nominal.core.connection import Connection
 from nominal.core.data_review import DataReview, DataReviewBuilder
-from nominal.core.dataset import Dataset, _get_dataset, _get_datasets
+from nominal.core.dataset import Dataset, _create_ingest_request, _create_mcap_channels, _get_dataset, _get_datasets
 from nominal.core.filetype import FileType, FileTypes
 from nominal.core.log import Log, LogSet, _get_log_set
 from nominal.core.run import Run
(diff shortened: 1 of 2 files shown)


nominal/core/client.py
```diff
@@ -67,7 +67,11 @@ class NominalClient:
 
     @classmethod
     def create(
-        cls, base_url: str, token: str | None, trust_store_path: str | None = None, connect_timeout: float = 30
+        cls,
+        base_url: str,
+        token: str | None,
+        trust_store_path: str | None = None,
+        connect_timeout: float = 30,
```
alkasm, line 74: lets revert this edit

nominal/core/client.py
```diff
@@ -726,7 +742,8 @@ def get_channel(self, rid: str) -> Channel:
                 This typically occurs when there is no such channel for the given RID.
         """
         return Channel._from_conjure_logicalseries_api(
-            self._clients, self._clients.logical_series.get_logical_series(self._clients.auth_header, rid)
+            self._clients,
+            self._clients.logical_series.get_logical_series(self._clients.auth_header, rid),
```
alkasm, line 746: my guess is that you have `black` set to autoformat rather than `ruff` and hence isn't picking up the `ruff` options here and formatting differently? should make sure the PR isn't changing other formatting stuff as it gives greater chance for conflicts.

nominal/core/dataset.py
```diff
@@ -209,6 +210,49 @@ def add_to_dataset_from_io(
...
+        if include_topics is not None and exclude_topics is not None:
+            include_topics = [t for t in include_topics if t not in exclude_topics]
+        if include_topics is not None:
+            channels = ingest_api.McapChannels(
+                include=[api.McapChannelLocator(topic=topic) for topic in include_topics]
+            )
+        elif exclude_topics is not None:
+            channels = ingest_api.McapChannels(
+                exclude=[api.McapChannelLocator(topic=topic) for topic in exclude_topics]
+            )
```
alkasm, line 240: nit: lets turn this into a function to reuse?

nominal/core/dataset.py
```diff
@@ -209,6 +210,49 @@ def add_to_dataset_from_io(
...
+
+        request = ingest_api.IngestRequest(
```
alkasm, line 242: might also be able to lump this into a shared function to create the request, parameterized by the target?

alkasm (approved):

PR #213, opened 2025-02-12
feat: rework streaming connections

v-tupuri:


nominal/core/queueing.py
v-tupuri, line 1: I see we process_batches sequentially in mavstream (i think), is that fine still? I think the previous implementation up to 3 at once (via the semaphore) in stream.py @alkasm

alkasm: I don't know if you saw the update that I pushed yesterday, but I updated it so that the write thread actually sends the network requests via a thread pool

nominal/core/client.py
```diff
@@ -796,23 +796,32 @@ def create_streaming_connection(
         datasource_id: str,
         connection_name: str,
         datasource_description: str | None = None,
+        nominal_data_source_rid: str | None = None,
```
alkasm, line 799: datasource vs data_source - let's figure out what the correct casing is and be consistent

nominal/core/client.py
```diff
@@ -796,23 +796,32 @@ def create_streaming_connection(
...
+                    id=datasource_id,
+                    description=datasource_description,
+                ),
+            )
+        else:
+            datasource_response = self._clients.storage.batch_get(self._clients.auth_header, [nominal_data_source_rid])[
+                0
+            ]
+            if datasource_response is None:
+                raise ValueError(f"no datasource found with RID {nominal_data_source_rid!r}")
```
alkasm, line 816: why would this be None? wouldn't it rather index error if there was no datasource found?

nominal/core/connection.py
```diff
@@ -41,21 +42,31 @@ def datasource(self) -> scout_datasource.DataSourceService: ...
...
         def storage_writer(self) -> storage_writer_api.NominalChannelWriterService: ...
 
+        @property
+        def proto_write(self) -> ProtoWriteService: ...
+
     @classmethod
-    def _from_conjure(cls, clients: _Clients, response: scout_datasource_connection_api.Connection) -> Connection:
+    def _from_conjure(
+        cls, clients: _Clients, response: scout_datasource_connection_api.Connection
+    ) -> Connection | NominalStreamingConnection:
```
alkasm, line 55: is the `Nominal` prefix necessary/semantic?

nominal/core/connection.py
```diff
@@ -141,13 +152,104 @@ def get_channel(self, name: str, tags: dict[str, str] | None = None) -> Channel:
...
+        max_wait: timedelta = timedelta(seconds=1),
+        data_format: Literal["json", "protobuf"] = "json",
+        backpressure_mode: BackpressureMode = BackpressureMode.BLOCK,
+        maxsize: int = 0,
+    ) -> WriteStreamV2:
+        """Stream to write non-blocking messages to a datasource.
+
+        Args:
+        ----
+            batch_size (int): How big the batch can get before writing to Nominal. Default 50,000
```
alkasm, line 201: nit: let's name this `max_batch_size` since it's more in line with `max_wait` (while we're here, "wait" is kinda a weird name for what it is)

nominal/core/connection.py
```diff
@@ -21,7 +21,9 @@
 from nominal.core._utils import HasRid
 from nominal.core.batch_processor import process_batch_legacy
 from nominal.core.channel import Channel
-from nominal.core.stream import WriteStream
+from nominal.core.stream import BatchItem, WriteStream
+from nominal.core.stream_v2 import WriteStreamV2
+from nominal.ts import BackpressureMode
```
alkasm, line 26: ts is for "timestamp"-related functions lol

nominal/core/connection.py
```diff
@@ -141,13 +152,104 @@ def get_channel(self, name: str, tags: dict[str, str] | None = None) -> Channel:
...
+                proto_write=self._clients.proto_write,
+            )
+
+    def get_write_stream_v2(
+        self,
+        batch_size: int = 50_000,
+        max_wait: timedelta = timedelta(seconds=1),
+        data_format: Literal["json", "protobuf"] = "json",
+        backpressure_mode: BackpressureMode = BackpressureMode.BLOCK,
+        maxsize: int = 0,
```
alkasm, line 194: nit: these arguments have some mutual dependence (maxsize of 0 implies that backpressure mode doesn't matter)

nominal/core/queueing.py
```diff
@@ -0,0 +1,139 @@
...
+            elif self.mode == BackpressureMode.DROP_OLDEST:
+                while True:
+                    try:
+                        self.put_nowait(item)
+                        break
+                    except Full:
+                        try:
+                            self.get_nowait()
+                        except Empty:
+                            continue
```
alkasm, line 137: This is racy. You need the to hold the cond var to support this. We'd likely have to write our own queue. I have a pypi library [with an evicting queue](https://alkasm.github.io/theta/theta.html#queues) but maybe it'd be best to actually redo this with the same interface that `queue.Queue` has? Also I would need a `task_done()` type of thing to shutdown in the same way.

nominal/core/queueing.py
```diff
@@ -0,0 +1,139 @@
...
+    q.shutdown()
+
+
+class BackpressureQueue(Queue[_T]):
+    """A queue that implements different backpressure strategies."""
+
+    def __init__(self, maxsize: int = 0, mode: BackpressureMode = BackpressureMode.BLOCK):
+        """Initialize the queue with a maximum size and a backpressure mode."""
+        super().__init__(maxsize=maxsize)
+        self.mode = mode
```
alkasm, line 114: generally prefer having different types rather than bifurcated implementations based on some runtime value

nominal/core/connection.py
v-tupuri: lol for some reason my brain associated ts with types

v-tupuri:
nominal/core/queueing.py:8: error: Module "queue" has no attribute "ShutDown"  [attr-defined]
nominal/core/queueing.py:130: error: "DropNewestQueue[_T]" has no attribute "is_shutdown"  [attr-defined]
nominal/core/queueing.py:146: error: "DropOldestQueue[_T]" has no attribute "is_shutdown"  [attr-defined]
yikes our checks are failing because it looks like Shutdown was introduced on python 3.13

@alkasm does it make sense to just add this to our queue class ourselves then

nominal/core/queueing.py
v-tupuri: see fix. not sure between holding the cond var or the mutex

nominal/core/connection.py
v-tupuri: is there a standard clean way to handle this or should I just throw if something doesnt align

nominal/core/batch_processor_proto.py
```diff
@@ -80,9 +80,7 @@ def process_batch(
 
     api_batches = [list(api_batch) for _, api_batch in api_batched]
 
-    if nominal_data_source_rid is None:
-        raise ValueError("Writing not implemented for this connection type")
-
+    request = create_write_request(api_batches)
```
fluffetyfluff, line 83: this is duplicated and literally like 2x slowed down my code

v-tupuri: whoops

nominal/core/connection.py
```diff
@@ -141,13 +151,79 @@ def get_channel(self, name: str, tags: dict[str, str] | None = None) -> Channel:
...
+            process_batch(
+                batch=batch,
+                nominal_data_source_rid=self.nominal_data_source_rid,
+                auth_header=self._clients.auth_header,
+                proto_write=self._clients.proto_write,
+            )
+
+    def get_write_stream_v2(
+        self,
+        max_batch_size: int = 50_000,
```
fluffetyfluff, line 187: let's default this to `30_000` too

fluffetyfluff: ideally we can just plug and play this with full default settings and it just works

nominal/core/stream_v2.py
```diff
@@ -0,0 +1,229 @@
...
+class BackpressureMode(enum.Enum):
+    BLOCK = "block"
+    DROP_NEWEST = "drop_newest"
+    DROP_OLDEST = "drop_oldest"
+
+
+@dataclass()
+class WriteStreamV2:
+    _process_batch: Callable[[Sequence[BatchItem]], None]
+    max_batch_size: int = 50_000
```
fluffetyfluff, line 42: same here

nominal/core/stream_v2.py
```diff
@@ -0,0 +1,229 @@
...
+        cls,
+        nominal_data_source_rid: str,
+        process_batch: Callable[[Sequence[BatchItem]], None],
+        max_batch_size: int = 50_000,
+        max_wait: timedelta = timedelta(seconds=1),
+        max_queue_size: int = 0,
+        backpressure_mode: BackpressureMode = BackpressureMode.BLOCK,
+        client_factory: Callable[[], ProtoWriteService] | None = None,
+        auth_header: str | None = None,
+        max_workers: int = 2,
```
fluffetyfluff, line 63: workers can probably be safely defaulted to 10, i don't think multiprocessing will actually spawn new processes if it doesn't need them

nominal/core/stream_v2.py
```diff
@@ -0,0 +1,229 @@
...
+    _process_thread: threading.Thread | None = field(default=None)
+    _executor: concurrent.futures.Executor | None = field(default=None)
+    backpressure_mode: BackpressureMode = field(default=BackpressureMode.BLOCK)
+
+    @classmethod
+    def create(
+        cls,
+        nominal_data_source_rid: str,
+        process_batch: Callable[[Sequence[BatchItem]], None],
+        max_batch_size: int = 50_000,
```
fluffetyfluff, line 57: same here

fluffetyfluff (commented):
could probably add a v2 test stream test

nominal/core/stream_v2.py
alkasm: this doesn't need a default if the main entrypoint to it already has a default

nominal/core/stream_v2.py
alkasm: ^ correct, the processes are lazily instantiated, but almost certainly it will be able to utilize the max number of processes since the requests are also happening.

The processes spun up will look like this in `ps aux`:
```
.../python -c from multiprocessing.spawn import spawn_main; spawn_main(tracker_fd=11, pipe_handle=19) --multiprocessing-fork
```

nominal/core/stream_v2.py
alkasm: prefer defaults primarily at user interfaces and not in internal code (unless you need multiple behaviors for diff functions)

nominal/core/connection.py
alkasm: Ya I'd either throw ValueError or, in other cases, see if there are good ways to have one or the other behavior but not  both - e.g. you could use a union type

nominal/core/connection.py
```diff
@@ -141,13 +151,78 @@ def get_channel(self, name: str, tags: dict[str, str] | None = None) -> Channel:
...
+    def unarchive(self) -> None:
+        """Unarchive this connection, making it visible in the UI."""
+        self._clients.connection.unarchive_connection(self._clients.auth_header, self.rid)
+
+
+@dataclass(frozen=True)
+class StreamingConnection(Connection):
+    nominal_data_source_rid: str
+
+    def process_batch(self, batch: Sequence[BatchItem], data_format: Literal["json", "protobuf"] = "json") -> None:
```
v-tupuri, line 169: todo - think I can delete this. not used anywhere

nominal/core/batch_processor_proto.py
```diff
@@ -71,25 +74,40 @@ def create_write_request(api_batches: list[list[BatchItem]]) -> WriteRequestNomi
...
+    max_points_per_request = 10000
+    # Group items by channel/tags
     api_batched = groupby(sorted(batch, key=_to_api_batch_key), key=_to_api_batch_key)
-
     api_batches = [list(api_batch) for _, api_batch in api_batched]
 
-    if nominal_data_source_rid is None:
-        raise ValueError("Writing not implemented for this connection type")
+    # Split into smaller requests
+    requests = []
```
v-tupuri, line 90: is there a more optimal way to do this

nominal/core/client.py
```diff
@@ -796,23 +796,33 @@ def create_streaming_connection(
         datasource_id: str,
```
v-tupuri, line 796: this method needs to be cleaner still

nominal/core/queueing.py
```diff
@@ -0,0 +1,111 @@
...
+
+
+def spawn_batching_thread(
+    items: ReadQueue[_T | QueueShutdown],
+    max_batch_size: int,
+    max_batch_duration: timedelta,
+    max_queue_size: int = 0,
+) -> tuple[threading.Thread, ReadQueue[Sequence[_T]]]:
+    """Enqueue items from a queue into batches in a separate thread."""
+    batches = Queue[Sequence[_T]](maxsize=max_queue_size)
```
alkasm, line 74: the type params should be on the annotation not in the evaluated expression

nominal/core/queueing.py
```diff
@@ -0,0 +1,111 @@
...
+        q.task_done()
+
+
+class BackpressureQueue(ABC, Queue[_T]):
+    """Abstract base class for queues with different backpressure strategies."""
+
+    @abstractmethod
+    def put_with_backpressure(self, item: _T) -> None:
+        """Put an item in the queue using the specific backpressure strategy."""
+        pass
```
alkasm, line 104: I think we should just remove all of this stuff and just use a `queue.Queue` for now, we may want to abstract this differently. It's internal-only anyways at this point

nominal/core/stream_v2.py
```diff
@@ -0,0 +1,219 @@
...
+        Args:
+            process_batch: Function to process batches of items
+            max_batch_size: How many items to accumulate before processing
+            max_wait: Maximum time to wait before processing a partial batch
+            max_queue_size: Maximum number of items that can be queued (0 for unlimited)
+            backpressure_mode: How to handle queue overflow:
+                - BLOCK: Block until space is available (default)
+                - DROP_NEWEST: Drop new items when queue is full
+                - DROP_OLDEST: Drop oldest items when queue is full (ring buffer)
+            executor: executor for parallel batch processing.
```
alkasm, line 77: might need to review your docstrings

nominal/core/stream_v2.py
```diff
@@ -0,0 +1,219 @@
...
+            client_factory: Factory function to create ProtoWriteService instances
+            nominal_data_source_rid: Nominal data source rid
+            auth_header: Authentication header
+            max_workers: Maximum number of worker threads for parallel processing
+        """
+        context = WorkerContext(
+            nominal_data_source_rid=nominal_data_source_rid,
+            auth_header=auth_header,
+            proto_write=client_factory(),
```
alkasm, line 86: this is safe? I'd imagine you should probs pass just the base params here and construct the client yourself on the other end, I'm not sure if this would have any bad side-effects as the requests client is stateful

nominal/core/stream_v2.py
```diff
@@ -0,0 +1,219 @@
...
+        )
+
+        executor = ProcessPoolExecutor(max_workers=max_workers, initializer=worker_init, initargs=(context,))
+
+        instance = cls(
```
alkasm, line 91: nit: call this `stream` or something more semantic

nominal/core/stream_v2.py
```diff
@@ -0,0 +1,219 @@
...
+    _process_batch: Callable[[Sequence[BatchItem]], None]
+    max_batch_size: int
+    max_wait: timedelta
+    max_queue_size: int
+    backpressure_mode: BackpressureMode
+    _item_queue: BackpressureQueue[BatchItem | QueueShutdown] = field(init=False)
+    _batch_queue: ReadQueue[Sequence[BatchItem]] | None = field(default=None)
+    _batch_thread: threading.Thread | None = field(default=None)
+    _process_thread: threading.Thread | None = field(default=None)
+    _executor: concurrent.futures.Executor | None = field(default=None)
```
alkasm, line 51: why do these allow `None`?

nominal/nominal.py
```diff
@@ -596,19 +596,29 @@ def upload_mcap_video(
 
 
 def create_streaming_connection(
-    datasource_id: str,
     connection_name: str,
+    datasource_id: str | None = None,
```
alkasm, line 600: moving a positional arg is a breaking change unfortunately, so we'll need to do something different here

nominal/core/batch_processor_proto.py
```diff
@@ -71,25 +74,40 @@ def create_write_request(api_batches: list[list[BatchItem]]) -> WriteRequestNomi
...
+    with concurrent.futures.ThreadPoolExecutor() as executor:
+        futures = [
+            executor.submit(
+                proto_write.write_nominal_batches,
+                auth_header=auth_header,
+                data_source_rid=nominal_data_source_rid,
+                request=request,
+            )
+            for request in requests
+        ]
```
alkasm, line 106: do this in a for loop so you can submit while creating each chunk

alkasm: also I think most things you could keep in an iterable here and don't need to necessarily materialize

nominal/core/batch_processor_proto.py
```diff
@@ -71,25 +74,40 @@ def create_write_request(api_batches: list[list[BatchItem]]) -> WriteRequestNomi
...
+    # Group items by channel/tags
     api_batched = groupby(sorted(batch, key=_to_api_batch_key), key=_to_api_batch_key)
-
     api_batches = [list(api_batch) for _, api_batch in api_batched]
 
-    if nominal_data_source_rid is None:
-        raise ValueError("Writing not implemented for this connection type")
+    # Split into smaller requests
+    requests = []
+    for i in range(0, len(api_batches), max_points_per_request):
```
alkasm, line 91: is this right? I'm confused

nominal/core/multiprocessed_stream.py
```diff
@@ -0,0 +1,190 @@
...
+
+        batch_thread, batch_queue = spawn_batching_thread(
+            item_queue,
+            max_batch_size,
+            max_wait,
+            max_queue_size=batch_maxsize,
+        )
+
+        instance = cls(
+            _serialize_batch=serialize_batch,
```
alkasm, line 83: It'd be nice if we turned this into something like a `Serializer` type that could have a multiprocessing, multithreading, or main thread implementation

nominal/core/multiprocessed_stream.py
```diff
@@ -0,0 +1,190 @@
...
+            _item_queue=item_queue,
+            _batch_thread=batch_thread,
+            _batch_queue=batch_queue,
+            _client_factory=client_factory,
+            _auth_header=auth_header,
+            _nominal_data_source_rid=nominal_data_source_rid,
+        )
+
+        process_thread = threading.Thread(target=instance._process_worker, daemon=True)
+        instance._process_thread = process_thread
```
alkasm, line 98: why is this attribute assigned after the fact?

nominal/core/multiprocessed_stream.py
```diff
@@ -0,0 +1,190 @@
...
+        instance._process_thread.start()
+
+        return instance
+
+    def close(self, wait: bool = True) -> None:
```
alkasm, line 103: is there any reason to not wait?

nominal/core/multiprocessed_stream.py
```diff
@@ -0,0 +1,190 @@
...
+        if self._item_queue:
+            self._item_queue.put(QueueShutdown())
+
+        if wait and self._batch_thread and self._process_thread:
+            self._batch_thread.join()
+            self._process_thread.join()
+
+            self._process_pool.shutdown(wait=True)
+            self._thread_pool.shutdown(wait=True)
+            self._item_queue = Queue()
```
alkasm, line 114: why is the queue reassigned?

nominal/core/__init__.py
```diff
@@ -37,4 +38,5 @@
     "CheckViolation",
     "DataReviewBuilder",
     "WriteStream",
+    "MultiProcessedWriteStream",
```
alkasm, line 41: I think this is "beta" enough that we shouldn't be exporting it just yet.

This would actually be a good candidate for exposing an `experimental` or `beta` subpackage.

nominal/core/multiprocessed_stream.py
v-tupuri: the alternative is to move _process_worker outside the class then right?

nominal/core/connection.py
```diff
@@ -141,12 +151,59 @@ def get_channel(self, name: str, tags: dict[str, str] | None = None) -> Channel:
...
+        max_batch_size: int = 30000,
+        max_wait: timedelta = timedelta(seconds=1),
+        max_queue_size: int = 0,
+        max_workers: int = 10,
+    ) -> MultiProcessedWriteStream:
+        """Stream to write non-blocking messages to a datasource.
+
+        Args:
+        ----
+            max_batch_size (int): How big the batch can get before writing to Nominal. Default 50,000
```
alkasm, line 180: Don't duplicate types and defaults in the docstring - the signature already covers it and it gives too many opportunities for inconsistencies (like this here).

alkasm (approved):
Still needs some later cleanup for clean shutdown, but believe this should be good enough

PR #214, opened 2025-02-12
feat: allow python client to ingest gcs

v-tupuri:

### DIFF
--- a/nominal/core/client.py
+++ b/nominal/core/client.py
@@ -990,6 +990,53 @@ class NominalClient:
         response = self._clients.datareview.get(self._clients.auth_header, rid)
         return DataReview._from_conjure(self._clients, response)
 
+    def create_dataset_from_gcs(
+        self,
+        gcs_path: str,
+        name: str,
+        timestamp_column: str,
+        timestamp_type: _AnyTimestampType,
+        description: str | None = None,
+        *,
+        labels: Sequence[str] = (),
+        properties: Mapping[str, str] | None = None,
+        prefix_tree_delimiter: str | None = None,
+    ) -> Dataset:
+        """Create a dataset from a file in Google Cloud Storage.
+        
+        The file must be one of:
+        * CSV (*.csv)
+        * Compressed CSV (*.csv.gz)
+        * Parquet (*.parquet)
+        * Parquet archives (*.parquet.tar, *.parquet.tar.gz, *.parquet.zip)
+        """
+        request = ingest_api.TriggerFileIngest(
+            source=ingest_api.IngestSource(
+                gcs=ingest_api.GcsIngestSource(path=gcs_path)
+            ),
+            source_metadata=ingest_api.IngestSourceMetadata(
+                timestamp_metadata=ingest_api.TimestampMetadata(
+                    series_name=timestamp_column,
+                    timestamp_type=_to_typed_timestamp_type(timestamp_type)._to_conjure_ingest_api(),
+                ),
+            ),
+            destination=ingest_api.IngestDestination(
+                new_dataset=ingest_api.NewDatasetIngestDestination(
+                    dataset_name=name,
+                    dataset_description=description,
+                    properties={} if properties is None else dict(properties),
+                    labels=list(labels),
+                    channel_config=(
+                        None
+                        if prefix_tree_delimiter is None
+                        else ingest_api.ChannelConfig(prefix_tree_delimiter=prefix_tree_delimiter)
+                    ),
+                )
+            ),
+        )
+        response = self._clients.ingest.trigger_file_ingest(self._clients.auth_header, request)
+        return self.get_dataset(response.dataset_rid)
+
 
 def _create_search_runs_query(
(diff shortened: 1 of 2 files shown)


aarij-nominal:
Me and Varun spoke in person, but the suggestion was instead of creating one dedicated GCS method that can ingest tabular file types, we can just extend all of the other methods to have an optional `source` that can be the GCS path. This allows us to reuse the existing custom methods for all of the different file types.

nominal/core/client.py
```diff
@@ -990,6 +969,50 @@ def get_data_review(self, rid: str) -> DataReview:
         response = self._clients.datareview.get(self._clients.auth_header, rid)
         return DataReview._from_conjure(self._clients, response)
 
+    def _create_source_from_path(self, path: Path | str, name: str, file_type: FileType) -> ingest_api.IngestSource:
```
aarij-nominal, line 972: I like the idea of refactoring the creation of a source out into it's own method. However, using `path` here as a variable with different meanings can be problematic. 

If you're using this library, when you call one of the `create_dataset` functions, you pass in the argument `path`, which is a `Path | str`. There's no way to know that you can also pass in a GCS (or S3 at some point) path. I don't know if it's actually possible, but if a weird file/path started with that string also, we'd wrongly assume it's a GCS path when it's a local file.

In a perfect world, the user could give us an `IngestSource` object, and we put the responsibility of providing an unambiguous source on them, but that's probably too much to ask for a user. I think the ideal situation is we find some sort of happy medium. We need have to have an unambiguous way of interpreting the file source. 

- Maybe we have an optional argument on all these methods called `gcs_path` and we error if the user provides both `path` and `gcs_path`? (doesn't scale well ☹️ )
- Maybe we change the typing on the `path` to be `Path | str | GcsPath`, where `GcsPath` is a new object we create? (Scales a bit better, but can get ugly if we have like 5 different options) 
- Maybe `path` should become generalizable to `path: Path | str | DataSource`. Where `DataSource` is another class that we can define, but we can easily extend it as we introduce more types? (I like something along these lines, but it's a bit of a bigger refactor, and maybe a bit incongruent with the current design patterns 😓).

To be honest, I'm not sure what is the best, and I have a lot of ideas here, but I'm happy to help you making a larger change if you're inclined. 

I'm kinda saying a lot, but the gist of it is 1) this is too hard coded for me 2) there are other ways to do it, and I'm open to ideas.

nominal/core/client.py
```diff
@@ -990,6 +969,50 @@ def get_data_review(self, rid: str) -> DataReview:
...
+        return ingest_api.DatasetIngestTarget(
+            new=ingest_api.NewDatasetIngestDestination(
+                labels=list(labels),
+                properties={} if properties is None else dict(properties),
+                dataset_description=description,
+                dataset_name=name,
+            )
+        )
+
+    def _create_dataset_from_request(self, request: ingest_api.IngestRequest | ingest_api.TriggerIngest) -> Dataset:
```
aarij-nominal, line 1005: I think is an interesting idea to move from the `triggerIngest` endpoint to the `ingest` endpoint, but I wonder if it's easy enough to just move this entire file to just use the `ingest` endpoint (we want to migrate everything there anyways)

aarij-nominal (changes requested):

nominal/core/client.py
alkasm: one other simple option is just another function. That scales kinda poor with all of the `create_blah_dataset` methods, but really I'd rather have all of those be one function with a union argument (as that matches the new APIs).

aarij-nominal: I think this is the best way to go long term. However, what do you think about the whole issue of "ingesting this type of file requires these types of arguments"? Do you think we should start exposing the conjure types like "CsvOps" and on our global "ingest" method, we just have them provide something like that? <br><br>The counter to the above is it kind of becomes a lot of responsibility on the user vs the current library works with super simple primitives (like just give me a string to a path lol). Happy to talk more about this though!

alkasm: Yeah that's what I meant by a union argument that matches the new APIs. It's maybe more pythonic to do separate functions though with semantic kwargs, which matches more closely with what we do now, e.g. `upload_csv(..., skip_lines=...)` rather than `upload_dataset(..., options=CsvOptions(skip_lines=...))` but neither seem that bad.

alkasm: In this case since it's mostly about documentation, what about:

```python
GcsPath: TypeAlias = str
def upload_dataset(path: Path | str | GcsPath, ...): ...
```

We could be more explicit in the types (which I prefer but isn't as user friendly):

```python
@dataclass
class GcsPath:
    uri: str

def upload_dataset(path: Path | str | GcsPath, ...): ...
```

I suppose we could do both

```python
@dataclass
class GcsPath:
    uri: str

def upload_dataset(path: Path | str | GcsPath, ...):
    if isinstance(path, str):
        _upload_dataset_gcs(GcsPath(path), ...)
    if isinstance(path, GcsPath):
        _upload_dataset_gcs(path, ...)
    ...
```

PR #220, opened 2025-02-19
feat: metrics on streaming connections

v-tupuri:
goals are to measure stats abt write stream latency in nominal channels!

1. Point staleness at enqueue time
2. Point staleness at request initiated time (for newest point in batch and oldest point in batch)
3. RTT of the request

corresponding channels

1. enqueue_dict_start_staleness
2. least_recent_before_request_diff/most_recent_before_request_diff
3. rtt

### DIFF
--- a/nominal/core/_clientsbunch.py
+++ b/nominal/core/_clientsbunch.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 from dataclasses import dataclass
-from datetime import datetime, timedelta
+from datetime import datetime, timezone
 from functools import partial
 from typing import Protocol
 
@@ -30,6 +30,8 @@ from nominal_api import (
 )
 from typing_extensions import Self
 
+from nominal.ts import IntegralNanosecondsUTC
+
 
 class ProtoWriteService(Service):
     def write_nominal_batches(
@@ -51,9 +53,9 @@ class ProtoWriteService(Service):
         auth_header: str,
         data_source_rid: str,
         request: bytes,
-        most_recent_timestamp: datetime,
-        least_recent_timestamp: datetime,
-    ) -> tuple[timedelta, timedelta, timedelta, timedelta, timedelta]:
+        most_recent_timestamp: IntegralNanosecondsUTC,
+        least_recent_timestamp: IntegralNanosecondsUTC,
+    ) -> tuple[float, float, float, float, float]:
         _headers = {
             "Accept": "application/json",
             "Content-Type": "application/x-protobuf",
@@ -61,26 +63,25 @@ class ProtoWriteService(Service):
         }
         _path = f"/storage/writer/v1/nominal/{data_source_rid}"
 
-        most_recent_timestamp_diff_in_batch_before_request = timedelta(
-            seconds=datetime.now().timestamp() - most_recent_timestamp.timestamp()
-        )
-        least_recent_timestamp_diff_in_batch_before_request = timedelta(
-            seconds=datetime.now().timestamp() - least_recent_timestamp.timestamp()
-        )
+        current_time_ns = int(datetime.now(timezone.utc).timestamp() * 1e9)
+        most_recent_timestamp_diff_in_batch_before_request = (current_time_ns - most_recent_timestamp) / 1e9
+        least_recent_timestamp_diff_in_batch_before_request = (current_time_ns - least_recent_timestamp) / 1e9
 
-        before_req = datetime.now()
+        before_req = int(datetime.now(timezone.utc).timestamp() * 1e9)
         self._request("POST", self._uri + _path, params={}, headers=_headers, data=request)
-        request_rtt = timedelta(seconds=datetime.now().timestamp() - before_req.timestamp())
 
-        oldest_total_rtt = timedelta(seconds=datetime.now().timestamp() - least_recent_timestamp.timestamp())
-        newest_total_rtt = timedelta(seconds=datetime.now().timestamp() - most_recent_timestamp.timestamp())
+        current_time_ns = int(datetime.now(timezone.utc).timestamp() * 1e9)
+        request_rtt = (current_time_ns - before_req) / 1e9
+
+        largest_e2e_rtt = (current_time_ns - least_recent_timestamp) / 1e9
+        smallest_e2e_rtt = (current_time_ns - most_recent_timestamp) / 1e9
 
         return (
             most_recent_timestamp_diff_in_batch_before_request,
             least_recent_timestamp_diff_in_batch_before_request,
             request_rtt,
-            oldest_total_rtt,
-            newest_total_rtt,
+            largest_e2e_rtt,
+            smallest_e2e_rtt,
         )
 
     def write_prometheus_batches(self, auth_header: str, data_source_rid: str, request: bytes) -> None:
--- a/nominal/core/_clientsbunch.py
+++ b/nominal/core/_clientsbunch.py
@@ -37,8 +37,6 @@ class ProtoWriteService(Service):
         auth_header: str,
         data_source_rid: str,
         request: bytes,
-        most_recent_timestamp: datetime | None = None,
-        least_recent_timestamp: datetime | None = None,
     ) -> None:
         _headers = {
             "Accept": "application/json",
--- a/nominal/core/_clientsbunch.py
+++ b/nominal/core/_clientsbunch.py
@@ -39,7 +39,23 @@ class ProtoWriteService(Service):
         request: bytes,
         most_recent_timestamp: datetime | None = None,
         least_recent_timestamp: datetime | None = None,
-    ) -> tuple[timedelta | None, timedelta | None, timedelta | None, timedelta | None, timedelta | None]:
+    ) -> None:
+        _headers = {
+            "Accept": "application/json",
+            "Content-Type": "application/x-protobuf",
+            "Authorization": auth_header,
+        }
+        _path = f"/storage/writer/v1/nominal/{data_source_rid}"
+        self._request("POST", self._uri + _path, params={}, headers=_headers, data=request)
+
+    def write_nominal_batches_with_metrics(
+        self,
+        auth_header: str,
+        data_source_rid: str,
+        request: bytes,
+        most_recent_timestamp: datetime,
+        least_recent_timestamp: datetime,
+    ) -> tuple[timedelta, timedelta, timedelta, timedelta, timedelta]:
         _headers = {
             "Accept": "application/json",
             "Content-Type": "application/x-protobuf",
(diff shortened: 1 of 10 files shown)


v-tupuri:
I'm not the greatest at typing so I switched off a bunch of the generics in queueing.py so mypy wouldnt hate me. Working on figuring out that stuff now

nominal/core/_clientsbunch.py
```diff
@@ -29,9 +30,16 @@
...
+from nominal.ts import IntegralNanosecondsUTC
+
 
 class ProtoWriteService(Service):
-    def write_nominal_batches(self, auth_header: str, data_source_rid: str, request: bytes) -> None:
+    def write_nominal_batches(
+        self,
+        auth_header: str,
+        data_source_rid: str,
+        request: bytes,
```
alkasm, line 41: super nit: remove the comma so this isn't actually changed in the pr

nominal/core/_clientsbunch.py
```diff
@@ -40,6 +48,42 @@ def write_nominal_batches(self, auth_header: str, data_source_rid: str, request:
...
         self._request("POST", self._uri + _path, params={}, headers=_headers, data=request)
 
+    def write_nominal_batches_with_metrics(
+        self,
+        auth_header: str,
+        data_source_rid: str,
+        request: bytes,
+        oldest_timestamp: IntegralNanosecondsUTC,
+        newest_timestamp: IntegralNanosecondsUTC,
+    ) -> tuple[float, float, float, float, float]:
```
alkasm, line 58: can we do a NamedTuple or dataclass so the values are semantically named

nominal/core/_clientsbunch.py
```diff
@@ -40,6 +48,42 @@ def write_nominal_batches(self, auth_header: str, data_source_rid: str, request:
...
+        _headers = {
+            "Accept": "application/json",
+            "Content-Type": "application/x-protobuf",
+            "Authorization": auth_header,
+        }
+        _path = f"/storage/writer/v1/nominal/{data_source_rid}"
+
+        current_time_ns = int(datetime.now(timezone.utc).timestamp() * 1e9)
```
alkasm, line 66: `time.time_ns()`

nominal/core/_queueing.py
```diff
@@ -22,35 +35,43 @@ def put(self, item: _T_contra) -> None: ...
...
+    newest_timestamp: IntegralNanosecondsUTC = 0
     next_batch_time = time.time() + max_batch_duration.total_seconds()
     while True:
         now = time.time()
         try:
             item = q.get(timeout=max(0, next_batch_time - now))
             if isinstance(item, QueueShutdown):
                 if batch:
-                    yield batch
+                    yield Batch(batch, oldest_timestamp, newest_timestamp)
```
alkasm, line 53: technically this could return `Batch(batch, MAX_INT64, 0)` which would be invalid data

alkasm: actually i guess it can't since it's in `if batch`

wonder if we can make this guarantee a bit more cleanly tho..

nominal/experimental/stream_v2/__init__.py
```diff
@@ -31,6 +32,7 @@ def create_write_stream(
         max_queue_size: Maximum number of items that can be queued (0 for unlimited).
         write_thread_workers: Number of threads to use for writing to Nominal.
         serialize_process_workers: Number of processes to use for serializing batches of protobufs.
+        track_metrics: Whether to track metrics on latency.
```
alkasm, line 35: this doesn't really explain the parameter

nominal/experimental/stream_v2/_serializer.py
```diff
@@ -29,7 +29,7 @@ def create(cls, max_workers: int) -> Self:
         pool = ProcessPoolExecutor(max_workers=max_workers)
         return cls(pool=pool)
 
-    def serialize(self, batch: Sequence[BatchItem]) -> Future[bytes]:
+    def serialize(self, batch: Batch) -> Future[tuple[bytes, IntegralNanosecondsUTC, IntegralNanosecondsUTC]]:
```
alkasm, line 32: same here re: namedtuple or dataclass

nominal/experimental/stream_v2/_write_stream.py
```diff
@@ -116,9 +122,32 @@ def enqueue_from_dict(
...
+        current_time_ns = int(datetime.now(timezone.utc).timestamp() * 1e9)
+        enqueue_dict_timestamp_diff = (current_time_ns - timestamp_normalized) / 1e9
+
         for channel, value in channel_values.items():
             self.enqueue(channel, timestamp, value)
 
+        current_time_ns = int(datetime.now(timezone.utc).timestamp() * 1e9)
+        last_enqueue_timestamp = (current_time_ns - timestamp_normalized) / 1e9
+
+        if self._track_metrics:
```
alkasm, line 135: this value isn't dynamic so it's kinda a bummer to check the condition on every append.

also do you only check this for `enqueue_from_dict`?

nominal/experimental/stream_v2/_write_stream.py
```diff
@@ -136,16 +165,74 @@ def _write_serialized_batch(
...
+                    newest_timestamp_diff_in_batch_before_request,
+                    rtt,
+                    largest_e2e_rtt,
+                    smallest_e2e_rtt,
+                ) = f.result()  # Check for exceptions
+
+                current_time_ns = int(datetime.now(timezone.utc).timestamp() * 1e9)
+                item_queue.put(
+                    BatchItem(
+                        channel_name="metric.oldest_timestamp_diff_in_batch_before_request",
```
alkasm, line 198: preference to use like a double underscore or some other way to disambiguate it from customer data

nominal/ts.py
```diff
@@ -394,6 +394,27 @@ def _time_unit_to_conjure(unit: _LiteralTimeUnit) -> api.TimeUnit:
 )
 
 
+def _normalize_timestamp(timestamp: str | datetime | IntegralNanosecondsUTC) -> IntegralNanosecondsUTC:
```
alkasm, line 397: you should be able to just use `_SecondsNanos.from_flexible(timestamp).to_nanoseconds()`

nominal/core/_clientsbunch.py
```diff
@@ -29,6 +30,21 @@
...
 from typing_extensions import Self
 
+from nominal.ts import IntegralNanosecondsUTC
+
+
+@dataclass(frozen=True)
+class RequestMetrics:
+    oldest_timestamp_diff_before_request: (
+        float  # Time difference between current time and oldest timestamp before request (seconds)
+    )
```
alkasm: you can use docstrings on attributes too instead of inlining the comment, e.g.

```suggestion
    oldest_timestamp_diff_before_request: float
    """Time difference between current time and oldest timestamp before request (seconds)"""
```

nominal/experimental/stream_v2/__init__.py
```diff
@@ -31,7 +32,7 @@ def create_write_stream(
         max_queue_size: Maximum number of items that can be queued (0 for unlimited).
         write_thread_workers: Number of threads to use for writing to Nominal.
         serialize_process_workers: Number of processes to use for serializing batches of protobufs.
-
+        track_metrics: Whether to publish metrics on latency to nominal channels on your connection
```
alkasm, line 35:
```suggestion
        track_metrics: Whether to publish metrics on latency to nominal channels on the connection
```

nominal/experimental/stream_v2/_write_stream.py
```diff
@@ -58,17 +63,38 @@ def create(
...
             _item_queue=item_queue,
             _batch_thread=batch_thread,
             _batch_serialize_thread=batch_serialize_thread,
             _serializer=serializer,
+            _track_metrics=track_metrics,
             _clients=clients,
+            _add_metric=add_metric_fn,
         )
 
+    def add_metric(self, channel_name: str, timestamp: int, value: float) -> None:
```
alkasm, line 94: `int` or `IntegralNanosecondsUTC`?

alkasm: I think I wouldn't expose this method in the public API and just keep the internal `_` method

nominal/experimental/stream_v2/_write_stream.py
```diff
@@ -116,9 +145,19 @@ def enqueue_from_dict(
...
         """
+        timestamp_normalized = _SecondsNanos.from_flexible(timestamp).to_nanoseconds()
+        current_time_ns = time.time_ns()
+        enqueue_dict_timestamp_diff = (current_time_ns - timestamp_normalized) / 1e9
+
         for channel, value in channel_values.items():
             self.enqueue(channel, timestamp, value)
 
+        current_time_ns = time.time_ns()
+        last_enqueue_timestamp = (current_time_ns - timestamp_normalized) / 1e9
```
alkasm, line 156: these 1e9s everywhere seem kinda messy, maybe we should just do the conversion inside `_add_metric`?

alkasm: this also isnt a timestamp but instead a delta right?

nominal/core/_clientsbunch.py
```diff
@@ -40,6 +56,42 @@ def write_nominal_batches(self, auth_header: str, data_source_rid: str, request:
...
+        _path = f"/storage/writer/v1/nominal/{data_source_rid}"
+
+        current_time_ns = time.time_ns()
+        oldest_timestamp_diff_in_batch_before_request = (current_time_ns - oldest_timestamp) / 1e9
+        newest_timestamp_diff_in_batch_before_request = (current_time_ns - newest_timestamp) / 1e9
+
+        before_req = time.time_ns()
+        self._request("POST", self._uri + _path, params={}, headers=_headers, data=request)
+
+        current_time_ns = time.time_ns()
```
alkasm, line 81: `after_resp` maybe to mirror `before_req`?

nominal/core/_clientsbunch.py
```diff
@@ -40,6 +56,42 @@ def write_nominal_batches(self, auth_header: str, data_source_rid: str, request:
...
+        largest_e2e_rtt = (current_time_ns - oldest_timestamp) / 1e9
+        smallest_e2e_rtt = (current_time_ns - newest_timestamp) / 1e9
+
+        return RequestMetrics(
+            oldest_timestamp_diff_before_request=oldest_timestamp_diff_in_batch_before_request,
+            newest_timestamp_diff_before_request=newest_timestamp_diff_in_batch_before_request,
+            request_rtt=request_rtt,
+            largest_e2e_latency=largest_e2e_rtt,
+            smallest_e2e_latency=smallest_e2e_rtt,
+        )
```
alkasm, line 93: this is a _lot_ of lines of code for what it does - this should be cleaned up. for e.g.

```python
before_req = time.time_ns()
self._request("POST", self._uri + _path, params={}, headers=_headers, data=request)
after_response = time.time_ns()

return RequestMetrics.from_nanos(
    oldest_timestamp_diff_before_request=(before_req - oldest_timestamp),
    newest_timestamp_diff_before_request=(before_req - newest_timestamp),
    request_rtt=(after_response - before_req),
    largest_e2e_latency=(after_response - oldest_timestamp),
    smallest_e2e_latency=(after_response - newest_timestamp),
)
```

nominal/experimental/stream_v2/_write_stream.py
```diff
@@ -58,17 +63,39 @@ def create(
             max_queue_size=batch_queue_maxsize,
         )
         batch_serialize_thread = spawn_batch_serialize_thread(
-            write_pool, clients, serializer, nominal_data_source_rid, batch_queue
+            write_pool, clients, serializer, nominal_data_source_rid, batch_queue, item_queue, track_metrics
         )
+
+        def add_metric_impl(channel_name: str, timestamp: IntegralNanosecondsUTC, value: float) -> None:
+            value = value / 1e9
```
alkasm, line 70: eh ok since there's no indication that `value` is a nanosecond timestamp, maybe it makes less sense to do the `/ 1e9` in here (at least if it's going to be named something like `add_metric`). Let's just put it at the call-site?

nominal/core/_clientsbunch.py
```diff
@@ -40,6 +67,34 @@ def write_nominal_batches(self, auth_header: str, data_source_rid: str, request:
...
+    ) -> RequestMetrics:
+        _headers = {
+            "Accept": "application/json",
+            "Content-Type": "application/x-protobuf",
+            "Authorization": auth_header,
+        }
+        _path = f"/storage/writer/v1/nominal/{data_source_rid}"
+        before_req = time.time_ns()
+
+        self._request("POST", self._uri + _path, params={}, headers=_headers, data=request)
```
alkasm, line 86: I guess we can get rid of some of this code if you just call the other impl method?

alkasm (approved):
I think this is looking good to go when you're comfortable

PR #225, opened 2025-02-21
fix: return type of create_streaming_connection and clean .close() and enqueue metric

v-tupuri:
- Throw an exception in `create_streaming_connection` if it does not return a `StreamingConnection`, and enforce the method return type as `StreamingConnection`.
- Use `WriteStreamV2` as a context manager to ensure graceful shutdown.
- `cancel_futures` only on user keyboard interruption; otherwise, allow in-flight futures to complete.
- Fixed the ordering of shutdown and close methods in `WriteStreamV2.close()` to prevent deadlocks.


nominal/experimental/stream_v2/_serializer.py
```diff
@@ -37,4 +37,4 @@ def __enter__(self) -> BatchSerializer:
     def __exit__(
         self, exc_type: Type[BaseException] | None, exc_value: BaseException | None, traceback: TracebackType | None
     ) -> None:
-        self.close()
+        self.close(cancel_futures=exc_type is not None)
```
v-tupuri, line 40: do we need this? does the exit for serializer ever get called? and is it just good to keep for posterity

alkasm: it does not! we should use a ctx manager for it, i.e.

```python
with serializer:
    with stream:
        yield stream
```

v-tupuri: hm, wont the close get called like twice then?

alkasm: Oh, right. Yeah then we can remove the `__exit__` if it's unused

nominal/core/client.py
```diff
@@ -870,7 +870,10 @@ def create_streaming_connection(
                 should_scrape=True,
             ),
         )
-        return Connection._from_conjure(self._clients, connection_response)
+        conn = Connection._from_conjure(self._clients, connection_response)
+        if isinstance(conn, StreamingConnection):
+            return conn
+        raise TypeError(f"Expected StreamingConnection but got {type(conn).__name__}")
```
alkasm, line 876: TypeError and ValueError are similar in that they are primarily used for argument validation, e.g.

```python
>>> "a" + 5
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: can only concatenate str (not "int") to str
```

alkasm: We probably should just wrap everything in the client lib in a `NominalError` or something

v-tupuri: separate pr?

alkasm: sure, but this shouldn't be a `NominalError` or an `Exception` or something, not a type error

PR #228, opened 2025-02-21
fix: make metrics manager impl cleaner and fix some metrics calc

v-tupuri:


nominal/experimental/stream_v2/_write_stream.py
```diff
@@ -24,6 +24,36 @@
 logger = logging.getLogger(__name__)
 
 
+@dataclass()
```
v-tupuri, line 27: hm doing the no op methods here is a little tricky bc of the count variable. ill think a little more abt it

v-tupuri: I can work around it with like .setattr but is that too messy

PR #232, opened 2025-02-25
feat: bump client for legacy ingest response

ritwikdixit:
This should add back support for `asyncHandle`, a deprecated/optional field returned from legacy ingest endpoints.


v-tupuri:
good to close in favor of drake's pr to unify ingest endpoint?

PR #233, opened 2025-02-25
feat: use ingest V2 endpoint everywhere in python client

drake-nominal:

### DIFF
--- a/nominal/core/client.py
+++ b/nominal/core/client.py
@@ -376,24 +376,37 @@ class NominalClient:
 
         file_type = FileType(*file_type)
         s3_path = upload_multipart_io(self._clients.auth_header, dataset, name, file_type, self._clients.upload)
-        request = ingest_api.TriggerIngest(
-            labels=list(labels),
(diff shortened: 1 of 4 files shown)


ritwikdixit (commented):
This lgtm but I'll let @alkasm approve, I don't know if we have E2E ingest tests

aarij-nominal (approved):

nominal/core/client.py
```diff
@@ -376,24 +376,37 @@ def create_dataset_from_io(
...
+                            dataset_description=description,
+                            dataset_name=name,
+                        )
+                    ),
+                    timestamp_metadata=ingest_api.TimestampMetadata(
+                        series_name=timestamp_column,
+                        timestamp_type=_to_typed_timestamp_type(timestamp_type)._to_conjure_ingest_api(),
+                    ),
+                    additional_file_tags=None,
+                    channel_prefix=prefix_tree_delimiter,
```
alkasm, line 401: this is a bug - this is not the delimiter but a string to prefix all channel names by!

PR #234, opened 2025-02-25
feat: expose channel update() method

drake-nominal:
Provide an `update()` method on a `nominal.Channel` such that users may set the unit and description on a channel object.

### DIFF
--- a/nominal/core/channel.py
+++ b/nominal/core/channel.py
@@ -20,6 +20,7 @@ from nominal_api.api import Timestamp
 from typing_extensions import Self
 
 from nominal.core._clientsbunch import HasAuthHeader
+from nominal.core._utils import update_dataclass
 from nominal.ts import _MAX_TIMESTAMP, _MIN_TIMESTAMP, IntegralNanosecondsUTC, _SecondsNanos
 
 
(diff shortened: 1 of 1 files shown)


nominal/core/channel.py
```diff
@@ -55,6 +55,14 @@ def logical_series(self) -> timeseries_logicalseries.LogicalSeriesService: ...
         @property
         def compute(self) -> scout_compute_api.ComputeService: ...
 
+    def set_description(self, description: str) -> None:
```
v-tupuri, line 58: docstring 😄

alkasm:
is implementing the full `update()` blocked on anything?

nominal/core/channel.py
stefanv: I suspect users will try `chan.description = "foo"`.

To prevent this from happening, it may be wise to hide `rid`, `name` etc. as `_rid`, `_name` etc., and then expose them via getters & setters.

[Example from the Python docs](https://docs.python.org/3/library/functions.html#property.deleter)

nominal/core/channel.py
```diff
@@ -55,6 +55,14 @@ def logical_series(self) -> timeseries_logicalseries.LogicalSeriesService: ...
...
+        self._clients.logical_series.batch_update_logical_series(
```
stefanv, line 59: Does this sync back the `.description` to the local object?

nominal/core/channel.py
```diff
@@ -58,9 +59,38 @@ class _Clients(HasAuthHeader, Protocol):
...
+        Args:
+            description: Human-readable description of data within the channel
+            unit: Unit symbol to apply to the channel
+        """
+        channel_metadata = self._clients.channel_metadata.update_channel_metadata(
+            self._clients.auth_header,
+            timeseries_channelmetadata_api.UpdateChannelMetadataRequest(
+                channel_identifier=timeseries_channelmetadata_api.ChannelIdentifier(
+                    channel_name=self.name,
+                    data_source_rid=self.data_source,
```
v-tupuri, line 85: if no description is provided, will it make it None?

v-tupuri (approved):

v-tupuri:
maybe throw a deprecation warning on set_description now?

PR #235, opened 2025-02-25
feat: expose run `asset` field

joseph-gio:

### DIFF
--- a/nominal/core/run.py
+++ b/nominal/core/run.py
@@ -35,8 +35,7 @@ class Run(HasRid):
     start: IntegralNanosecondsUTC
     end: IntegralNanosecondsUTC | None
     run_number: int
-    asset: str | None
-    """The RID of the asset representing this run."""
+    assets: list[str]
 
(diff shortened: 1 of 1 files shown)


nominal/core/run.py
```diff
@@ -35,6 +35,8 @@ class Run(HasRid):
     start: IntegralNanosecondsUTC
     end: IntegralNanosecondsUTC | None
     run_number: int
+    asset: str | None
+    """The RID of the asset representing this run."""
```
joseph-gio, line 39: note: not sure if this field should be `str | None` or just `str`. Supposedly every run should have this field, but in the underlying generated API this field is `str | None`

v-tupuri (approved):

v-tupuri:
this one looks good to me - approved. just double check that assets is not optional on the conjure type

alkasm:
> this one looks good to me - approved. just double check that assets is not optional on the conjure type ( i dont think it is)

that's what static type checks are for

PR #236, opened 2025-02-25
feat: create, get, search events

joseph-gio:


nominal/core/client.py
```diff
@@ -963,6 +965,36 @@ def get_data_review(self, rid: str) -> DataReview:
...
+    def create_event(
+        self,
+        name: str,
+        asset_rids: list[str],
+        start: datetime | IntegralNanosecondsUTC,
+        duration: timedelta,
+        type: EventType,
+        *,
+        properties: Mapping[str, str] = {},
+        labels: list[str] = [],
```
alkasm, line 977: https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments

You should be able to copy the pattern used in other examples in this file

nominal/core/client.py
```diff
@@ -963,6 +965,36 @@ def get_data_review(self, rid: str) -> DataReview:
...
+                duration=_to_api_duration(duration),
+                origins=[],
+                properties=dict(properties),
+                labels=labels,
+                type=type._to_api_event_type(),
+            ),
+        )
+        return Event._from_conjure(self._clients, response)
+
+    def get_events(self, uuids: list[str]) -> list[Event]:
```
alkasm, line 994: does this need UUIDs or RIDs? Generally UUIDs shouldn't be user-facing at all, only RIDs afaik

alkasm: Also, we should make less guarantees about the return type and allow a wider arg type (i.e., we don't need a list in particular, so we shouldn't ask for one)

joseph-gio: Events don't appear to have RIDs: https://docs.nominal.io/api-reference/event/get-events

nominal/core/client.py
joseph-gio: I don't think this argument is being appended to?

v-tupuri: yea idt it should cause any issues but might be better to stick to the existing pattern. will defer to @alkasm

alkasm: In general this is something every linter would not permit and there's no reason to use a mutable default. Defaults should always be immutable; hence it's common to use an empty tuple. Technically for dicts you could use a `MappingProxyType` but no-one really does that (probably because they're not that old in python history), it's just way more common to use `Mapping[KT, VT] | None = None`

nominal/core/client.py
```diff
@@ -963,6 +965,36 @@ def get_data_review(self, rid: str) -> DataReview:
         response = self._clients.datareview.get(self._clients.auth_header, rid)
         return DataReview._from_conjure(self._clients, response)
 
+    def create_event(
+        self,
+        name: str,
+        asset_rids: Sequence[str],
+        start: datetime | IntegralNanosecondsUTC,
```
alkasm, line 972: can we use the same list of timestamp types we use everywhere else?

nominal/core/client.py
```diff
@@ -963,6 +965,36 @@ def get_data_review(self, rid: str) -> DataReview:
...
+        duration: timedelta,
```
alkasm, line 973: I wonder if it's more or less useful to default this to a duration of 0? that's kinda what the front-end does

nominal/core/client.py
```diff
@@ -963,6 +965,36 @@ def get_data_review(self, rid: str) -> DataReview:
         response = self._clients.datareview.get(self._clients.auth_header, rid)
         return DataReview._from_conjure(self._clients, response)
 
+    def create_event(
+        self,
+        name: str,
+        asset_rids: Sequence[str],
```
alkasm, line 971: Don't need to change anything here, but just noting this is another example of what @stefanv previously pointed out with the annoying fact that a `str` _is a_ `Sequence[str]` but is generally invalid as input - we should probably holistically check that across the whole library in a diff PR

nominal/core/event.py
```diff
@@ -0,0 +1,134 @@
...
+    def _from_conjure(cls, clients: _Clients, event: event.Event) -> Self:
+        return cls(
+            uuid=event.uuid,
+            asset_rids=tuple(event.asset_rids),
+            name=event.name,
+            start=_SecondsNanos.from_api(event.timestamp).to_nanoseconds(),
+            duration=timedelta(
+                seconds=event.duration.seconds,
+                microseconds=event.timestamp.nanos / 1e3
+                + (0 if event.timestamp.picos is None else event.timestamp.picos / 1e6),
```
alkasm, line 97: the nanos and picos precision would be lost here, it's immediately truncated to integral precision:

```py
>>> timedelta(microseconds=5.5)
datetime.timedelta(microseconds=6)
```

Elsewhere in the library we use `IntegralNanosecondsUTC` for API responses for an unamiguously typed value (because similarly `datetime` only supports microsecond precision), maybe we should do similar here with an `IntegralNanosecondsDuration` or something? We don't meaningfully support picos yet in the client lib, so I think we could utilize something like this for now and later would need to make the change across the whole lib to utilize picos.

(so for now maybe just throw or warn if picos is not None?)

alkasm: fwiw it would probably be fine to use `timedelta | IntegralNanosecondsDuration` or similar for _inputs_, but when retrieving from the API I don't want to truncate the nanos

nominal/core/client.py
joseph-gio: can you be more specific? the library seems to alternately use `datetime | IntegralNanosecondsUTC` (as above) or `str | datetime | IntegralNanosecondsUTC`

alkasm: ah you're right, I think I initially aimed for str support only in the top-level functions and not the core, but it's probably diffused ovver time

nominal/core/client.py
```diff
@@ -1030,6 +1033,36 @@ def get_data_review(self, rid: str) -> DataReview:
         response = self._clients.datareview.get(self._clients.auth_header, rid)
         return DataReview._from_conjure(self._clients, response)
 
+    def create_event(
+        self,
+        name: str,
+        type: EventType,
+        start: str | datetime | IntegralNanosecondsUTC,
```
alkasm, line 1040:
```suggestion
        start: datetime | IntegralNanosecondsUTC,
```

alkasm: keeping the types more restricted in core

nominal/core/event.py
```diff
@@ -0,0 +1,142 @@
...
+from nominal.core.checklist import _to_api_duration
+from nominal.core.connection import Connection
+from nominal.core.dataset import Dataset
+from nominal.core.log import LogSet
+from nominal.core.video import Video
+from nominal.ts import IntegralNanosecondsDuration, IntegralNanosecondsUTC, _SecondsNanos
+
+
+@dataclass(frozen=True)
+class Event(HasRid):
```
alkasm, line 27: this doesn't have a rid

nominal/core/event.py
```diff
@@ -0,0 +1,142 @@
...
+        e = self.__class__._from_conjure(self._clients, response)
+        update_dataclass(self, e, fields=self.__dataclass_fields__)
+        return self
+
+    @classmethod
+    def _from_conjure(cls, clients: _Clients, event: event.Event) -> Self:
+        if event.duration.picos:
+            warnings.warn(
+                f"event '{event.name}' ({event.uuid}) has a duration specified in picoseconds: "
+                "this is not currently supported in nominal-client",
```
alkasm, line 96: let's just be explicit in the warning message that it's ignored in particular

nominal/core/event.py
alkasm: this passed type checks??

stefanv: It's rather frustrating, but when you subclass a protocol without *following* the protocol, then you're implicitly creating a new protocol: which is what `Event` is, in this case.

mypy won't catch that error until your code tries to instantiate Event, which the python-client never does.

pyright (`pip install basedpyright`) will warn you, however, and overall it may be worth considering switching to that tool (although it will be a bit of work; its report is fairly long, atm).

Example:

```python
from typing import Protocol

class HasRid(Protocol):
    rid: str

class X(HasRid):
    name : str = ""

# x = X()  # will fail, even under mypy
```

```
$ mypy foo.py 
Success: no issues found in 1 source file
```

```
$ basedpyright foo.py
/tmp/foo.py
  /tmp/foo.py:6:7 - warning: Class "X" is implicitly a `Protocol` because it extends a `Protocol` without implementing all abstract symbols. If this is intentional, add `Protocol` or `ABC` to its base classes, or use `metaclass=ABCMeta`.
    "HasRid.rid" is not implemented (reportImplicitAbstractClass)
0 errors, 1 warning, 0 notes
```

Uncommenting the last line:

```
$ mypy foo.py 
foo.py:9: error: Cannot instantiate abstract class "X" with abstract attribute "rid"  [abstract]
Found 1 error in 1 file (checked 1 source file)
```
```

stefanv: (But, yes, the Event should define `rid`.)

alkasm: > It's rather frustrating, but when you subclass a protocol without following the protocol, then you're implicitly creating a new protocol: which is what Event is, in this case.

Is that correct? In https://peps.python.org/pep-0544/#merging-and-extending-protocols it states
> Subclassing a protocol class would not turn the subclass into a protocol unless it also has typing.Protocol as an explicit base class. Without this base, the class is “downgraded” to a regular ABC that cannot be used with structural subtyping.

and it seems like this subclassing usage should be checked, from https://peps.python.org/pep-0544/#explicitly-declaring-implementation
> To explicitly declare that a certain class implements a given protocol, it can be used as a regular base class. In this case a class could use default implementations of protocol members. Static analysis tools are expected to automatically detect that a class implements a given protocol. So while it’s possible to subclass a protocol explicitly, it’s not necessary to do so for the sake of type-checking.

Either way...I guess I'm expecting this to behave like an `abc.ABC` so maybe we should use that, or we should simply just use the protocol fully structurally and not subclass.

alkasm: I guess these links are from the PEP and not from the docs of the actually implemented thing

stefanv: Yes, not 100% sure, but I think your expectation, that the type checker should have caught this, is correct. The docs seem to state as much:

https://typing.readthedocs.io/en/latest/spec/protocol.html#explicitly-declaring-implementation

alkasm (approved):

PR #242, opened 2025-02-27
feat: expose more parameters in top-level `create_run`

joseph-gio:
I noticed that the free function for `create_run` doesn't allow specifying `properties`, `labels`, or `attachments`, although the method defined on the client object does, which means that users have to call `get_default_client` if they want to create a run with any of these values.

### DIFF
--- a/nominal/core/client.py
+++ b/nominal/core/client.py
@@ -94,8 +94,8 @@ class NominalClient:
     def create_run(
         self,
         name: str,
-        start: str | datetime | IntegralNanosecondsUTC,
-        end: str | datetime | IntegralNanosecondsUTC | None,
+        start: datetime | IntegralNanosecondsUTC,
+        end: datetime | IntegralNanosecondsUTC | None,
         description: str | None = None,
         *,
         properties: Mapping[str, str] | None = None,
--- a/nominal/core/client.py
+++ b/nominal/core/client.py
@@ -94,8 +94,8 @@ class NominalClient:
     def create_run(
         self,
         name: str,
-        start: datetime | IntegralNanosecondsUTC,
-        end: datetime | IntegralNanosecondsUTC | None,
+        start: str | datetime | IntegralNanosecondsUTC,
+        end: str | datetime | IntegralNanosecondsUTC | None,
         description: str | None = None,
         *,
         properties: Mapping[str, str] | None = None,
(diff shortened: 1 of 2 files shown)


nominal/nominal.py
```diff
@@ -269,6 +269,10 @@ def create_run(
     start: datetime | str | ts.IntegralNanosecondsUTC,
     end: datetime | str | ts.IntegralNanosecondsUTC | None,
     description: str | None = None,
+    *,
+    properties: Mapping[str, str] | None = None,
+    labels: Sequence[str] = (),
+    attachments: Iterable[Attachment] | Iterable[str] = (),
```
joseph-gio, line 275: I didn't add the `asset` parameter even though it's defined on the client method, since there's been talks of potentially changing how runs are associated with assets

nominal/nominal.py
```diff
@@ -279,9 +283,12 @@ def create_run(
     conn = get_default_client()
     return conn.create_run(
         name,
-        start=ts._SecondsNanos.from_flexible(start).to_nanoseconds(),
-        end=None if end is None else ts._SecondsNanos.from_flexible(end).to_nanoseconds(),
+        start=start,
+        end=end,
```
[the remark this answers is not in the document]
v-tupuri: hm, in conn.create_run we dont do the to_nanoseconds right?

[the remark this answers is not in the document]
v-tupuri: nvm. it shouldnt matter

v-tupuri (approved):

nominal/core/client.py
```diff
@@ -94,8 +94,8 @@ def get_user(self) -> User:
     def create_run(
         self,
         name: str,
-        start: datetime | IntegralNanosecondsUTC,
-        end: datetime | IntegralNanosecondsUTC | None,
+        start: str | datetime | IntegralNanosecondsUTC,
+        end: str | datetime | IntegralNanosecondsUTC | None,
```
alkasm, line 98:
```suggestion
        start: datetime | IntegralNanosecondsUTC,
        end: datetime | IntegralNanosecondsUTC | None,
```

nominal/nominal.py
alkasm, line 287:
```suggestion
        start=ts._SecondsNanos.from_flexible(start).to_nanoseconds(),
        end=None if end is None else ts._SecondsNanos.from_flexible(end).to_nanoseconds(),
```

alkasm (approved):

nominal/nominal.py
alkasm: keeping this so the types remain a bit more restricted in core.py, which is intentional

PR #248, opened 2025-03-04
feat: use new channel metadata endpoints

v-tupuri:
- Support `Channel` creation from both old (logical series) and new (series archetype) data formats.
- Deprecate the `rid `property on `Channel`.
- Remove usage of `rid` in `Channel` methods.
- Create a `DataSource` class as a parent to `Connection` and `Dataset`, consolidating overlapping methods.
- Implement methods on `DataSource` using the new `ChannelMetadata` API spec.
- Deprecate `get_channel(rid)` and `set_channel_units` on the client in favor of implementation on `DataSource`
- ensure backcompat using custom decorators


nominal/core/dataset.py
```diff
@@ -245,6 +245,66 @@ def _from_conjure(cls, clients: DataSource._Clients, dataset: scout_catalog.Enri
...
+    )
+    def get_channels(
+        self,
+        exact_match: Sequence[str] = (),
+        fuzzy_search_text: str = "",
+        *,
+        channel_names: list[str] | None = None,
+    ) -> Iterable[Channel]:
+        """Look up the metadata for all matching channels associated with this dataset.
+        NOTE: Provided channels may also be associated with other datasets-- use with caution.
```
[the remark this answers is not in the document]
v-tupuri: oh yea, I did this to temporarily satisfy mypu otherwise it was made abt type signature of get_channels on dataset and datasource being diff

nominal/_utils.py
```diff
@@ -64,3 +64,86 @@ def wrapper(*args: Param.args, **kwargs: Param.kwargs) -> T:
...
+            
+            # If we're here, none of the deprecated args were used
+            # Check if the new kwarg is provided
+            if new_kwarg in kwargs:
+                # Call the parent class method with the new kwarg
+                if is_instance_method and len(args) > 0:
+                    instance_or_cls = args[0]
+                    parent_method = getattr(super(instance_or_cls.__class__, instance_or_cls), fallback_method_name)
+                    # Call the parent method with just the new kwarg
+                    return cast(T, parent_method(**{new_kwarg: kwargs[new_kwarg]}))
```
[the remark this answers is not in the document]
v-tupuri: made a change - will resolve after we discuss irl tmrw

nominal/core/datasource.py
```diff
@@ -0,0 +1,360 @@
+from __future__ import annotations
+
+import logging
+from dataclasses import dataclass, field
+from datetime import datetime
+from typing import Iterable, Mapping, Protocol, Sequence
+
+import pandas as pd
```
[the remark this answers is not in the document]
v-tupuri: isnt it the return type of our to_pandas method? where else could we import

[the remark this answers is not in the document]
alkasm:
```py
if TYPE_CHECKING:
```

nominal/core/dataset.py
```diff
@@ -2,38 +2,26 @@
...
+from typing import BinaryIO, Iterable, Mapping, Sequence
+
+from nominal_api import api, datasource_api, ingest_api, scout_catalog
 from typing_extensions import Self
 
-from nominal.core._clientsbunch import HasAuthHeader
-from nominal.core._conjure_utils import _available_units, _build_unit_update
+from nominal._utils import (
+    deprecate_arguments,
+)
```
alkasm, line 17:
```suggestion
from nominal._utils import deprecate_arguments
```

nominal/core/datasource.py
```diff
@@ -0,0 +1,361 @@
...
+            Yields a sequence of channel metadata objects which match the provided query parameters
+
+        """
+        if not names:
+            names = [channel.name for channel in self.search_channels()]
+
+        # Process in batches of 500
+        batch_size = 500
+        for i in range(0, len(names), batch_size):
+            batch_channel_names = names[i : i + batch_size]
```
alkasm, line 97: In general, you rarely need to actually slice with indexing. It's not a biggie really but it means that you need to require a sequence rather than a generic iterable. It's more flexible for consumers and for the future if you just pin yourself to _always_ allow a general iterable unless there's no other way around it.

Here I'd recommend copying the `itertools.batched` recommendation into a utils.py file - https://docs.python.org/3/library/itertools.html#itertools.batched

(copying because `batched` is only avail on 3.12+)

nominal/core/datasource.py
```diff
@@ -0,0 +1,361 @@
...
+    ) -> Iterable[Channel]:
+        """Look up channels associated with a datasource.
+
+        Args:
+            exact_match: Filter the returned channels to those whose names match all provided strings
+                (case insensitive).
+            fuzzy_search_text: Filters the returned channels to those whose names fuzzily match the provided string.
+            tags: Dictionary of tags to filter channels by
+            start: The minimum data updated time to filter channels by
+            end: The maximum data start time to filter channels by
```
alkasm, line 126: .

nominal/core/datasource.py
```diff
@@ -0,0 +1,361 @@
...
+
+        Example:
+        -------
+        ```
+        import nominal as nm
+
+        rid = "..." # Taken from the UI or via the SDK
+        dataset = nm.get_dataset(rid)
+        s = dataset.to_pandas()
+        print("index:", s.index, "index mean:", s.index.mean())
```
alkasm, line 184: should use `df` rather than `s` (probably copied over from the channel one where the result is a `pd.Series` rather than a `pd.DataFrame`). Also `index` and `index.mean()` seem like kinda weird examples

nominal/core/datasource.py
```diff
@@ -0,0 +1,361 @@
...
+            )
+        )
+
+        # print("filtered channels", filtered_channels)
+
+        batch_size = 20
+        all_dataframes = []
+
+        for i in range(0, len(filtered_channels), batch_size):
+            batch_channels = filtered_channels[i : i + batch_size]
```
alkasm, line 204: batch here as well

alkasm: then you don't need to materialize `filtered_channels` into a list either

nominal/core/datasource.py
```diff
@@ -0,0 +1,361 @@
...
+        Raises:
+        ------
+            ValueError: Unsupported unit symbol provided
+            conjure_python_client.ConjureHTTPError: Error completing requests.
+        """
+        # Get the set of all available unit symbols
+        supported_symbols = set(
+            [unit.symbol for unit in _available_units(self._clients.auth_header, self._clients.units)]
+        )
+        print("channels_to_units", type(channels_to_units))
```
alkasm, line 310: .

nominal/_utils.py
```diff
@@ -64,3 +119,53 @@ def wrapper(*args: Param.args, **kwargs: Param.kwargs) -> T:
...
+) -> Callable[[Callable[Param, T]], Callable[Param, T]]:
+    """Decorator to deprecate specific positional and keyword arguments in favor of a keyword-only argument.
+
+    This decorator handles the case where a method has arguments that are being deprecated
+    in favor of a keyword-only argument. If any deprecated arguments are provided (either as
+    positional or keyword arguments), it will:
+    1. Issue a warning
+    2. Execute the original method (which contains the legacy logic)
+    3. If no deprecated arguments are provided but the new keyword argument is,
+       it will call the new_route method with the new keyword argument.
```
alkasm, line 133:
```suggestion
       it will call new_method with the new keyword argument.
```

nominal/_utils.py
```diff
@@ -44,6 +44,61 @@ def reader_writer() -> Iterator[tuple[BinaryIO, BinaryIO]]:
...
+        from functools import wraps
+        from typing import cast
+
+        sig = inspect.signature(func)
+        param_names = list(sig.parameters.keys())
+
+        @wraps(func)
+        def wrapper(*args: Param.args, **kwargs: Param.kwargs) -> T:
+            # Check if deprecated argument is in kwargs
+            if argument_name in kwargs:
```
alkasm, line 70: let's remove these redundant comments

nominal/_utils.py
```diff
@@ -44,6 +44,61 @@ def reader_writer() -> Iterator[tuple[BinaryIO, BinaryIO]]:
...
+                warnings.warn(
+                    warning_message,
+                    UserWarning,
+                    stacklevel=2,
+                )
```
alkasm, line 75: almost certainly this indent isn't needed anymore

nominal/_utils.py
```diff
@@ -44,6 +44,61 @@ def reader_writer() -> Iterator[tuple[BinaryIO, BinaryIO]]:
...
+                filtered_kwargs.pop(argument_name)
+                # Cast to satisfy the type checker
+                return func(*args, **cast(Param.kwargs, filtered_kwargs))
+
+            elif len(args) > len(param_names) - 1:
+                warnings.warn(
+                    warning_message,
+                    UserWarning,
+                    stacklevel=2,
+                )
```
alkasm, line 87: .

nominal/core/client.py
```diff
@@ -749,11 +749,17 @@ def get_channel(self, rid: str) -> Channel:
...
+            UserWarning,
+        )
         return Channel._from_conjure_logicalseries_api(
             self._clients, self._clients.logical_series.get_logical_series(self._clients.auth_header, rid)
         )
 
-    def set_channel_units(self, rids_to_types: Mapping[str, str | None]) -> Sequence[Channel]:
+    def set_channel_units(
+        self, rids_to_types: Mapping[str, str | None]
+    ) -> Iterable[Channel]:  # TODO(vtupuri): do this correctly
```
alkasm, line 762: todo?

nominal/core/datasource.py
```diff
@@ -0,0 +1,352 @@
...
+        if not all_dataframes:
+            logger.warning(f"No data found for export from datasource {self.rid}")
+            raise RuntimeError(f"No data found for export from datasource {self.rid}")
+
+        result_df = pd.concat(all_dataframes, axis=0)
+        return result_df
+
+    def _construct_export_request(
+        self, channels: Sequence[Channel], start: api.Timestamp, end: api.Timestamp, tags: dict[str, str] | None
+    ) -> scout_dataexport_api.ExportDataRequest:
```
alkasm, line 215: can you move these sort of utility functions out of the class? these should really be free functions, the class doesn't need to be so long

alkasm: same with some of the new channel private methods

nominal/_utils.py
```diff
@@ -44,6 +47,47 @@ def reader_writer() -> Iterator[tuple[BinaryIO, BinaryIO]]:
...
+        sig = inspect.signature(func)
+        param_names = list(sig.parameters.keys())
+
+        @wraps(func)
+        def wrapper(*args: Param.args, **kwargs: Param.kwargs) -> T:
+            if argument_name in kwargs:
+                warnings.warn(warning_message, UserWarning, stacklevel=2)
+                filtered_kwargs = kwargs.copy()
+                filtered_kwargs.pop(argument_name)
+                return func(*args, **cast(Param.kwargs, filtered_kwargs))
```
alkasm, line 73: technically this cast is wrong, but I'm not sure it's important...either way, I think I'd just not cast and instead do the specific ignore?

v-tupuri:
🙏

PR #259, opened 2025-03-13
feat: expose searching for checklists and data reviews

drake-nominal:
To perform the following task:

* Upon uploading data to a run / creating a run in Nominal, run a pre-existing checklist on the run

Users will have to do the following:
* search for checklist by name / properties / labels
* apply to run

To then query the failing checks, the first step they'll have to do is:
* search for datareview

Future work will need to be done to then go from datareview -> status of each check

### DIFF
--- a/nominal/core/data_review.py
+++ b/nominal/core/data_review.py
@@ -5,7 +5,13 @@ from datetime import timedelta
 from time import sleep
 from typing import Protocol, Sequence
 
-from nominal_api import scout, scout_checks_api, scout_datareview_api, scout_integrations_api
+from nominal_api import (
+    scout,
+    scout_checklistexecution_api,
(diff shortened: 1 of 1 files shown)


v-tupuri:
this looks good to me. potential flup to support more search types (other than and)

v-tupuri (approved):

nominal/core/client.py
```diff
@@ -638,6 +642,43 @@ def get_checklist(self, rid: str) -> Checklist:
         response = self._clients.checklist.get(self._clients.auth_header, rid)
         return Checklist._from_conjure(self._clients, response)
 
+    def search_checklists(
```
v-tupuri, line 645: just curious. when do we kwarg only args vs use them like this?

PR #266, opened 2025-03-18
feat: utilities to fetch scopes by name from assets

drake-nominal:
As I was documenting using assets from python, I noticed that pretty much every instance of using assets relied upon the users writing functions like this in their processing scripts-- adding them directly to the client to make this simpler!


v-tupuri:
if the answer to my other question is "yes", this pr looks good to me

v-tupuri (approved):

PR #267, opened 2025-03-18
feat: add functionality to add a video to an asset

drake-nominal:
Same vein as #266, allows users to add video to assets


v-tupuri:
this is more than just video right? (logset and attachments). approved but might wanna update description/title

v-tupuri (approved):

PR #269, opened 2025-03-18
feat: uniformly expose channel prefix tree delimiter in all create dataset methods

drake-nominal:
PR also adds misc. documentation along the way.

If I have channels named:

```
left_wing.roll
left_wing.pitch
left_wing.yaw
right_wing.roll
right_wing.yaw
right_wing.pitch
```

and I ingest with a channel prefix tree delimiter of `.`, then when I put the channel viewer in tree-mode, it'll represent them logically like:

```
left_wing
  |-- roll
  |-- pitch
  |-- yaw

right_wing
  |-- roll
  |-- pitch
  |-- yaw
```

### DIFF
--- a/nominal/core/client.py
+++ b/nominal/core/client.py
@@ -387,7 +387,7 @@ class NominalClient:
             where {unit} is one of: nanoseconds | microseconds | milliseconds | seconds | minutes | hours | days
 
         Args:
-            dataset: Binary file-like MCAP stream
+            dataset: Binary file-like tabular data stream
             name: Name of the dataset to create
             timestamp_column: Column of data containing timestamp information for all other columns
             timestamp_type: Type of timestamps contained within timestamp_column


nominal/core/client.py
```diff
@@ -381,6 +385,22 @@ def create_dataset_from_io(
             "epoch_{unit}": epoch timestamps in UTC (floats or ints),
             "relative_{unit}": relative timestamps (floats or ints),
             where {unit} is one of: nanoseconds | microseconds | milliseconds | seconds | minutes | hours | days
+
+        Args:
+            dataset: Binary file-like MCAP stream
```
v-tupuri, line 390: not MCAP stream right

nominal/core/client.py
```diff
@@ -381,6 +385,22 @@ def create_dataset_from_io(
...
+            name: Name of the dataset to create
+            timestamp_column: Column of data containing timestamp information for all other columns
+            timestamp_type: Type of timestamps contained within timestamp_column
+            file_type: Type of file being ingested (e.g. CSV, parquet, etc.). Used for naming the file uploaded
```
v-tupuri, line 394: should we error if they try passing in a FileType that's not tabular to this method?

v-tupuri: and same for the other _from_io methods?

v-tupuri (approved):

PR #274, opened 2025-03-21
feat: prevent duplicated file extensions when using upload_multipart_io

drake-nominal:
Fixes NOM-5514

### DIFF
--- a/nominal/core/client.py
+++ b/nominal/core/client.py
@@ -34,7 +34,7 @@ from nominal import _config
 from nominal._utils import deprecate_keyword_argument
 from nominal.core._clientsbunch import ClientsBunch
 from nominal.core._conjure_utils import _available_units, _build_unit_update
-from nominal.core._multipart import path_upload_name, upload_multipart_io
+from nominal.core._multipart import path_upload_name, upload_multipart_file, upload_multipart_io
 from nominal.core._utils import construct_user_agent_string, rid_from_instance_or_string
 from nominal.core.asset import Asset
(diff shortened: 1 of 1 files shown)


nominal/core/client.py
```diff
@@ -398,6 +395,7 @@ def create_dataset_from_io(
             properties: Key-value properties to apply to the cleated dataset
             prefix_tree_delimiter: If present, the delimiter to represent tiers when viewing channels hierarchically.
             channel_prefix: Prefix to apply to newly created channels
+            file_name: Name of the file (without extension) to create when uploading.
```
v-tupuri, line 398: why cant we derive this from the name arg

v-tupuri: i feel like having two sep parameters is kinda weird here

nominal/core/_multipart.py
```diff
@@ -206,7 +218,7 @@ def upload_multipart_file(
     if file_type is None:
         file_type = FileType.from_path(file)
```
v-tupuri, line 220: if we just did the normalization in the upload_multipart_io, this would save us from a lot of the changes you made in the client.py as well right?

nominal/core/client.py
alkasm: It's bad that the backend requires this in the first place. It shouldn't be necessary / shouldn't be tied to the file extension (extension is not disambiguating enough). Especially in the `from_io` classes where there may not be a file that it came from in the first place (such as in `upload_pandas()`)

PR #278, opened 2025-03-27
feat: add workspaces

alkasm:
This PR adds basic support for workspaces. Workspace IDs (their name) are specified when you construct the client. All API calls that depend on workspace as a parameter have the workspace passed in. Backwards compatibility is supported as it is in the APIs currently, by leaving the workspace param optional, defaulting to the org's default workspace.

Note that you can only use workspaces with `NominalClient.create(..., workspace)` currently---the global "default" client will have no workspace set.

To find all of the places where workspace _could_ be provided, I generated conjure types locally where I changed all instances of `optional<ids.WorkspaceId>` to `ids.WorkspaceId` so the type checker could find them all. After all instances were found, I then fell back to the newer public APIs.

Intentionally left out of this PR for now:
- ability to list/search across workspaces
- ability to specify workspaces with the global client

### DIFF
--- a/nominal/core/client.py
+++ b/nominal/core/client.py
@@ -77,7 +77,13 @@ class NominalClient:
 
     @classmethod
     def create(
-        cls, base_url: str, token: str | None, trust_store_path: str | None = None, connect_timeout: float = 30
+        cls,
+        base_url: str,
+        token: str | None,
+        trust_store_path: str | None = None,
+        connect_timeout: float = 30,
+        *,
+        workspace: str | None = None,
     ) -> Self:
         """Create a connection to the Nominal platform.
 
@@ -85,6 +91,8 @@ class NominalClient:
         token: An API token to authenticate with. By default, the token will be looked up in ~/.nominal.yml.
         trust_store_path: path to a trust store CA root file to initiate SSL connections. If not provided,
             certifi's trust store is used.
+        connect_timeout: The timeout in seconds for any API call to Nominal.
+        workspace: The workspace to use for the client. If not provided, the org's default workspace is used.
         """
         if token is None:
             token = _config.get_token(base_url)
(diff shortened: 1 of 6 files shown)


alkasm:
In draft while we decide if we want to specify workspace RIDs or IDs in the APIs.

nominal/core/client.py
```diff
@@ -1156,7 +1175,10 @@ def list_streaming_checklists(self, asset: Asset | str | None = None) -> Iterabl
             if asset is None:
                 response = self._clients.checklist_execution.list_streaming_checklist(
                     self._clients.auth_header,
-                    scout_checklistexecution_api.ListStreamingChecklistRequest(page_token=next_page_token),
+                    scout_checklistexecution_api.ListStreamingChecklistRequest(
+                        page_token=next_page_token,
+                        workspace_ids=[self._clients.workspace] if self._clients.workspace is not None else [],
```
niwong-nmnl, line 1180: this param might be renamed to `workspace` after @helenyugithub fixes it to be consistent: https://nominal-io.slack.com/archives/C082EPB6F19/p1743111142514139?thread_ts=1743028250.526259&cid=C082EPB6F19

nominal/core/client.py
niwong-nmnl, line 1: If there are any `searchX` endpoint usage in the python client, those will return results across all workspaces the user has access to unless we `SearchQuery.and([...]` in a `SearchQuery.workspace(...)` option. Those aren't able to be caught by the temporary conjure required strategy, since the workspace is just one of the union types. Unsure whether limiting the Python client to a single workspace's worth of search results (the "current" workspace's) is a feature or a bug though :thinking:

For reference, any endpoints in row 3 of [this Google Sheet](https://docs.google.com/spreadsheets/d/1QeV8zNyd6i84gicOJC0ksTk-yx_TEV4CT7-IEntvw0I/edit?usp=sharing) have been given new request slots for `workspace`

alkasm:
The APIs changed to require RIDs instead of IDs. Created a new PR for this here: https://github.com/nominal-io/nominal-client/pull/299

PR #284, opened 2025-03-28
feat: allow users to optionally be warned when setting display-only units

drake-nominal:
Generally improves ergonomics when setting units at the dataset level.

image

Call it a cardinal sin, but I think it made more sense to delete the broken unit tests than fix them here. They didn't particularly add value-- just mocking everything and testing that the code is the same as when the tests were written.

Closes: NOM-4310
Closes: NOM-4631

### DIFF
--- a/nominal/core/unit.py
+++ b/nominal/core/unit.py
@@ -68,7 +68,7 @@ def _build_unit_update(unit: UnitLike) -> timeseries_logicalseries_api.UnitUpdat
         return timeseries_logicalseries_api.UnitUpdate(unit=unit)
 
 
-def _warn_on_invalid_units(unit_map: UnitMapping, unit_service: scout.UnitsService, auth_header: str) -> None:
+def _error_on_invalid_units(unit_map: UnitMapping, unit_service: scout.UnitsService, auth_header: str) -> None:
     # Normalize unit map to refer to channel names and unit symbols
     channels_to_units = {channel: _unit_symbol_from_unit_like(unit) for channel, unit in unit_map.items()}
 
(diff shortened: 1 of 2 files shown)


nominal/core/unit.py
```diff
@@ -28,3 +39,63 @@ def _from_conjure(cls, api_unit: api.Unit | scout_units_api.Unit | scout_run_api
...
+        if unit_symbol is None:
+            continue
+
+        try:
+            resolved_unit = unit_service.get_unit(auth_header, unit_symbol)
+            if resolved_unit is None:
+                invalid_units.add(unit_symbol)
+        except conjure_python_client.ConjureHTTPError as ex:
+            logger.debug("Error fetching unit '%s': '%s'", unit_symbol, ex)
+            invalid_units.add(unit_symbol)
```
alkasm, line 89: We can merge this as-is, or we can wait for the batched `/get-units` PR to merge and just support that. It would be an easy PR to update as an impl detail later, but I imagine this is likely to 429 so might be better to just wait a day?

nominal/core/channel.py
```diff
@@ -69,15 +70,17 @@ def update(
...
         """Replace channel metadata within Nominal, and updates / returns the local instance.
 
         Only the metadata passed in will be replaced, the rest will remain untouched.
 
         Args:
             description: Human-readable description of data within the channel
-            unit: Unit symbol to apply to the channel
+            unit: Unit symbol to apply to the channel.
+                NOTE: in contrast to 'set_channel_units' on a `nominal.Dataset`, providing `None` as the unit
+                      here does *not* clear the unit, but rather, leaves the unit for the channel unaffected.
```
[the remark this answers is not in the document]
alkasm: Ah sorry didn't realize the clash of terms here, for the `UNSET` pattern I mean a default sentinel arg, something like this (changed `UNSET` to `NOT_PROVIDED` for clarity):

```py
class NotProvided: pass
NOT_PROVIDED = NotProvided()
...

def f(unit: str | None | NotProvided = NOT_PROVIDED) -> None:
    """
    If unit has type str, update the unit.
    If unit is None, clear the unit.
    If unit is not provided (default), don't update the unit.
    """
```

Now you can detect if the caller did not pass an argument vs if the caller explicitly passed in `None`

nominal/core/unit.py
```diff
@@ -28,3 +38,57 @@ def _from_conjure(cls, api_unit: api.Unit | scout_units_api.Unit | scout_run_api
...
+    channels_to_units = {channel: _unit_symbol_from_unit_like(unit) for channel, unit in unit_map.items()}
+
+    resolved_units = unit_service.get_batch_units(
+        auth_header, [unit_symbol for unit_symbol in set(channels_to_units.values()) if unit_symbol is not None]
+    )
+
+    # Get set of all provided invalid units
+    invalid_units = set(
+        [unit_symbol for unit_symbol in channels_to_units.values() if unit_symbol not in resolved_units]
+    )
```
alkasm, line 82: if you store the `set(channels_to_units.values())` above you can just do a set difference here

alkasm (commented):
Now that we are explicitly checking for validity, should we actually throw rather than warn?

alkasm (approved):

PR #285, opened 2025-03-31
fix: poll until ingestion completed exits if status is error

v-tupuri:
we don't need to poll until ingestion completed now that we have concurrent ingests


uv.lock
```diff
@@ -1644,7 +1643,6 @@ requires-dist = [
     { name = "tabulate", specifier = ">=0.9.0,<0.10" },
     { name = "typing-extensions", specifier = ">=4,<5" },
 ]
-provides-extras = ["hdf5", "protos"]
```
alkasm, line 1647: why was this removed

alkasm (approved):

PR #290, opened 2025-04-01
feat: add nm.upload_tabular function

najork:

### DIFF
--- a/nominal/nominal.py
+++ b/nominal/nominal.py
@@ -212,6 +212,61 @@ def _upload_csv(
     return dataset
 
 
+def upload_tabular(
+    file: Path | str,
+    name: str | None,
+    timestamp_column: str,
+    timestamp_type: ts._AnyTimestampType,
+    description: str | None = None,
+    channel_name_delimiter: str | None = None,
+    *,
+    wait_until_complete: bool = True,
+) -> Dataset:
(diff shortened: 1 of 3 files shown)


nominal/nominal.py
```diff
@@ -212,6 +212,61 @@ def _upload_csv(
...
+    """Create a dataset in the Nominal platform from a table-like file (CSV, parquet, etc.)
+
+    If `name` is None, the dataset is created with the name of the file.
+
+    If `wait_until_complete=True` (the default), this function waits until the dataset has completed ingestion before
+        returning. If you are uploading many datasets, set `wait_until_complete=False` instead and call
+        `wait_until_ingestions_complete()` after uploading all datasets to allow for parallel ingestion.
+    """
+    conn = get_default_client()
+    return _upload_tabular(
```
alkasm, line 234: can we just inline it here? I think the only other delegations in this file are because it's used multiple times

alkasm: (also I'm about to start deprecating these top level functions in general, so might be best to just not write new functions w/ new scripts that depend on them?)

alkasm (approved):

nominal/nominal.py
najork: If you're planning on deprecating these in the near term, let's close this. That's a nicer long-term path anyway, I just wanted to have consistency here so people don't see upload_csv and think that's all we support.

PR #297, opened 2025-04-04
feat: update nominal-api to 0.618.0

drake-nominal:


nominal/core/_batch_processor.py
```diff
@@ -34,6 +34,31 @@ def make_points(api_batch: Sequence[BatchItem]) -> storage_writer_api.Points:
     raise ValueError("only float and string are supported types for value")
 
 
+def make_points_external(api_batch: Sequence[BatchItem]) -> storage_writer_api.PointsExternal:
```
alkasm, line 37: wait why didn't you just change the existing impl above?

PR #299, opened 2025-04-08
feat: add workspace rids

alkasm:
Subsumes #278 

This PR adds basic support for workspaces. Workspace RIDs (not their name/ID) are specified when you construct the client. All API calls that depend on workspace as a parameter have the workspace passed in. Backwards compatibility is supported as it is in the APIs currently, by leaving the workspace param optional, defaulting to the org's default workspace.

Note that you can only use workspaces with `NominalClient.create(..., workspace_rid)` currently---the global "default" client will have no workspace set.

To find all of the places where workspace _could_ be provided, I generated conjure types locally where I changed all instances of `optional<rids.WorkspaceRid>` to `rids.WorkspaceRid` so the type checker could find them all. After all instances were found, I then fell back to the newer public APIs.

Intentionally left out of this PR for now:
- customizing list/search with workspaces
- ability to specify workspaces with the global client

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -25,7 +25,6 @@ requires-python = ">=3.9,<4"
 dependencies = [
     "click>=8,<9",
     "conjure-python-client>=2.8.0,<3",
-    "nominal-api==0.618.0",
     "nptdms>=1.9.0,<2",
     "tabulate>=0.9.0,<0.10",
     "typing-extensions>=4,<5",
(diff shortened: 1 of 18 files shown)


pyproject.toml
```diff
@@ -36,6 +35,7 @@ dependencies = [
     "pyyaml>=0.0.0",
     "requests>=0.0.0",
     "ffmpeg-python>=0.2.0",
+    "nominal-api>=0.621.0",
```
niwong-nmnl, line 38: just checked, anything `>= 0.616.0` should capture all workspace-related API changes 👍

niwong-nmnl (approved):
Thanks for doing this! Pre-approving

Just confirming that this change intentionally doesn't use workspace for search/list endpoints? E.g., if someone was in multiple workspaces in their organization, search results would span all workspaces

PR #302, opened 2025-04-10
feat: expose tags to enqueue_from_dict

nickpeng95:
Expose tags to the `enqueue_from_dict` streaming function

### DIFF
--- a/nominal/core/write_stream_base.py
+++ b/nominal/core/write_stream_base.py
@@ -82,7 +82,7 @@ class WriteStreamBase(abc.ABC):
             timestamp: The shared timestamp to use for all items to enqueue.
             channel_values: A dictionary mapping channel names to their respective values.
             tags: Key-value tags associated with the data being uploaded.
-                NOTE: This *must* include all `required_tags` used when creating a `Connection` to Nominal.
+                NOTE: This *should* include all `required_tags` used when creating a `Connection` to Nominal.
         """
         for channel, value in channel_values.items():
             self.enqueue(channel, timestamp, value, tags)
--- a/nominal/core/write_stream_base.py
+++ b/nominal/core/write_stream_base.py
@@ -71,6 +71,7 @@ class WriteStreamBase(abc.ABC):
         self,
         timestamp: str | datetime | IntegralNanosecondsUTC,
         channel_values: dict[str, float | str],
+        tags: dict[str, str] | None = None,
     ) -> None:
         """Write multiple channel values at a given timestamp using a flattened dictionary.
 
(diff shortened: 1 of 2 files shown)


nominal/experimental/stream_v2/_write_stream.py
```diff
@@ -121,6 +121,7 @@ def enqueue_from_dict(
         self,
         timestamp: str | datetime | IntegralNanosecondsUTC,
         channel_values: dict[str, float | str],
+        tags: dict[str, str] | None = None,
```
alkasm, line 124: We should use `Mapping` rather than `dict` here (same in the other params)

alkasm, line 124: We should use `Mapping` rather than `dict` here (same in the other params)

nominal/experimental/stream_v2/_write_stream.py
```diff
@@ -130,12 +131,14 @@ def enqueue_from_dict(
         Args:
             timestamp: The common timestamp to use for all enqueued items.
             channel_values: A dictionary mapping channel names to their values.
+            tags: Key-value tags associated with the data being uploaded.
+                NOTE: This *must* include all `required_tags` used when creating a `Connection` to Nominal.
```
alkasm, line 135: Aren't required tags a read-side concept, not a write-side concept?

nickpeng95: I copied the note from the enqueue function but is this mostly reminding users that they need to write with the required tags so it's not accidentally filtered by the connection if the required tags aren't written?

alkasm: Let's rephrase as SHOULD rather than MUST

nominal/core/write_stream_base.py
```diff
@@ -80,9 +81,11 @@ def enqueue_from_dict(
         Args:
             timestamp: The shared timestamp to use for all items to enqueue.
             channel_values: A dictionary mapping channel names to their respective values.
+            tags: Key-value tags associated with the data being uploaded.
+                NOTE: This *must* include all `required_tags` used when creating a `Connection` to Nominal.
```
alkasm, line 85:
```suggestion
                NOTE: This *should* include all `required_tags` used when creating a `Connection` to Nominal.
```

nominal/experimental/stream_v2/_write_stream.py
alkasm, line 135:
```suggestion
                NOTE: This *should* include all `required_tags` used when creating a `Connection` to Nominal.
```

alkasm: updated the PR

nominal/core/write_stream_base.py
```diff
@@ -71,6 +71,7 @@ def enqueue_from_dict(
         self,
         timestamp: str | datetime | IntegralNanosecondsUTC,
         channel_values: dict[str, float | str],
```
alkasm, line 73: I think this also needs to be a `Mapping`
```suggestion
        channel_values: Mapping[str, float | str],
```

PR #308, opened 2025-04-16
feat: expose workspaces to client

drake-nominal:

### DIFF
--- a/nominal/core/client.py
+++ b/nominal/core/client.py
@@ -121,7 +121,7 @@ class NominalClient:
 
         Raises:
             RuntimeError: Raises a RuntimeError if a workspace is not provided, but there is no configured default
-                workspace for the current workspace.
+                workspace for the current user.
         """
         if workspace_rid is None:
             raw_workspace = self._clients.workspace.get_default_workspace(self._clients.auth_header)


nominal/core/client.py
```diff
@@ -109,6 +110,28 @@ def get_user(self) -> User:
...
+            raw_workspace = self._clients.workspace.get_default_workspace(self._clients.auth_header)
+            if raw_workspace is None:
+                raise RuntimeError(
+                    "Could not retrieve default workspace! "
+                    "Either the user is not authorized to access or there is no default workspace."
+                )
+
+            return Workspace._from_conjure(raw_workspace)
+        else:
+            raw_workspace = self._clients.workspace.get_workspace(self._clients.auth_header, workspace_rid)
```
niwong-nmnl, line 125: are we okay if this throws an unhandled 4xx?

alkasm: yes

nominal/core/client.py
```diff
@@ -109,6 +110,28 @@ def get_user(self) -> User:
         """Retrieve the user associated with this client."""
         return _get_user(self._clients.auth_header, self._clients.authentication)
 
+    def get_workspace(self, workspace_rid: str | None = None) -> Workspace:
+        """Get workspace via given RID, or the default workspace if no RID is provided."""
```
alkasm, line 114: can you add the exception for default workspaces in the docstring?

nominal/core/client.py
alkasm: (we are not too consistent about this currently, so it happens everywhere anyways)

PR #311, opened 2025-04-17
feat: allow creating empty videos

drake-nominal:


ritwikdixit (approved):
maybe just smoke test on staging

PR #316, opened 2025-04-22
feat: allow listing all ingested files on a dataset

alkasm:


nominal/core/dataset.py
```diff
@@ -411,25 +411,25 @@ def get_channels(
...
-                    )
-
+            yield from files_page.files
             if files_page.next_page is None:
                 break
-            else:
-                next_page_token = files_page.next_page
+            next_page_token = files_page.next_page
+
+    def list_files(self, *, successful_only: bool = True) -> Iterable[DatasetFile]:
```
v-tupuri, line 423: while we are at it, wdyt about making the param ingest_status: enum instead of succesful_only: bool

so they can query for files of a specific status

alkasm: we would be doing that as a post-filter anyways, so might as well let them do their own post-filter

v-tupuri (approved):

PR #319, opened 2025-04-27
ci: add pre-commit config

stefanv:
Developers can do `pip install pre-commit; pre-commit install` once, whereafter ruff & mypy checks are enforced on each commits.

This prevents the frustration of having CI reject a commit because of linting errors, by having code checked automatically locally.


stefanv:
Yes, definitely possible. This creates a new venv for running the tests in, which has some benefits, but I think OK to run rust-just in the local env too. Not sure we can tell it to install a standard set of type dependencies then, though, so I have slight preference for the current version.

PR #321, opened 2025-04-27
chore: allow untyped defs in test suite

stefanv:
This avoids MyPy complaining about `def test_foo(): ...` not being annotated with `-> None`.


alkasm (approved):
alternatively we could just make the tests all `-> None` but I'm not really sure how strongly I feel about whether tests need to be strictly type checked anyways

PR #322, opened 2025-04-27
fix: fix incompatible type assignments

stefanv:
Please check whether https://github.com/nominal-io/nominal-client/pull/322/files#diff-631a5e21882596a3d77dc013e11f0cb1f5001c730c6c33d3c8152d2ebdbfe35bR23 is desirable

### DIFF
--- a/nominal/core/stream.py
+++ b/nominal/core/stream.py
@@ -20,7 +20,7 @@ logger = logging.getLogger(__name__)
 @dataclass(frozen=True)
 class BatchItem:
     channel_name: str
-    timestamp: datetime | IntegralNanosecondsUTC
+    timestamp: IntegralNanosecondsUTC
     value: float | str
     tags: Mapping[str, str] | None = None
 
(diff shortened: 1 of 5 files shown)


stefanv:
I wasn't sure which way to take it. This required fewest changes (the code already internally converts in most cases), but happy to go the other way.

nominal/core/stream.py
```diff
@@ -20,7 +20,7 @@
 @dataclass(frozen=True)
 class BatchItem:
     channel_name: str
-    timestamp: IntegralNanosecondsUTC
+    timestamp: datetime | IntegralNanosecondsUTC
```
alkasm, line 23: I don't think we should turn this into a union, should just be integral ns and we should fix anywhere that doesn't use that

alkasm (changes requested):

nominal/core/stream.py
stefanv: OK, will fix up!

stefanv:
> Would it not be better to just change the tests to pass in int nanos instead

This is now done.

alkasm (approved):

stefanv:
@drake-nominal Did you want to take another look before merging?

PR #325, opened 2025-04-30
chore: indicate that client is compatible with Python 3.13

stefanv:
I've been running it under 3.13 for a while now without any issues.

3.9 is EOL October 2025, FWIW, around the same time 3.14 is being released.


alkasm (approved):

alkasm:
> I've been running it under 3.13 for a while now without any issues.

same!

PR #326, opened 2025-04-30
chore: mypy 1.15 will flag `ParamSpec.args` / `ParamSpec.kwargs` being split

stefanv:
Work around it by manually extracting those keywords, and ignoring concerns about assignment (we know click will do its job).

This use-case is not supported by PEP612, although it may be handled in the future.


alkasm:
> This use-case is not supported by PEP612

Just for posterity, here's the section directly talking about it: https://peps.python.org/pep-0612/#concatenating-keyword-parameters

alkasm (approved):

PR #331, opened 2025-05-13
feat: gzip post contents by default

drake-nominal:
When we use conjure's default `RequestClient`, the data that we send in requests is sent as a raw byte string of the json contents of the request. This is typically sufficient, however, is drastically slower when it comes to things like:

- streaming points to nominal (without using encoded protobufs)
- uploading files to nominal (that aren't themselves gzipped) using the upload service (NOTE: not done in practice by the client anymore... but worked for proof of concept)

When network constrained, such as on crappy wifi, laptop, etc., we can see up to 4-6x speedups in practice by default-gzipping these request bodies!

Here's the before and after on streaming 1,000,000 points using default `dataset.get_write_stream()` and `write_stream.enqueue()` in a tight loop on my machine / crappy wifi:

Before:

```
36.97300410270691 seconds for 1000000 points (27046.760853462456 points/second)
```

After:

```
8.587828159332275 seconds for 1000000 points (116443.87631502777 points/second)
```

### DIFF
--- a/nominal/core/_networking.py
+++ b/nominal/core/_networking.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import gzip
+import os
 from typing import Any, Callable, Mapping, Type, TypeVar
 
 import requests
@@ -36,8 +37,8 @@ class GzipRequestsAdapter(TransportAdapter):
         if isinstance(body, (bytes, str)):
             content_length = len(body)
         else:
-            content_length = body.seek(0, 2)
-            body.seek(0, 0)
+            content_length = body.seek(0, os.SEEK_END)
+            body.seek(0, os.SEEK_SET)
 
         headers = {
             self.ACCEPT_ENCODING: "gzip",


nominal/network/nominal_session.py
```diff
@@ -0,0 +1,92 @@
...
+
+import requests
+from conjure_python_client import RequestsClient, ServiceConfiguration
+from conjure_python_client._http.requests_client import RetryWithJitter, TransportAdapter
+
+logger = logging.getLogger(__name__)
+T = TypeVar("T")
+
+
+class NominalRequestsClient(RequestsClient):
```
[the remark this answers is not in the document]
alkasm: can you also permalink to the conjure impl in a comment?

nominal/network/nominal_session.py
```diff
@@ -0,0 +1,92 @@
...
+            session,
+            service_config.uris,
+            service_config.connect_timeout,
+            service_config.read_timeout,
+            verify,
+            return_none_for_unknown_union_types,
+        )
+
+
+class NominalRequestsSession(requests.Session):
```
alkasm, line 64: I think we should be customizing a transport adapter, not the session. Also would name it more semantically.

nominal/network/nominal_session.py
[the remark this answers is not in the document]
alkasm: I also don't think we get any benefit from subclassing from RequestsClient here since it's just a static method and the class doesn't have any other functionality

nominal/network/nominal_session.py
alkasm: tldr I want to avoid the redundancy

nominal/network/__init__.py
```diff
@@ -0,0 +1,3 @@
+from nominal.network.nominal_session import GzipRequestsClient, GzipRequestsSession
```
alkasm, line 1: We should probably throw this in a private module inside core? Or is there a use-case you were thinking of for keeping it outside of there?

nominal/network/gzip_requests_adapter.py
```diff
@@ -0,0 +1,79 @@
...
+        if stream:
+            # Typechecking
+            if isinstance(request.body, (bytes, str)):
+                raise ValueError("Expected request body to not be bytes or string if stream")
+            elif request.body is None:
+                raise ValueError("Expected request body to be non-null whin streaming")
+
+            # Having a file-like object, therefore we need to stream the
+            # content into a new one through the compressor.
+            compressed_body = io.BytesIO()
```
alkasm, line 63: typically a stream object _is_ a stream specifically to avoid not reading everything in memory, so probably would nix this. Most of the conjure endpoints don't use a stream anyways (only the binary types, and I think for all of those we have a multipart upload variant that the client uses) so I think we could safely just `if stream: super().send(...)`

nominal/network/gzip_requests_adapter.py
```diff
@@ -0,0 +1,79 @@
...
+        if body is None:
+            return
+
+        if isinstance(body, (bytes, str)):
+            content_length = len(body)
+        else:
+            content_length = body.seek(0, 2)
+            body.seek(0, 0)
+
+        headers: MutableMapping[str, str] = {
```
alkasm, line 37: this annotation shouldn't be needed

nominal/network/gzip_requests_client.py
```diff
@@ -0,0 +1,60 @@
...
+    """Wrapper around conjures RequestClient to automatically gzip post-requested data.
+
+    In bandwidth constrained scenarios, this has been measured to have 5-6x speedups in
+    uploading data, depending on its compressability.
+
+    See: https://github.com/palantir/conjure-python-client/blob/60d6d7639502a3b0fe18fad388ce84cbc54eb613/conjure_python_client/_http/requests_client.py#L181
+    """
+
+    @classmethod
+    def create(
```
alkasm, line 27: i guess we could just turn this into a function

nominal/core/_networking.py
```diff
@@ -0,0 +1,144 @@
...
+
+    COMPRESSION_LEVEL = 1
+
+    def add_headers(self, request: requests.PreparedRequest, **kwargs: Any) -> None:
+        """Tell the server that we support compression."""
+        super().add_headers(request, **kwargs)  # type: ignore[no-untyped-call]
+
+        body = request.body
+        if body is None:
+            return
```
alkasm, line 32: or if `request.stream` is true as well now, I guess

alkasm (approved):

nominal/core/_networking.py
```diff
@@ -0,0 +1,146 @@
+from __future__ import annotations
+
+import gzip
+from typing import Any, Callable, Mapping, Type, TypeVar
```
stefanv, line 4:
```suggestion
from typing import Any, Callable, Mapping, Type, TypeVar
from os import SEEK_END
```

nominal/core/_networking.py
```diff
@@ -0,0 +1,146 @@
...
+        body = request.body
+        if body is None:
+            return
+        elif kwargs.get("stream", False):
+            return
+
+        if isinstance(body, (bytes, str)):
+            content_length = len(body)
+        else:
+            content_length = body.seek(0, 2)
```
stefanv, line 39:
```suggestion
            content_length = body.seek(0, SEEK_END)
```

stefanv (approved):

PR #334, opened 2025-05-16
docs: organize reference docs around namespace packages

stefanv:
This is a minor re-organization of the API reference docs to make the namespace structure clearer, since that is now recommended usage (`nominal.core.NominalClient` etc.).

I could not get `mkdocs` to recognize `nominal.thirdparty.pandas` and `nominal.thirdparty.tdms`, so commented those out for now.


alkasm (approved):

stefanv:
`griffe`, which `mkdocstrings` calls under the hood, seems to require a `__init__.py` in the namespace module dir for discovery. I don't imagine it will cause problems, but highlighting it here just in case.

PR #336, opened 2025-05-16
fix: remove warning on uploading CSV

stefanv:
Presumably, this warning was for the case where filetype was *neither* parquet or csv for a tabular upload. Right now, it shows up when calling `add_tabular_data` with a CSV file.

### DIFF
--- a/nominal/core/dataset.py
+++ b/nominal/core/dataset.py
@@ -609,7 +609,7 @@ def _construct_new_ingest_options(
             )
         )
     else:
-        if file_type.is_csv():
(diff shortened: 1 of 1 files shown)


nominal/core/dataset.py
```diff
@@ -609,7 +609,7 @@ def _construct_new_ingest_options(
             )
         )
     else:
-        if file_type.is_csv():
+        if not file_type.is_csv():
```
alkasm, line 612: good catch! 👀

alkasm (approved):

PR #343, opened 2025-05-19
feat: add nix flake

alkasm:
This allows the `nominal-client` repo itself to be used as a nix flake input. Exposes output packages `nominal-client`, `nominal-api`, and `nominal-api-protos`, incidentally along with `conjure-python-client`.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -35,7 +35,7 @@ dependencies = [
     "requests>=0.0.0",
     "ffmpeg-python>=0.2.0",
     "nominal-api==0.703.0",
-    "coverage[toml]>=7.8.0",
+    "coverage[toml]>=7.0.0",
 ]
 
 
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -26,12 +26,10 @@ requires-python = ">=3.9,<4"
 dependencies = [
     "click>=8,<9",
     "conjure-python-client>=2.8.0,<3",
-    "nptdms>=1.9.0,<2",
     "tabulate>=0.9.0,<0.10",
     "typing-extensions>=4,<5",
     # Any version is okay for these dependencies
     "pandas>=0.0.0",
-    "polars>=0.0.0",
```
stefanv, line 34: Until `upload_polars` is deprecated, we may need to keep polars around, at least in the development extras, for type checking to succeed.

nionata: @alkasm that would be great! Let me know when that is available.

PR #350, opened 2025-05-28
feat: add get or create dataset by refname on assets

bryli:
For a slightly smoother file -> dataset -> asset -> run workflow, want to add the following functions:

* `client.get_or_create_asset`
* `asset.get_or_create_dataset`
* `dataset.add_files_to_dataset`

Of the three, the get/create dataset by refname is the simplest and at least risk of future changes.

Get/create asset is pending better text search, and add multiple files to dataset is pending tag additions.

### DIFF
--- a/nominal/core/dataset.py
+++ b/nominal/core/dataset.py
@@ -532,8 +532,8 @@ def _create_dataset(
     request = scout_catalog.CreateDataset(
         name=name,
         description=description,
-        labels=[*labels],
-        properties={} if properties is None else {**properties},
+        labels=list(labels),
+        properties={} if properties is None else dict(properties),
         is_v2_dataset=True,
         metadata={},
         origin_metadata=scout_catalog.DatasetOriginMetadata(),


nominal/core/asset.py
```diff
@@ -229,6 +230,36 @@ def list_data_scopes(self) -> Sequence[tuple[str, ScopeType]]:
         """
         return (*self.list_datasets(), *self.list_connections(), *self._list_logsets(), *self.list_videos())
 
+    def get_or_create_dataset(
+        self,
+        *,
+        data_scope_name: str,
```
alkasm, line 236:
```suggestion
        data_scope_name: str,
        *,
```
this seems fine as a positional arg

nominal/core/asset.py
```diff
@@ -229,6 +230,36 @@ def list_data_scopes(self) -> Sequence[tuple[str, ScopeType]]:
...
+        name: str | None = None,
+        description: str | None = None,
+        labels: Sequence[str] = (),
+        properties: Mapping[str, str] | None = None,
+    ) -> Dataset:
+        """Retrieve a dataset by data scope name, or create a new one if it does not exist."""
+        try:
+            return self.get_dataset(data_scope_name)
+        except ValueError:
```
alkasm, line 245: we should probably make get_dataset return a more nominal-specific error rather than a value error...but that doesn't have to be in this PR

nominal/core/asset.py
```diff
@@ -229,6 +230,36 @@ def list_data_scopes(self) -> Sequence[tuple[str, ScopeType]]:
...
+                    description=description,
+                    properties={} if properties is None else dict(properties),
+                    labels=list(labels),
+                    is_v2_dataset=True,
+                    metadata={},
+                    origin_metadata=scout_catalog.DatasetOriginMetadata(),
+                    workspace=self._clients.workspace_rid,
+                ),
+            )
+            dataset = Dataset._from_conjure(self._clients, enriched_dataset)
```
alkasm, line 259: hmmm I'd like to just reuse `client.create_dataset` rather than duplicate the impl, but we don't have that method available here. kinda surprised this is the first time we've had that issue...

let's make a private utility func at least that they can both depend on

nominal/core/dataset.py
```diff
@@ -519,6 +519,29 @@ def _get_dataset(
...
+    description: str | None = None,
+    labels: Sequence[str] = (),
+    properties: Mapping[str, str] | None = None,
+    workspace_rid: str | None = None,
+) -> scout_catalog.EnrichedDataset:
+    request = scout_catalog.CreateDataset(
+        name=name,
+        description=description,
+        labels=[*labels],
+        properties={} if properties is None else {**properties},
```
alkasm, line 536:
```suggestion
        labels=list(labels),
        properties={} if properties is None else dict(properties),
```
functionally equivalent but this is how we do it everywhere else in the client

alkasm (approved):

PR #381, opened 2025-06-07
feat: unify run & asset implementation with base class, uniformly allow adding with tags & offset

drake-nominal:

### DIFF
--- a/nominal/core/run.py
+++ b/nominal/core/run.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 from dataclasses import dataclass, field
 from datetime import datetime, timedelta
 from types import MappingProxyType
-from typing import Iterable, Mapping, Protocol, Sequence, cast
+from typing import ClassVar, Iterable, Mapping, Protocol, Sequence, cast
 
 from nominal_api import (
     scout,
@@ -26,6 +26,15 @@ from nominal.ts import IntegralNanosecondsUTC, _SecondsNanos
 
 @dataclass(frozen=True)
 class Run(HasRid, _DataSourceContainer):
+    class _Clients(
+        Asset._Clients,
+        _DataSourceContainer._Clients,
+        HasScoutParams,
+        Protocol,
+    ):
+        @property
+        def run(self) -> scout.RunService: ...
+
     rid: str
     name: str
     description: str
@@ -37,15 +46,6 @@ class Run(HasRid, _DataSourceContainer):
     assets: Sequence[str]
 
     _clients: _Clients = field(repr=False)
-
-    class _Clients(
-        Asset._Clients,
-        _DataSourceContainer._Clients,
-        HasScoutParams,
-        Protocol,
-    ):
-        @property
-        def run(self) -> scout.RunService: ...
 
     @property
     def nominal_url(self) -> str:


nominal/core/datasource_container.py
```diff
@@ -0,0 +1,328 @@
...
+
+
+class DatasourceContainer(abc.ABC):
+    def __init__(self, clients: _DatasourceContainerClients):
+        """Initialize datasource container
+
+        Args:
+            clients: Clientsbunch to use for requests
+        """
+        self.__clients = clients
```
alkasm, line 43: why double underscore? is the class name mangling important for some reason?

nominal/core/datasource_container.py
```diff
@@ -0,0 +1,328 @@
...
+    DataSource._Clients,
+    LogSet._Clients,
+    Video._Clients,
+    HasScoutParams,
+    Protocol,
+):
+    pass
+
+
+class DatasourceContainer(abc.ABC):
```
alkasm, line 36: I think data source is two words in the APIs?

alkasm: this also shouldn't be public as this isn't a part of the data model

nominal/core/datasource_container.py
```diff
@@ -0,0 +1,328 @@
...
+from nominal.core._clientsbunch import HasScoutParams
+from nominal.core._utils import HasRid, rid_from_instance_or_string
+from nominal.core.attachment import Attachment
+from nominal.core.connection import Connection, _iter_connections
+from nominal.core.dataset import Dataset, _get_datasets
+from nominal.core.datasource import DataSource
+from nominal.core.log import LogSet, _get_log_set
+from nominal.core.video import Video, _get_video
+
+ScopeType: TypeAlias = "Connection | Dataset | LogSet | Video"
```
alkasm, line 19: you shouldn't need the string here

nominal/core/datasource_container.py
alkasm: You can mix dataclasses and abcs; i'd vote that we should for uniformity

nominal/core/asset.py
```diff
@@ -315,14 +199,19 @@ def unarchive(self) -> None:
...
     ) -> None:
-        data_scope_names = data_scope_names or []
-        data_sources = data_sources or []
+        """Remove data scopes from this asset.
+
+        `names` are scope names.
+        `scopes` are rids or scope objects.
+        """
+        data_scope_names = set(names or [])
+        data_sources = set([rid_from_instance_or_string(scope) for scope in scopes] if scopes else [])
```
alkasm, line 214: I would prefer that we keep uniformity like we do elsewhere in the client of explicitly checking for None in a ternary and not utilizing as much short circuiting

nominal/core/datasource_container.py
```diff
@@ -0,0 +1,328 @@
...
+ScopeTypeSpecifier: TypeAlias = Literal["dataset", "video", "connection", "logset"]
+
+logger = logging.getLogger(__name__)
+
+
+class _DatasourceContainerClients(
```
alkasm, line 25: This can still be `_Clients` in the class which is then used as `DataSourceContainer._Clients` in the subclasses, no?

nominal/core/asset.py
```diff
@@ -35,16 +36,23 @@ class Asset(HasRid):
...
+        _DatasourceContainerClients,
         HasScoutParams,
         Protocol,
     ):
         @property
         def assets(self) -> scout_assets.AssetService: ...
 
+    def __post_init__(self) -> None:
+        """Initialize clientsbunch in DatasourceContainer abstract class."""
+        super().__init__(clients=self._clients)
```
alkasm, line 48: I think you might be able to get rid of the post init if you just swapped the order of the attribute definitions (i.e. if you put `_clients` first)

nominal/core/data_source_container.py
```diff
@@ -0,0 +1,484 @@
...
+
+    # Require that confirming classes have some clientsbunch
+    # that inherits from this one
+    @property
+    @abc.abstractmethod
+    def _clients(self) -> _Clients: ...
+
+    @abc.abstractmethod
+    def _rids_by_source_name(self, stype: SourceTypeSpecifier) -> Mapping[str, str]:
+        """Retrieve a mapping of scop names to rids.
```
alkasm, line 46:
```suggestion
        """Retrieve a mapping of scope names to rids.
```

PR #383, opened 2025-06-10
feat: allow adding datasources to runs and assets with tags and offsets

drake-nominal:

### DIFF
--- a/nominal/core/asset.py
+++ b/nominal/core/asset.py
@@ -98,6 +98,11 @@ class Asset(HasRid):
         Assets map "data_scope_name" (their name within the asset) to a Dataset (or dataset rid). The same type of
         datasets should use the same data scope name across assets, since checklists and templates use data scope names
         to reference datasets.
+
+        Args:
+            data_scope_name: logical name for the data scope within the asset
+            dataset: dataset to add to the asset
+            series_tags: Key-value tags to pre-filter the dataset with before adding to the asset.
         """
         request = scout_asset_api.AddDataScopesToAssetRequest(
             data_scopes=[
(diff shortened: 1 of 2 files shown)


nominal/core/asset.py
```diff
@@ -215,6 +228,10 @@ def list_logsets(self) -> Sequence[tuple[str, LogSet]]:
         """
         return self._list_logsets()
 
+    @deprecated(
+        "LogSets are deprecated and will be removed in a future version. "
+        "Add logs to an existing dataset with dataset.write_logs instead."
+    )
```
alkasm, line 234:
```suggestion
```

this is an internal method that's called by the method that's already got a deprecation warning

alkasm (approved):

nominal/core/asset.py
```diff
@@ -86,20 +88,27 @@ def nominal_url(self) -> str:
...
+        dataset: Dataset | str,
+        *,
+        series_tags: Mapping[str, str] | None = None,
+        offset: timedelta | IntegralNanosecondsDuration | None = None,
+    ) -> None:
         """Add a dataset to this asset.
 
         Assets map "data_scope_name" (their name within the asset) to a Dataset (or dataset rid). The same type of
         datasets should use the same data scope name across assets, since checklists and templates use data scope names
         to reference datasets.
```
stefanv, line 103: Missing documentation for `series_tags` and `offset`; also below.

nominal/core/asset.py
```diff
@@ -86,20 +88,27 @@ def nominal_url(self) -> str:
...
+        self,
+        data_scope_name: str,
+        dataset: Dataset | str,
+        *,
+        series_tags: Mapping[str, str] | None = None,
+        offset: timedelta | IntegralNanosecondsDuration | None = None,
+    ) -> None:
         """Add a dataset to this asset.
 
         Assets map "data_scope_name" (their name within the asset) to a Dataset (or dataset rid). The same type of
```
stefanv, line 101: I notice that `ref_name` is used in runs, but `data_scope_name` in assets. They both seem to be a mapping of 'name within X -> rid`. If these are fundamentally different concepts, perhaps something to be explained in the user guide. Otherwise, should the name be consistent?

stefanv (commented):
LGTM; would be good to have docstrings describe the new parameters.

nominal/core/asset.py
alkasm: Yeah you're right, this is a product decision that we need to make!

PR #391, opened 2025-06-23
feat: deprecate all top level functions in nominal.py

drake-nominal:
Use deprecation warnings to more forcefully push users towards the [profile migration guide](https://docs.nominal.io/python/profile-migration)

### DIFF
--- a/nominal/nominal.py
+++ b/nominal/nominal.py
@@ -71,14 +71,18 @@ def set_token(base_url: str, token: str) -> None:
     _config.set_token(base_url, token)
 
 
+def _get_default_client() -> NominalClient:
+    token = _config.get_token(_global_base_url)
+    return _get_or_create_connection(_global_base_url, token)
+
+
 @typing_extensions.deprecated(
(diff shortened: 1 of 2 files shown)


nominal/nominal.py
```diff
@@ -63,12 +71,20 @@ def set_token(base_url: str, token: str) -> None:
     _config.set_token(base_url, token)
 
 
+@typing_extensions.deprecated(
+    "nominal.get_default_client is deprecated and will be removed in a future version. "
+    "Use `nominal.NominalClient.from_profile` instead, see https://docs.nominal.io/python/profile-migration"
+)
 def get_default_client() -> NominalClient:
```
alkasm, line 78: can you move the impl of `get_default_client` to a private function without a deprecation warning, so that all of the funcs below which call `get_default_client` won't double (or triple, etc) warn?

alkasm: there may be other instances of this as well, haven't looked deeply

alkasm (approved):

PR #398, opened 2025-06-27
feat: overhaul data export from nominal

drake-nominal:
## Description:

Allow users to `get_read_stream` on a `DataSource`. 

Currently, I only support streaming pandas dataframes, but leave the framework available for adding other export types (e.g. pyarrow tables, polars, raw protos, etc.). 

I did some benchmarking on my machine using a sample dataset featuring really high rate data for 10 minutes of data. 
From these benchmarks, I was able to determine that with proper tuning, I was able to get my downloads to run significantly (~30%) faster by using process pools rather than thread pools, and was able to get much closer to saturating my network connection over the course of the download-- I suspect this is due to the gzipping at play. 

As a result, I let the users configure everything-- by default, we use a single threaded threadpool to perform requests, but let users upgrade their stream to an N-worker process pool in a very opt-in way if they'd like to receive improved performance.

## Benchmarks:

Test with 254,582,023 points (nans=0, non-nans=254,582,023) across 108 channels.

Both cases provide performance approx on par with 2 million points / second with ~100mbps internet.

### Process pool (16 workers, 5_000_000 points per request)

* Total time: ~110-130s

### Thread pool (16 workers, 5_000_000 points per request)

* Total time: 120-140s

### Thread pool (32 workers, 1_000_000 points per request)

* Total time: ~5 minutes

### Thread pool (16 workers, 10_000_000 points per request)

* Total time: 110s

### Thread pool (32 workers, 10_000_000 points/request, 300_000_000 points/batch)

* Total time: 120-130s

### Process pool (32 workers, 10_000_000 points/request, 300_000_000 points/batch)

* Total time: 90s

### DIFF
--- a/nominal/_utils/process_tools.py
+++ b/nominal/_utils/process_tools.py
@@ -1,6 +1,8 @@
 import concurrent.futures
 import contextlib
-from typing import Any, Generator, Literal, TypeAlias
+from typing import Any, Generator, Literal
+
+from typing_extensions import TypeAlias
 
 PoolType: TypeAlias = Literal["thread", "process"]
 
--- a/nominal/_utils/typing_tools.py
+++ b/nominal/_utils/typing_tools.py
@@ -1,5 +1,7 @@
 from collections.abc import Callable
-from typing import Any, ParamSpec, TypeVar, cast
+from typing import Any, TypeVar, cast
+
+from typing_extensions import ParamSpec
 
 T = TypeVar("T")
 P = ParamSpec("P")


pyproject.toml
```diff
@@ -35,6 +35,7 @@ dependencies = [
     "requests>=0.0.0",
     "ffmpeg-python>=0.2.0",
     "nominal-api==0.708.0",
+    "babel>=2.17.0",
```
alkasm, line 38: we already have `dateutil`--is that sufficient for the same use-case? I'd prefer to not add a new dep for a single function call if we can avoid it!

nominal/thirdparty/pandas/_pandas.py
```diff
@@ -215,36 +221,89 @@ def _to_pandas_timestamp(timestamp: Timestamp) -> pd.Timestamp:
...
+    ----
+        datasource: The datasource to download data from
+        channel_exact_match: Filter the returned channels to those whose names match all provided strings
+            (case insensitive).
+            For example, a channel named 'engine_turbine_rpm' would match against ['engine', 'turbine', 'rpm'],
+            whereas a channel named 'engine_turbine_flowrate' would not!
+        channel_fuzzy_search_text: Filters the returned channels to those whose names fuzzily match the provided
+            string.
+        channels: List of channels to fetch data for. If provided, supercedes search parameters of
+            `channel_exact_match` and `channel_fuzzy_search_text`.
```
alkasm, line 250: Bah I don't love this function signature (I know it was there before). Generally I try to avoid function signatures where the signature itself can't enforce the calling convention. I wonder if we can use `typing.overload` to enforce that if you call with `channels` you can't use the other args?

nominal/thirdparty/pandas/_pandas.py
```diff
@@ -215,36 +221,89 @@ def _to_pandas_timestamp(timestamp: Timestamp) -> pd.Timestamp:
...
+            datasource.search_channels(
+                exact_match=channel_exact_match,
+                fuzzy_search_text=channel_fuzzy_search_text,
+            )
+        )
+    elif channel_exact_match is not None or channel_fuzzy_search_text is not None:
+        logger.warning(
+            "'channel_exact_match' and 'channel_fuzzy_search_text' are ignored when a list of channels "
+            "are provided to 'datasource_to_dataframe'."
+        )
```
alkasm, line 289: ugh I really would prefer to throw here but I know this code already existed...

nominal/thirdparty/pandas/_pandas.py
```diff
@@ -215,36 +221,89 @@ def _to_pandas_timestamp(timestamp: Timestamp) -> pd.Timestamp:
...
+    start: str | datetime | ts.IntegralNanosecondsUTC | None = None,
+    end: str | datetime | ts.IntegralNanosecondsUTC | None = None,
+    tags: dict[str, str] | None = None,
+    *,
+    channels: Sequence[Channel] | None = None,
+    num_workers: int = 1,
+    channel_batch_size: int = 20,
+    relative_to: datetime | ts.IntegralNanosecondsUTC | None = None,
+    relative_resolution: ts._LiteralTimeUnit = "nanoseconds",
+) -> Generator[pd.DataFrame, None, None]:
```
alkasm, line 237: You can use `Iterable[pd.DataFrame]`

nominal/core/channel.py
```diff
@@ -214,70 +215,26 @@ def _decimate_request(
...
-        relative_resolution: _LiteralTimeUnit = "nanoseconds",
-        *,
-        enable_gzip: bool = True,
-    ) -> BinaryIO:
-        """Get the channel data as a CSV file-like object.
+    def _to_compute_series(self, tags: Mapping[str, str] | None = None) -> scout_compute_api.Series:
+        if tags:
+            series_tags = {key: scout_compute_api.StringConstant(literal=value) for key, value in tags.items()}
+        else:
+            series_tags = {}
```
alkasm, line 222: we should add a lint rule for this, but we should not conditionally define variables

nominal/core/export_stream.py
```diff
@@ -0,0 +1,161 @@
...
+import multiprocessing
+import time
+from multiprocessing.managers import ValueProxy
+from typing import BinaryIO, Generator, Generic, Mapping, Sequence, TypeVar, cast
+
+from nominal.core.channel import Channel
+from nominal.core.read_stream_base import ExportJob, ReadStreamBase, TimeRange
+from nominal.ts import (
+    _AnyTimestampType,
+)
```
alkasm, line 16:
```suggestion
from nominal.ts import _AnyTimestampType
```

nominal/core/export_stream.py
```diff
@@ -0,0 +1,161 @@
...
+    def export(
+        self,
+        channels: Sequence[Channel],
+        start_ns: int,
+        end_ns: int,
+        tags: Mapping[str, str] | None = None,
+        batch_duration: datetime.timedelta | None = None,
+        timestamp_type: _AnyTimestampType = "epoch_seconds",
+        buckets: int | None = None,
+        resolution_ns: int | None = None,
```
alkasm, line 35: let's use the nanosecond aliases here instead of suffixing the var names

nominal/_utils/process_tools.py
```diff
@@ -0,0 +1,21 @@
...
+
+from typing_extensions import TypeAlias
+
+PoolType: TypeAlias = Literal["thread", "process"]
+
+DEFAULT_POOL_TYPE: PoolType = "thread"
+
+
+@contextlib.contextmanager
+def BackgroundPool(
```
alkasm, line 13:
```suggestion
def background_pool(
```

nominal/_utils/process_tools.py
```diff
@@ -0,0 +1,21 @@
...
+    max_workers: int, pool_type: PoolType = DEFAULT_POOL_TYPE, **kwargs: Any
```
alkasm, line 14: simpler to use the pool type itself rather than doing the string indirection

nominal/core/datasource.py
```diff
@@ -142,6 +149,50 @@ def get_write_stream(
...
+        format: Literal["pandas"],
+        *,
+        points_per_request: int = DEFAULT_POINTS_PER_REQUEST,
+        points_per_batch: int = DEFAULT_POINTS_PER_BATCH,
+        channels_per_request: int = DEFAULT_CHANNELS_PER_REQUEST,
+        num_workers: int = 1,
+        pool_type: PoolType = DEFAULT_POOL_TYPE,
+    ) -> ExportStream[pd.DataFrame]: ...
+
+    def get_read_stream(
```
alkasm, line 164: I'm hesitant to name this as a "stream" as it's not really using a streaming API.

nominal/core/read_stream_base.py
```diff
@@ -0,0 +1,473 @@
...
+                    timestamp_type=timestamp_type,
+                )
+                for channel_group in channel_groups
+            ]
+
+            # For large channels, we need to subdivide the time range based on their data rates
+            for channel in large_channels:
+                channel_rate = points_per_second[channel.name]
+                sub_offset = datetime.timedelta(seconds=self._points_per_request / channel_rate)
+                sub_offset_ns = int(sub_offset.total_seconds() * 1e9)
```
alkasm, line 460: nit but the intermediate type doesn't really do anything for us

PR #402, opened 2025-07-02
feat: add description and rid to events

drake-nominal:
Adds the description field to events returned from the backend API. I also switch to using a `rid` instead of a `uuid` with an event, and stop using deprecated event service endpoints.

As part of this, I also deprecate usage of the `uuid` field in the message class, as well as retrieving events by uuid in the `NominalClient` class.

BREAKING CHANGE: `nominal.NominalClient.get_events(uuids=["uuid"...])` => `nominal.NominalClient.get_events(rids=["rid"...])`

### DIFF
--- a/nominal/core/client.py
+++ b/nominal/core/client.py
@@ -971,21 +971,13 @@ class NominalClient:
         return Event._from_conjure(self._clients, response)
 
     def get_event(self, rid: str) -> Event:
-        events = self.get_events_by_rid([rid])
+        events = self.get_events([rid])
         if len(events) != 1:
             raise RuntimeError(f"Expected to receive exactly one event, received {len(events)}")
 
         return events[0]
 
-    @deprecated(
-        "Getting an event by uuid is deprecated and will be removed in a future release. "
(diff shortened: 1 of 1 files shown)


nominal/core/client.py
```diff
@@ -968,10 +970,25 @@ def create_event(
...
+
+    @deprecated(
+        "Getting an event by uuid is deprecated and will be removed in a future release. "
+        "Use `get_events_by_rid` instead"
+    )
     def get_events(self, uuids: Sequence[str]) -> Sequence[Event]:
         responses = self._clients.event.get_events(self._clients.auth_header, event.GetEvents(list(uuids)))
         return [Event._from_conjure(self._clients, response) for response in responses]
 
+    def get_events_by_rid(self, rids: Sequence[str]) -> Sequence[Event]:
```
alkasm, line 988:
```suggestion
    def get_events(self, rids: Sequence[str]) -> Sequence[Event]:
```
?

alkasm (approved):

PR #407, opened 2025-07-04
feat: expose workbooks and workbook templates more fully

drake-nominal:

### DIFF
--- a/nominal/core/workbook.py
+++ b/nominal/core/workbook.py
@@ -40,9 +40,19 @@ class Workbook(HasRid):
     rid: str
     title: str
     description: str
-    run_rids: Sequence[str] | None
-    asset_rids: Sequence[str] | None
     workbook_type: WorkbookType
+
+    run_rids: Sequence[str] | None
+    """Mutually exclusive with `asset_rids`.
+
+    May be empty when a workbook is a fresh comparison workbook.
+    """
+
+    asset_rids: Sequence[str] | None
+    """Mutually exclusive with `run_rids`.
+
+    May be empty when a workbook is a fresh comparison workbook.
+    """
 
     _clients: _Clients = field(repr=False)
 


nominal/core/workbook.py
```diff
@@ -27,6 +31,65 @@ def nominal_url(self) -> str:
...
+        )
+
+    def lock(self) -> None:
+        """Locks the workbook, preventing changes from being made to it."""
+        self._clients.notebook.lock(self._clients.auth_header, self.rid)
+
+    def unlock(self) -> None:
+        """Unlocks the workbook, allowing changes to be made to it."""
+        self._clients.notebook.unlock(self._clients.auth_header, self.rid)
+
```
alkasm, line 92: are these ops idempotent?

alkasm: I expect that users will generally want to write code like:

```py
workbook = get_workbook()
workbook.unlock()
workbook.update_refnames()
workbook.lock()
```

which is fine, but just want to make sure that the first `.unlock()` isn't going to conditionally throw

nominal/core/workbook.py
```diff
@@ -16,6 +17,9 @@ class Workbook(HasRid):
     title: str
     description: str
     run_rid: str | None
+    asset_rids: Sequence[str] | None
+    is_draft: bool
+    is_locked: bool
```
alkasm, line 22: since you can mutate these things with methods, can we turn these into API calls instead? basically don't want to allow for inconsistencies here

nominal/core/workbook.py
```diff
@@ -1,21 +1,49 @@
...
+
 
 @dataclass(frozen=True)
 class Workbook(HasRid):
     rid: str
     title: str
     description: str
-    run_rid: str | None
+    run_rids: Sequence[str] | None
+    asset_rids: Sequence[str] | None
```
alkasm, line 44: Let's document that these are generally mutually exclusive, which is why they're optional

alkasm (approved):

PR #411, opened 2025-07-14
feat: allow downloading dataset files

drake-nominal:


nominal/_utils/download_tools.py
```diff
@@ -0,0 +1,28 @@
+import pathlib
+from urllib.parse import unquote, urlparse
+
+import httpx
```
alkasm, line 4: why httpx instead of requests? you can stream with requests as well fyi

alkasm: fwiw i actually like httpx's interface a lot but used requests for pairity with the conjure clients

nominal/core/dataset_file.py
```diff
@@ -36,8 +50,69 @@ def _from_conjure(cls, dataset_file: scout_catalog.DatasetFile) -> Self:
             uploaded_at=upload_time,
             ingested_at=ingest_time,
             ingest_status=IngestStatus._from_conjure(dataset_file.ingest_status),
+            _clients=clients,
         )
 
+    def download(self, destination: pathlib.Path) -> pathlib.Path:
```
alkasm, line 56: IIUC destination is a directory, the return is the actual file path? let's maybe use `output_dir` or something so that it's clear it should be a dir?

nominal/core/dataset_file.py
```diff
@@ -36,8 +50,69 @@ def _from_conjure(cls, dataset_file: scout_catalog.DatasetFile) -> Self:
...
+
+        Args:
+            destination: If an existing directory, downloads the file to the given directory.
+                Otherwise, downloads the file to the given path.
+
+        Returns:
+            Path that the file was downloaded to
+
+        Raises:
+            FileExistsError: File already exists at destination
```
alkasm, line 67: might be useful to have a `force` or `overwrite` func param

nominal/core/dataset_file.py
```diff
@@ -36,8 +50,69 @@ def _from_conjure(cls, dataset_file: scout_catalog.DatasetFile) -> Self:
...
+            RuntimeError: Error downloading file
+        """
+        uri = self._clients.catalog.get_dataset_file_uri(self._clients.auth_header, self.dataset_rid, self.id).uri
+        logger.info("Downloading %s (%s) => %s", self.name, uri, destination)
+        return download_presigned_uri(uri, destination)
+
+    def download_origin_files(self, destination: pathlib.Path) -> Sequence[pathlib.Path]:
```
alkasm, line 74: is it important to name this "origin" files?

nominal/core/dataset_file.py
```diff
@@ -36,8 +50,69 @@ def _from_conjure(cls, dataset_file: scout_catalog.DatasetFile) -> Self:
...
+
+        if destination.exists():
+            if len(origin_uris) == 1 and destination.is_file():
+                raise FileExistsError(f"Cannot download origin file to {destination}: already exists!")
+            elif destination.is_file():
+                raise FileExistsError(f"Cannot download origin files to {destination}: already exists as a file!")
+        elif len(origin_uris) == 1:
+            destination.parent.mkdir(parents=True, exist_ok=True)
+        else:
+            destination.mkdir(parents=True, exist_ok=True)
```
alkasm, line 101: I don't understand why you would create the parent but not the dir itself if origin_uris is 1

nominal/core/dataset_file.py
```diff
@@ -36,8 +50,69 @@ def _from_conjure(cls, dataset_file: scout_catalog.DatasetFile) -> Self:
...
+
+        with concurrent.futures.ThreadPoolExecutor(max_workers=len(origin_uris)) as pool:
```
alkasm, line 103: can we put a reasonable limit here?

nominal/_utils/download_tools.py
```diff
@@ -0,0 +1,28 @@
...
+    uri: str, destination: pathlib.Path, chunk_size: int = 8192, staleness_timeout: float = 30.0
+) -> pathlib.Path:
+    if destination.exists():
+        if destination.is_dir():
+            filename = unquote(pathlib.Path(urlparse(uri).path).name).replace(":", "-")
+            destination = destination / filename
+        else:
+            raise FileExistsError(f"Cannot download {uri} to {destination}-- already exists!")
+    else:
+        destination.parent.mkdir(parents=True, exist_ok=True)
```
alkasm, line 17: again, why create the parent of the output dir and not the output dir?

nominal/core/dataset_file.py
```diff
@@ -36,8 +52,138 @@ def _from_conjure(cls, dataset_file: scout_catalog.DatasetFile) -> Self:
...
+            Path that the file was downloaded to
+
+        Raises:
+            FileNotFoundError: Output directory doesn't exist and force=False
+            FileExistsError: File already exists at destination
+            ValueError: Output directory exists and is not a directory
+            RuntimeError: Error downloading file
+        """
+        if output_directory.exists() and not output_directory.is_dir():
+            raise ValueError(f"Output directory is not a directory: {output_directory}")
```
alkasm, line 79: really not important but we could use `NotADirectoryError`

nominal/core/dataset_file.py
```diff
@@ -36,8 +52,138 @@ def _from_conjure(cls, dataset_file: scout_catalog.DatasetFile) -> Self:
...
+        for attempt in range(num_retries):
+            origin_uris = self._presigned_origin_files()
+            origin_uri = None
+            for api_uri in origin_uris:
+                if api_uri.path == origin_path:
+                    origin_uri = api_uri.uri
+                    break
+
+            if origin_uri is None:
+                raise RuntimeError(f"No such origin path: {origin_path}")
```
alkasm, line 103: can we just return a dict or something instead of needing to have an optional that is later assigned and checked?

nominal/core/dataset_file.py
```diff
@@ -36,8 +52,138 @@ def _from_conjure(cls, dataset_file: scout_catalog.DatasetFile) -> Self:
...
+        ) from last_exception
+
+    def download_original_files(
+        self, output_directory: pathlib.Path, force: bool = True, parallel_downloads: int = 8, num_retries: int = 3
+    ) -> Sequence[pathlib.Path]:
+        """Download the input file(s) for a containerized extractor to a destination on local disk.
+
+        Args:
+            output_directory: Download file(s) to the given directory
+            force: If true, delete any files that exist / create parent directories if nonexistant
```
alkasm, line 140:
```suggestion
            force: If true, delete any files that exist / create parent directories if nonexistent
```

alkasm (approved):

PR #413, opened 2025-07-17
feat: allow searching datasets

ken-iks:

### DIFF
--- a/nominal/core/client.py
+++ b/nominal/core/client.py
@@ -242,6 +242,9 @@ class NominalClient:
         search_text: str | None = None,
         labels: Sequence[str] | None = None,
         properties: Mapping[str, str] | None = None,
+        before: str | datetime | IntegralNanosecondsUTC | None = None,
+        after: str | datetime | IntegralNanosecondsUTC | None = None,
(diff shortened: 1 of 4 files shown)


nominal/core/client.py
```diff
@@ -229,6 +232,38 @@ def search_users(self, exact_match: str | None = None, search_text: str | None =
         query = create_search_users_query(exact_match, search_text)
         return list(self._iter_search_users(query))
 
+    def _iter_search_datasets(self, query: scout_catalog.SearchDatasetsQuery) -> Iterable[Dataset]:
+        for raw_dataset in search_datasets_paginated(self._clients.catalog, self._clients.auth_header, query):
+            yield Dataset._from_conjure(self._clients, raw_dataset)
+
+    def search_datasets(
+        self,
+        exact_match: str | None = None,
```
alkasm, line 241:
```suggestion
        self,
        *,
        exact_match: str | None = None,
```

alkasm: forces callers to use kw-args to call. especially useful for these APIs that have a lot of string parameters in a row, which is unclear at the callsite otherwise

PR #424, opened 2025-07-25
feat: creating templates from yaml config files.

ken-iks:
First shot at programmatic template creation. Works for only timeseries panels right now. Error handling not so great but generally curious about if I am thinking about it the right way or if I should go in a different direction. Proposed structure for yaml would be something like this for example (still **very** open to change): 

version: <int>
name: "<template_title>"
labels: [<list_of_string_labels>]

tabs:
  "<tab_name_1>":
    panels:
      - type: "<panel_type>"
        rows:
          "<row_name_1>":
            <channel_name_1>:
              - "<color>"
              - "<axis_title>"
              - "<axis_side>"
            <channel_name_2>:
              - "<color>"
              - "<axis_title>"
              - "<axis_side>"
          "<row_name_2>":
            <channel_name_3>:
              - "<color>"
              - "<axis_title>"
              - "<axis_side>"
        comparison_runs:
          <run_name_1>:
            - "<color>"
            - "<run_rid>"
          <run_name_2>:
            - "<color>"
            - "<run_rid>"

### DIFF
--- a/nominal/experimental/templates/template_generation.py
+++ b/nominal/experimental/templates/template_generation.py
@@ -1,0 +1,447 @@
+import json
+import uuid
+from dataclasses import dataclass
+from enum import Enum
+
+import yaml
+from nominal_api import (
+    api,
+    scout_channelvariables_api,
+    scout_chartdefinition_api,
+    scout_comparisonrun_api,
+    scout_layout_api,
+    scout_rids_api,
+    scout_template_api,
+    scout_workbookcommon_api,
+)
+
+from nominal.core.client import NominalClient
+from nominal.core.workbook_template import WorkbookTemplate
+
+CHART_RID_BASE = "ri.scout.cerulean-staging.chart."
+
+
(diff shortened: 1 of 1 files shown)


najork:
I'm philosophically OK with this going into `experimental` as a short-term solution, but we should start exploring first-class infrastructure as code solutions for declarative state management, e.g. https://developer.hashicorp.com/terraform

Defining Nominal resources in a source of truth outside of Nominal itself will be a common workflow. Users will want to define things like templates, checklists, etc externally and have a mechanism for syncing them with Nominal, with support for other nice things like GitOps workflows, etc.

Leveraging existing tooling to the extent possible reduces the product surface area that we need to own and maintain.

Eventually (and this is looking far out on the horizon) we can programmatically generate most of this, i.e. have imperative state management operations (CRUD+List) exposed and generated through the CLI and then have declarative state management implemented in a tool like Terraform.

PR #430, opened 2025-07-31
feat: remove logsets

drake-nominal:


nominal/__init__.py
```diff
@@ -97,8 +91,6 @@
     "Run",
     "User",
     "Video",
-    "LogSet",
-    "Log",
```
ross-barrett, line 101: Why log too?

ross-barrett: Or is this tied to the old notion of a logset? Not logsv2?

PR #439, opened 2025-08-14
feat: add experimental compute impl

alkasm:
See README

### DIFF
--- a/nominal/experimental/compute/dsl/_node_impls.py
+++ b/nominal/experimental/compute/dsl/_node_impls.py
@@ -1,0 +1,517 @@
+from __future__ import annotations
+import typing
+from dataclasses import dataclass
+from nominal_api import scout_compute_api
+
+from nominal.experimental.compute.dsl import nodes, params
+
+
+@dataclass(frozen=True)
+class AbsNode(nodes.NumericNode):
+    _node: nodes.NumericNode
+
+    def _to_conjure(self) -> scout_compute_api.NumericSeries:
+        return scout_compute_api.NumericSeries(
+            arithmetic=scout_compute_api.ArithmeticSeries(
+                expression="|input|",
+                inputs={"input": self._node._to_conjure()},
+            )
+        )
+
+
+@dataclass(frozen=True)
+class AcosNode(nodes.NumericNode):
+    _node: nodes.NumericNode
+
+    def _to_conjure(self) -> scout_compute_api.NumericSeries:
+        return scout_compute_api.NumericSeries(
+            arithmetic=scout_compute_api.ArithmeticSeries(
+                expression="acos(input)",
+                inputs={"input": self._node._to_conjure()},
+            )
+        )
+
+
+@dataclass(frozen=True)
+class AsinNode(nodes.NumericNode):
+    _node: nodes.NumericNode
+
+    def _to_conjure(self) -> scout_compute_api.NumericSeries:
+        return scout_compute_api.NumericSeries(
+            arithmetic=scout_compute_api.ArithmeticSeries(
+                expression="asin(input)",
+                inputs={"input": self._node._to_conjure()},
+            )
+        )
+
+
+@dataclass(frozen=True)
+class Atan2Node(nodes.NumericNode):
+    _y_node: nodes.NumericNode
+    _x_node: nodes.NumericNode
+
+    def _to_conjure(self) -> scout_compute_api.NumericSeries:
+        return scout_compute_api.NumericSeries(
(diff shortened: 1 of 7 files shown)


nominal/experimental/compute/dsl/params.py
```diff
@@ -0,0 +1,65 @@
+from __future__ import annotations
+
+import typing
+
+from nominal_api import api, scout_compute_api, scout_run_api
+
+Nanoseconds = int
```
ross-barrett, line 7: Maybe worth validating this is positive where used and expected

nominal/experimental/compute/dsl/_node_impls.py
```diff
@@ -0,0 +1,519 @@
...
+            unary_arithmetic=scout_compute_api.UnaryArithmeticSeries(
+                input=self._node._to_conjure(),
+                operation=scout_compute_api.UnaryArithmeticOperation.LOG,
+            )
+        )
+
+
+@dataclass(frozen=True)
+class MaxNode(nodes.NumericNode):
+    _inputs: typing.Sequence[nodes.NumericNode]
```
ross-barrett, line 200: If you care about being consistent with other sequence inputs, SumNode uses _nodes

nominal/experimental/compute/dsl/nodes.py
```diff
@@ -0,0 +1,153 @@
...
+
+
+@dataclass(frozen=True)
+class NumericNode(Node):
+    @abstractmethod
+    def _to_conjure(self) -> scout_compute_api.NumericSeries:
+        raise NotImplementedError()
+
+    @classmethod
+    def channel(cls, asset_rid: str, data_scope_name: str, channel_name: str) -> NumericNode:
```
ross-barrett, line 24: Do you want a positional only denotation here too for consistency?

nominal/experimental/compute/dsl/nodes.py
```diff
@@ -0,0 +1,153 @@
...
+    def floor_divide(self, right_node: NumericNode, /) -> NumericNode:
+        return _node_impls.FloorDivideNode(_left_node=self, _right_node=right_node)
+
+    def integral(self, /, start_timestamp: params.NanosecondsUTC, time_unit: params.TimeUnitLiteral) -> NumericNode:
+        return _node_impls.IntegralNode(_node=self, _start_timestamp=start_timestamp, _time_unit=time_unit)
+
+    def ln(self, /) -> NumericNode:
+        return _node_impls.LnNode(_node=self)
+
+    def logarithm(self, /) -> NumericNode:
```
ross-barrett, line 65: Should we call this log10? since we're shortening ln?

nominal/experimental/compute/dsl/nodes.py
```diff
@@ -0,0 +1,153 @@
...
+    def _to_conjure(self) -> scout_compute_api.NumericSeries:
+        raise NotImplementedError()
+
+    @classmethod
+    def channel(cls, asset_rid: str, data_scope_name: str, channel_name: str) -> NumericNode:
+        return _node_impls.NumericChannel(
+            _asset_rid=asset_rid, _data_scope_name=data_scope_name, _channel_name=channel_name
+        )
+
+    def abs(self, /) -> NumericNode:
```
ross-barrett, line 29: We'll want docstrings for these, but can add later

nominal/experimental/compute/dsl/nodes.py
alkasm: the rule is: nodes are positional-only. other args can be positional or kw

nominal/experimental/compute/dsl/params.py
```diff
@@ -0,0 +1,65 @@
...
+NanosecondsUTC = int
+DoubleConstant = float
+TimeUnitLiteral = typing.Literal["ns", "us", "ms", "s", "m", "h", "d"]
+ThresholdOperatorLiteral = typing.Literal[">", ">=", "<", "<=", "==", "!="]
```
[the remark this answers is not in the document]
alkasm: it will fail if any other value is provided, because it does a dict lookup with these specific strings

nominal/experimental/compute/dsl/params.py
alkasm: 🤷🏼 eh

nominal/experimental/compute/dsl/nodes.py
alkasm: I guess we can use `ln()` and `log()` for the method names--it's just the actual impl node that would be poorly named `LogNode`

PR #440, opened 2025-08-17
feat: expose object creation times across client

drake-nominal:


varun-nominal (approved):

varun-nominal:
do dataclass property comments show up in docs? might be worth adding so people know the units

PR #447, opened 2025-08-22
feat: return datasetfile when adding data to a dataset, allow polling on a per file basis

drake-nominal:

### DIFF
--- a/nominal/core/dataset_file.py
+++ b/nominal/core/dataset_file.py
@@ -291,7 +291,7 @@ class DatasetFile:
             logger.debug("Sleeping for %f seconds before polling for ingest status", interval.total_seconds())
             time.sleep(interval.total_seconds())
 
-        return self.refresh()
+        return self
 
 
 class IngestStatus(Enum):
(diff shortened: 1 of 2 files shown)


nominal/core/dataset_file.py
```diff
@@ -241,6 +251,44 @@ def download_original_files(
...
+            self.refresh()
+            if self.ingest_status is IngestStatus.SUCCESS:
+                break
+            elif self.ingest_status is IngestStatus.IN_PROGRESS:
+                pass
+            elif self.ingest_status is IngestStatus.FAILED:
+                # Get error message to display to user
+                raw_file = self._clients.catalog.get_dataset_file(self._clients.auth_header, self.dataset_rid, self.id)
+                file_error = raw_file.ingest_status.error
+                if file_error is None:
```
alkasm, line 270: This is racy, I think we could solve with a `_refresh(ingest_status)` function so that we can use the same response to refresh?

This is different from the datasource poll method, which doesn't refresh on each iteration. This seems more correct though, would you be down to update that?

alkasm (approved):

PR #461, opened 2025-09-05
feat: improve default workspace handling in search methods

drake-nominal:
* make `client.get_workspace()` return either the workspace rid in the client profile, or if not there, use the `default_workspace` as provided by the backend.
* Search methods now default to the same deduction process if no workspace rid is provided.

A future followup that would be good would be to make all of the search methods consistently take a `workspace: Workspace | str | None` instead of some having `workspace_rid: str | None`, but I leave that for a future PR.

### DIFF
--- a/nominal/core/client.py
+++ b/nominal/core/client.py
@@ -237,7 +237,7 @@ class NominalClient:
             search_rid = workspace.rid
         elif isinstance(workspace, str):
             search_rid = workspace
-        elif workspace == WorkspaceSearchType.ALL:
+        elif workspace is WorkspaceSearchType.ALL:
             return None
         elif workspace is WorkspaceSearchType.DEFAULT:
             search_rid = None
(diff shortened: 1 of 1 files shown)


nominal/core/client.py
```diff
@@ -109,6 +110,17 @@
 DEFAULT_CONNECT_TIMEOUT = timedelta(seconds=30)
 
 
+class ALL_WORKSPACES:
+    """Sentinel value for requesting results from all workspaces."""
+
+
+class DEFAULT_WORKSPACE:
+    """Sentinel value for requesting results from the default configured workspace"""
```
alkasm, line 118: Let's use a `WorkspaceSearch` enum or something similar

nominal/core/client.py
```diff
@@ -212,6 +222,77 @@ def __repr__(self) -> str:
...
+        try:
+            # NOTE: raises a conjure exception if the given rid is not visible to the user (or doesn't exist period)
+            resolved_workspace = self.get_workspace(search_rid)
+            return resolved_workspace.rid
+        except NominalConfigError:
+            # re-raising with a more specific exception message
+            raise NominalConfigError(
+                "WorkspaceSearchType.DEFAULT provided for workspace rid, but no default configured. "
+                "Specify a workspace_rid within your config profile (see `nom config profile --help`), "
+                "specify a workspace_rid manually, or contact your Nominal representative to set a default "
```
alkasm, line 256:
```suggestion
                "specify a workspace RID manually, or contact your Nominal representative to set a default "
```
(since the argument isn't `workspace_rid` anymore)

alkasm (approved):

PR #471, opened 2025-09-13
feat: add experimental compute modules

alkasm:


nominal/core/_clientsbunch.py
```diff
@@ -142,6 +148,11 @@ class ClientsBunch:
     workspace: security_api_workspace.WorkspaceService
     containerized_extractors: ingest_api.ContainerizedExtractorService
     secrets: secrets_api.SecretService
+    module: module.ModuleService
+
+    def client_factory(self, service_cls: Type[ServiceT]) -> ServiceT:
+        factory = create_conjure_client_factory(user_agent=self.user_agent, service_config=self.service_config)
+        return factory(service_cls)
```
alkasm, line 155: This has nothing to do with modules. I will expand this in a future PR, but here the idea is instead of creating dozens of services and holding onto them, instead any code can simply construct the clients it needs while injecting the same service config.

PR #472, opened 2025-09-15
feat: allow using truststore for ssl bypass

drake-nominal:
I had chatgpt describe the context & solution proposed by this PR

# Why Python Does Not Use the Operating System Trust Store by Default

This document explains:
- what the original problem was,
- why Python behaves this way by default (with historical context),
- how the `truststore` library works, and
- why using `truststore.SSLContext` is the correct solution in a corporate network.

---

## What the Problem Was

Our corporate network sits in the middle of secure (TLS/HTTPS) connections in order to inspect and secure traffic.  
As a result, applications are not talking directly to external services; they are talking through a company-managed security system.

The operating system is configured to trust this system via installed root certificates.  
Most software—browsers in particular—automatically uses the operating system’s trust store and therefore works without issue.

Python, by default, does not.

When a Python program attempts to make a secure connection, it evaluates the connection using its own list of trusted certificate authorities. Since the corporate root certificate is not included in that list, Python rejects the connection as untrusted.

Nothing is actually wrong with the network or the service—Python is simply unaware of the organization’s trust decisions.

---

## Historical Context: Why Python Works This Way

Python’s TLS behavior is the result of early design decisions made when:

- Operating systems did not expose stable, portable APIs for accessing trusted certificates
- Certificate storage locations varied widely across Linux distributions
- Windows and macOS certificate APIs were complex, inconsistent, or poorly documented
- Python prioritized portability and deterministic behavior across platforms

To avoid platform-specific complexity, Python chose to ship and rely on its own certificate authority (CA) bundle rather than depend on the operating system.

This approach provided:
- consistent behavior across machines,
- reproducible results in different environments, and
- fewer platform-specific bugs.

At the time, this was a conservative and reasonable security decision.

---

## Why Browsers Behave Differently

Web browsers evolved under different constraints:

- They are deeply integrated with the operating system
- They explicitly support enterprise security policies
- They synchronize with system trust stores by design

As corporate TLS interception became widespread, browsers adapted.  
Python, as a general-purpose runtime rather than a managed client application, did not.

This difference explains why browsers typically work seamlessly behind corporate firewalls while Python applications often require additional configuration.

---

## The Modern Mismatch

In modern corporate environments:

- TLS interception is intentional and sanctioned
- Custom root certificates are installed at the OS level
- Trust policy is centrally managed

Python, however:
- continues to rely on a bundled CA list (often via `certifi`),
- ignores OS-installed enterprise certificates by default, and
- has no built-in awareness of corporate trust policy.

From Python’s perspective, intercepted connections genuinely appear unsafe.  
This is not a bug—it is a legacy design assumption colliding with modern infrastructure.

---

## How `truststore.SSLContext` Works

The `truststore` library provides an explicit, opt-in mechanism for Python to use the operating system’s trust store.

Instead of globally modifying Python’s SSL behavior, `truststore.SSLContext` creates an SSL context that:

- On Windows, consults the Windows Certificate Store
- On macOS, consults the system Keychain
- On Linux, consults the system trust store (e.g. `/etc/ssl/certs`)

Only connections that explicitly use this context inherit the operating system’s trust decisions.

This makes the trust relationship:
- explicit,
- scoped, and
- easy to reason about.

Example usage:

```python
import truststore
import ssl

ssl_context = truststore.SSLContext(ssl.PROTOCOL_TLS_CLIENT)
```

### DIFF
--- a/nominal/core/_utils/networking.py
+++ b/nominal/core/_utils/networking.py
@@ -1,20 +1,69 @@
 from __future__ import annotations
 
 import gzip
+import logging
 import os
+import ssl
 from typing import Any, Callable, Mapping, Type, TypeVar
 
 import requests
+import truststore
 from conjure_python_client import ServiceConfiguration
-from conjure_python_client._http.requests_client import RetryWithJitter, TransportAdapter
-from requests.adapters import CaseInsensitiveDict
+from conjure_python_client._http.requests_client import KEEP_ALIVE_SOCKET_OPTIONS, RetryWithJitter
+from requests.adapters import DEFAULT_POOLSIZE, CaseInsensitiveDict, HTTPAdapter
+from urllib3.connection import HTTPConnection
+from urllib3.util.retry import Retry
+
+logger = logging.getLogger(__name__)
 
 T = TypeVar("T")
 
 GZIP_COMPRESSION_LEVEL = 1
 
 
-class GzipRequestsAdapter(TransportAdapter):
+class SslBypassRequestsAdapter(HTTPAdapter):
+    """Transport adapter that allows customizing SSL options and forwarding host truststore.
+
+    NOTE: based on a combination of injecting `truststore.SSLContext` into
+        `conjure_python_client._http.requests_client.TransportAdapter`.
+    """
+
+    ENABLE_KEEP_ALIVE_ATTR = "_enable_keep_alive"
+    __attrs__ = [*HTTPAdapter.__attrs__, ENABLE_KEEP_ALIVE_ATTR]
+
+    def __init__(self, *args: Any, enable_keep_alive: bool = False, **kwargs: Any):
+        self._enable_keep_alive = enable_keep_alive
+        super().__init__(*args, **kwargs)
+
+    def init_poolmanager(
+        self,
+        connections: int,
+        maxsize: int,
+        block: bool = False,
+        **pool_kwargs: Mapping[str, Any],
+    ) -> None:
+        """Wrapper around the standard init_poolmanager from HTTPAdapter with modifications
+        to support keep-alive settings and injecting SSL context.
+        """
+        if self._enable_keep_alive:
+            keep_alive_kwargs: dict[str, Any] = {
+                "socket_options": [
+                    *HTTPConnection.default_socket_options,
+                    *KEEP_ALIVE_SOCKET_OPTIONS,
+                ]
+            }
+            pool_kwargs = {**pool_kwargs, **keep_alive_kwargs}
+
+        pool_kwargs["ssl_context"] = truststore.SSLContext(ssl.PROTOCOL_TLS_CLIENT)
+
+        super().init_poolmanager(connections, maxsize, block, **pool_kwargs)  # type: ignore[no-untyped-call]
+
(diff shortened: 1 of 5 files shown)


pyproject.toml
```diff
@@ -71,6 +71,7 @@ dev = [
     "types-tabulate>=0.9.0.20241207,<0.10",
     "nominal-api-protos>=0.708.0",
     "coverage[toml]>=7.0.0",
+    "truststore>=0.10.4; python_version >= '3.10'",
```
stefanv, line 74: `truststore` is being imported in the main script, but isn't being installed for Python 3.9.

For consistency, easiest to drop Python 3.9 support (it's EOL October). If that's not an option, then either: (a) install it on all versions or (b) put a guard around the import and context configuration.

stefanv:
This doesn't work for me locally, because the mechanism used to override `ssl_context` seems unsupported (`conjure-python-client` recreates certificate, then passes it in twice). Will see if I can add a workaround.

stefanv:
I suspect the tests here aren't exercising the added functionality, and that we need the following patch to `conjure_python_client`:

```patch
diff --git a/conjure_python_client/_http/requests_client.py b/conjure_python_client/_http/requests_client.py
index 4d77622..f22ecc5 100644
--- a/conjure_python_client/_http/requests_client.py
+++ b/conjure_python_client/_http/requests_client.py
@@ -234,7 +234,8 @@ class TransportAdapter(HTTPAdapter):
         self._pool_connections = connections
         self._pool_maxsize = maxsize
         self._pool_block = block
-        ssl_context = create_urllib3_context(ciphers=CIPHERS)
+        if 'ssl_context' not in pool_kwargs:
+            pool_kwargs['ssl_context'] = create_urllib3_context(ciphers=CIPHERS)
         keep_alive_pool_kwargs = {
             "socket_options": HTTPConnection.default_socket_options
             + KEEP_ALIVE_SOCKET_OPTIONS
@@ -245,7 +246,6 @@ class TransportAdapter(HTTPAdapter):
             num_pools=connections,
             maxsize=maxsize,
             block=block,
-            ssl_context=ssl_context,
             **pool_kwargs
         )
```

I can file that, or something similar, if you agree.

nominal/core/_utils/networking.py
```diff
@@ -131,15 +180,31 @@ def create_conjure_client_factory(
...
+    num_retries: int = 5,
+) -> requests.Session:
+    retries = Retry(
+        total=num_retries,
+        backoff_factor=0.5,
+        status_forcelist=(429, 500, 502, 503, 504),
+    )
+    session = requests.Session()
+    adapter = SslBypassRequestsAdapter(max_retries=retries, pool_maxsize=pool_size)
+    session.mount("https://", adapter)
```
alkasm: Should we mount both adapters to the same session?

pyproject.toml
```diff
@@ -32,6 +32,7 @@ dependencies = [
     "nominal-streaming==0.5.8",
     "rich>=14.1.0",
     "tabulate>=0.9.0,<0.10",
+    "truststore>=0.10.4",
```
alkasm: is the particular version important?

PR #479, opened 2025-09-29
feat: expose channel tags present, bump api version

drake-nominal:
This PR adds in a method i've had to use the private API for in scripts:

* searching for available tags (given an initial tag filter) on a channel in a period of time

To do this, I had to bump the nominal API version, and also the mypy version to fix a bug that was occurring in 0.17.1 (what had auto-installed for me) after doing the above (some completely nondescript error in the mypy main method...)

I also had to fix some parts of the compute expression library that had become broken by changes in the API

### DIFF
--- a/nominal/core/channel.py
+++ b/nominal/core/channel.py
@@ -11,6 +11,7 @@ from nominal_api import (
     datasource_api,
     scout_compute_api,
     scout_dataexport_api,
+    scout_datasource,
     timeseries_channelmetadata,
     timeseries_channelmetadata_api,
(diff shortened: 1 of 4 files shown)


nominal/core/dataset.py
```diff
@@ -124,6 +126,33 @@ def update(
 
         return self.refresh()
 
+    def update_bounds(
```
varun-nominal, line 129: what use do you have for this? in my ahead this was always one of those endpoints taht should only be called internally by our services

varun-nominal (approved):

PR #531, opened 2025-11-19
feat: forcefully remove top-level functionality

drake-nominal:


stefanv (commented):
Thanks for handling all these outstanding deprecations. After the next sync with main, we should get it merged so it doesn't drift into conflicts again.

Are there any plans for end-to-end or other testing?

drake-nominal:
Replaced by #738

PR #672, opened 2026-03-19
fix: don't record intermediate templates used for workbook migration

seanmreidy:
Implements part of MDEV-88

## Description

We shouldn't record workbook templates that were temporarily created for workbook migration.

## Testing
N/A


watwea-nominal (changes requested):
I agree with the bots here, this isn't the right level of abstraction. The implementation of skipping the mapping record should be pushed down. I made this commit which you can look at / cherry-pick / modify which tries to do this. b4d379b6eb68397166e87c6789d5130c55367592

seanmreidy:
No longer needed, since templates are no longer used for migration per https://github.com/nominal-io/nominal-client/pull/775

PR #683, opened 2026-03-23
feat: e2e testing for migration

seanmreidy:
## Summary

Add end to end testing for the top-level migrator methods.  This covers complete migration from scratch, migration in a partially completed state, and an idempotency check to verify that our continue migration path works as expected.

This testing suite covers assertions of resources created under the hood, their top-level properties, and their linkages with other resources. We don't have extensive testing of specific resource migrators (like DatasetMigrator) but can easily create them as a fast follow reusing the same code to set up / break down resources.

## Testing

CI workflows.

### DIFF
--- a/tests/e2e/migration/conftest.py
+++ b/tests/e2e/migration/conftest.py
@@ -19,7 +19,8 @@ for cleanup (archive) after each test, even on failure.
 from __future__ import annotations
 
 from io import BytesIO
-from typing import Callable
+from typing import Callable, Iterator
+from unittest import mock
 from uuid import uuid4
 
 import pytest
@@ -85,6 +86,17 @@ def dest_client(pytestconfig) -> NominalClient:
     base_url = pytestconfig.getoption("dest_base_url") or pytestconfig.getoption("base_url")
     print(f"Using dest NominalClient.create(base_url={base_url!r})")
     return NominalClient.create(base_url=base_url, token=auth_token)
+
+
+@pytest.fixture(scope="session", autouse=True)
+def set_connection(dest_client: NominalClient) -> Iterator[None]:
+    """Override the parent conftest's set_connection to use dest_client.
+
+    The parent fixture requires ``--profile`` which migration tests don't supply;
+    this shadows it so the module-level default client points at the destination env.
+    """
+    with mock.patch("nominal.nominal.get_default_client", return_value=dest_client):
+        yield
 
 
 @pytest.fixture
--- a/tests/e2e/migration/conftest.py
+++ b/tests/e2e/migration/conftest.py
@@ -39,13 +39,13 @@ def pytest_addoption(parser):
     parser.addoption("--source-profile", default=None, help="Source Nominal profile name (e.g. production)")
     parser.addoption("--source-auth-token", default=None, help="Source auth token (used with --source-base-url)")
     parser.addoption(
-        "--source-base-url", default="https://api.gov.nominal.io", help="Source base URL (default: production)"
+        "--source-base-url", default="https://api.gov.nominal.io/api", help="Source base URL (default: production)"
     )
     parser.addoption("--dest-profile", default=None, help="Destination Nominal profile name (e.g. staging)")
     parser.addoption("--dest-auth-token", default=None, help="Destination auth token (used with --dest-base-url)")
     parser.addoption(
         "--dest-base-url",
-        default="https://api-staging.gov.nominal.io",
+        default="https://api-staging.gov.nominal.io/api",
         help="Destination base URL (default: staging)",
     )
 
--- a/tests/e2e/migration/conftest.py
+++ b/tests/e2e/migration/conftest.py
@@ -37,7 +37,7 @@ def pytest_addoption(parser):
     parser.addoption("--source-profile", default=None, help="Source Nominal profile name (e.g. production)")
     parser.addoption("--source-auth-token", default=None, help="Source auth token (used with --source-base-url)")
     parser.addoption(
-        "--source-base-url", default="https://api.nominal.io", help="Source base URL (default: production)"
+        "--source-base-url", default="https://api.gov.nominal.io", help="Source base URL (default: production)"
     )
 
 
--- a/tests/e2e/migration/conftest.py
+++ b/tests/e2e/migration/conftest.py
@@ -50,7 +50,9 @@ def source_client(pytestconfig) -> NominalClient:
         return NominalClient.from_profile(profile)
     auth_token = pytestconfig.getoption("source_auth_token")
     if auth_token is None:
-        raise pytest.UsageError("Either --source-profile or --source-auth-token must be provided")
+        raise pytest.UsageError(
+            "Either --source-profile or --source-auth-token must be provided for migration source environment"
+        )
     base_url = pytestconfig.getoption("source_base_url")
     print(f"Using source NominalClient.create(base_url={base_url!r})")
     return NominalClient.create(base_url=base_url, token=auth_token)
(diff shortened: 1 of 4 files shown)


tests/e2e/migration/conftest.py
```diff
@@ -0,0 +1,117 @@
...
+        raise pytest.UsageError(
+            "Either --profile or --auth-token must be provided for migration destination environment"
+        )
+    base_url = pytestconfig.getoption("base_url")
+    print(f"Using dest NominalClient.create(base_url={base_url!r})")
+    return NominalClient.create(base_url=base_url, token=auth_token)
+
+
+@pytest.fixture
+def source_archive(request) -> ArchiveFn:
```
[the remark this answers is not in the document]
seanmreidy: Intentionally keeping `source_archive` and `dest_archive` as separate fixtures. They're functionally identical to the global `archive`, but having named variants means test code is self-documenting at a glance — you can tell which environment a cleanup belongs to without tracing back to where the object came from. The cost (two trivial fixtures) is worth the readability payoff.

tests/e2e/migration/test_migration.py
```diff
@@ -0,0 +1,533 @@
...
+    dest_asset_rid_1 = runner.migration_state.get_mapped_rid(ResourceType.ASSET, source_asset.rid)
+    dest_ds_rid_1 = runner.migration_state.get_mapped_rid(ResourceType.DATASET, source_ds.rid)
+    assert dest_asset_rid_1 is not None
+    assert dest_ds_rid_1 is not None
+    dest_archive(dest_client.get_asset(dest_asset_rid_1))
+    dest_archive(dest_client.get_dataset(dest_ds_rid_1))
+
+    runner.run_migration()
+    assert runner.migration_state.get_mapped_rid(ResourceType.ASSET, source_asset.rid) == dest_asset_rid_1
+    assert runner.migration_state.get_mapped_rid(ResourceType.DATASET, source_ds.rid) == dest_ds_rid_1
```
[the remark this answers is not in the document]
seanmreidy: Adding to the bot response - I don't see this causing a false pass, only a false failure. It's difficult to assume the state of the destination tenant, so I think this is acceptable for now.

tests/e2e/migration/conftest.py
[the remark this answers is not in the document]
alkasm: personally I found this confusing as written. Maybe because of the ambiguity of archive as a verb and a noun. But no matter if it's adjective-noun or adverb-verb it doesn't make sense followed by a noun at the callsites, e.g. `source_archive(event)`. Claude had three guesses at what that piece of code does, all of which were wrong

> Without more context, my best guess is that source_archive is a function being called with event as its argument, and it probably does one of these things:
> 1. Archives the source of an event — takes an event object and stores/saves it somewhere persistent (a database, file, cold storage, etc.). The name reads as verb-object: "source-archive this event."
> 2. Looks up the archived source for an event — retrieves the original source record associated with the event (e.g., which upstream system or file the event came from).
> 3. Returns a source_archive object constructed from the event — if source_archive is a class rather than a function, this would be instantiating it, similar to how you'd write datetime(timestamp).

It's also unclear this is a cleanup fixture and doesn't run in-line--the only way you'd know that is to:
1. follow the fixture definition
2. know that the fixture takes in `request`, which is a built-in magic pytest fixture
3. understand the rules of when the finalizers are called

Compare to something more like, idk, `register_cleanup(event.archive)`. I don't even mind inlining `request.addfinalizer(event.archive)` except that `request` as a fixture itself isn't obvious so I do like the added indirection as a single place to document all of this clearly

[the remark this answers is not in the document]
alkasm: regarding the two different functions, I think the qualifier for source vs dest makes sense on the resource, not the action; the resource is the thing that exists in either environment. Compare `register_cleanup(source_event.archive)` to `register_source_cleanup(event.archive)`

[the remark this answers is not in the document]
alkasm: Here's a link to the request fixture to add into docs: https://docs.pytest.org/en/stable/reference/reference.html#request

tests/e2e/migration/test_migration.py
```diff
@@ -0,0 +1,624 @@
...
+        labels=["migration-e2e"],
+        properties={"wbt-prop": "wbt-val"},
+    )
+    workbook = template.create_workbook(asset=source_asset, title=f"migration-e2e-workbook-{uuid4()}")
+    source_archive(workbook)
+    template.archive()
+    return workbook
+
+
+def _setup_source_resources(
```
alkasm: this function isn't reused and the callsite isn't shorter than it would be if the functionality were inlined. esp since it just returns a giant tuple..I'd just inline this, less navigation required.

tests/e2e/migration/test_migration.py
```diff
@@ -0,0 +1,624 @@
...
+    source_client: NominalClient, source_archive: ArchiveFn, mp4_data: bytes
+) -> tuple[Asset, Dataset, Event, Event, Run, Checklist, DataReview, Video, Workbook]:
+    """Create a source asset with one of every resource type, each populated with labels,
+    properties, and descriptions to exercise full metadata migration fidelity.
+    """
+    start, end = _create_random_start_end()
```
alkasm: is it important that the start and end are random?

tests/e2e/migration/test_migration.py
```diff
@@ -0,0 +1,624 @@
...
+    assert set(dest.labels) == set(source.labels)
+    assert dest.properties == source.properties
+    # Linkage: dataset is accessible under the expected scope on the destination asset.
+    dest_datasets = dict(dest_asset.list_datasets())
+    assert scope_name in dest_datasets
+    assert dest_datasets[scope_name].rid == dest.rid
+    # series_tags on the scope are preserved — get_or_create_dataset raises on tag mismatch.
+    if series_tags is not None:
+        matched = dest_asset.get_or_create_dataset(scope_name, series_tags=series_tags)
+        assert matched.rid == dest.rid
```
alkasm: a function labeled "assert" should not have side-effects

alkasm: either way this branch seems like it's never hit?

alkasm (commented):
The tests themselves look great, good functionality executed.

In the future, if necessary, we could isolate to running inside a single tenant.

Otherwise mostly just some comments on naming and structure.

tests/e2e/migration/test_migration.py
seanmreidy: Nope. Removed.

seanmreidy: Nope. Removed

tests/e2e/migration/test_migration.py
seanmreidy: Good catch, removed it. And yeah, we don't have series tags but it's worth it to add in the future.

tests/e2e/migration/test_migration.py
seanmreidy: Makes sense, removed.

tests/e2e/migration/conftest.py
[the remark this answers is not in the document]
seanmreidy: Good point on the wording. I think I was in my own little world and didn't consider the outsider perspective.

alkasm (approved):

seanmreidy:
### Merge activity

* **Apr 22, 5:13 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/nominal-io/nominal-client/683).
* **Apr 22, 5:13 PM UTC**: @seanmreidy merged this pull request with [Graphite](https://app.graphite.com/github/pr/nominal-io/nominal-client/683).

PR #699, opened 2026-04-06
feat: improve export batch calculations in polars export handler

drake-nominal:
## Summary                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                             
                                                   
  This PR rewrites how the Polars exporter plans its work so it handles every channel type correctly, not just floats.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                   
   
  Previously, only `DOUBLE` channels got a data-rate estimate — INT and string/enum channels either bypassed rate budgeting or got dropped from scheduling, so the planner was under-informed for most real-world exports. Now every supported type goes through the same peak-PPS estimator, and channel grouping is a single budget-driven bin-pack rather than a per-type partition.                                                                                                                                                                                                                  
                                                                      
  Preflight is dramatically faster too. The old data-presence probe made one API call per channel; the new `filter_channels_with_data` batches up to 200 at a time and retries per-channel on batch failure. A 1000-channel export drops from roughly 1000 API round-trips to 5. The filter is exposed on `nominal.core` so callers can use it without running a full export.                                                                                                                                                                                                                            
                                                                      
  `export()` keeps the same signature and yields the same DataFrame shape — this is a scheduler rewrite, not an interface change.                                                                                                                                                                                                                                                                                                                                                                                                                                                                        
                                                                      
  ## Bugs fixed along the way                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                            
  
  - **Sub-slice jobs carried the wrong `channel_types`.** When a large channel was split across sub-slices, each sub-slice job was built with `channel_types={ch.name for ch in channel_group}` — a set comprehension that produced a `set`, not a `dict`, and reused the group-wide collection. Each sub-slice now carries `{channel.name: channel.data_type}` for its single channel.                                                                                                                                                                                                                  
  - **Bogus `assert compute_result.success is not None`.** The preceding guard only checked `error is not None`, so a conjure response with both union variants `None` crashed the export with `AssertionError`. Replaced with a log-and-skip branch.
  - **Same-named channels across datasources shared rate estimates.** Internal maps were keyed by channel name only; two channels with the same name on different datasources would silently overwrite each other in the rate dict. All internal maps are now keyed by `(data_source, name)` tuples, end to end.                                                                                                                                                                                                                                                                                         
  - **Unsupported channel types at `_batch_channel_points_per_second` now raise** instead of being silently coerced to `None`. The defensive handling was unreachable dead code — the upstream type filter guarantees only `DOUBLE`/`INT`/`STRING` reach it.                                                                                                                                                                                                                                                                                                                                             
                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                         
  ## Verification                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                        
                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                         
  - `uv run pytest tests/` — 209 passed, 8 skipped                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                       
  - `uv run mypy nominal/` — clean across 137 files
  - `uv run ruff check nominal/ tests/` — clean

### DIFF
--- a/nominal/core/channel.py
+++ b/nominal/core/channel.py
@@ -1,10 +1,12 @@
 from __future__ import annotations
 
+import concurrent.futures
 import enum
+import functools
 import logging
 from dataclasses import dataclass, field
 from datetime import datetime
-from typing import BinaryIO, Iterable, Mapping, Protocol, cast, overload
+from typing import BinaryIO, Iterable, Iterator, Mapping, Protocol, Sequence, cast, overload
 
 from nominal_api import (
     api,
@@ -18,8 +20,9 @@ from nominal_api import (
 )
 from typing_extensions import Self
 
+from nominal._utils.iterator_tools import batched
 from nominal.core._clientsbunch import HasScoutParams
-from nominal.core._utils.api_tools import RefreshableMixin, create_api_tags
+from nominal.core._utils.api_tools import RefreshableMixin, build_compute_tag_filter, create_api_tags
 from nominal.core._utils.pagination_tools import paginate_rpc
 from nominal.core.log import LogPoint, _log_filter_operator
 from nominal.core.unit import UnitLike, _build_unit_update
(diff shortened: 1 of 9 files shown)


nominal/thirdparty/polars/polars_export_handler.py
```diff
@@ -52,60 +51,102 @@
...
-                channel.name,
-                channel.data_type,
-            )
-            expressions.append((channel, None))
-        elif tags and not _has_data_with_tags(channel, tags, start_ns, end_ns):
-            logger.warning("No points found in range for channel '%s'", channel.name)
-            expressions.append((channel, None))
-        else:
-            expressions.append((channel, exprs.NumericExpr.datasource_channel(channel.data_source, channel.name, tags)))
+) -> float:
```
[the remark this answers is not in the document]
msorial44: bad bot

nominal/core/channel.py
```diff
@@ -442,3 +445,147 @@ def _create_series_from_channel(
...
+                    if len(batch) == 1:
+                        still_failing.extend(batch)
+                        continue
+                    logger.warning(
+                        "Batch data-presence check failed for %d channels; retrying individually",
+                        len(batch),
+                    )
+                    for ch in batch:
+                        pending[pool.submit(check_batch, [ch])] = [ch]
+                    continue
```
alkasm, line 567: why would it fail, and why is there a fallback to try each independently? can we get some comments to explain?

nominal/core/channel.py
```diff
@@ -442,3 +445,147 @@ def _create_series_from_channel(
...
+    if all_underconstrained:
+        sample = all_underconstrained[:10]
+        joined = ", ".join(sample)
+        suffix = f", ... ({len(all_underconstrained) - 10} more)" if len(all_underconstrained) > 10 else ""
+        logger.warning(
+            "%d channels have underconstrained tags (multiple series matched): %s%s",
+            len(all_underconstrained),
+            joined,
+            suffix,
+        )
```
alkasm, line 593: is minimizing the log line really important here?

sroustom-nominal (approved):

PR #720, opened 2026-04-16
fix: dedupe concurrent shared-resource migrations

varun-nominal:
## Summary
- add a single-flight registry to `MigrationContext` keyed by resource type, source RID, and destination workspace
- route shared-resource migrators through the base migrator's single-flight path so concurrent callers reuse one in-flight migration result
- add focused concurrency tests proving duplicate callers collapse onto one migration and that failures do not poison retries

## Why
Parallel asset migration in `internal-tools` can drive the same shared child resource through multiple concurrent migrators. The old pattern only made `MigrationState` reads/writes individually thread-safe, which still left a check-then-create race for shared datasets, videos, runs, events, checklists, workbook templates, and attachments.

This change moves the concurrency safety down into `nominal-client`, where the migration invariants actually live, so callers can stay simple while shared resources remain safe under parallel execution.

## Testing
- `uv run pytest -q tests/core/test_migrator_base.py tests/core/test_attachment_migrator.py tests/core/test_migration_destination_client_resolution.py`
- `uv run ruff check nominal/experimental/migration/migrator tests/core/test_migrator_base.py`

### DIFF
--- a/nominal/experimental/migration/migrator/base.py
+++ b/nominal/experimental/migration/migrator/base.py
@@ -55,41 +55,52 @@ class Migrator(ABC, Generic[Resource, CopyOptions]):
         return self._get_existing_destination_resource(self.destination_client_for(source), mapped_rid)
 
     def copy_from(self, source: Resource, options: CopyOptions | None = None) -> Resource:
-        resolved_options = self.default_copy_options() if options is None else options
-        if resolved_options is None:
-            raise NotImplementedError(f"{type(self).__name__} requires explicit copy options.")
-        source_rid = source.rid
-        destination_client = self.destination_client_for(source)
+        def _copy() -> Resource:
+            resolved_options = self.default_copy_options() if options is None else options
+            if resolved_options is None:
+                raise NotImplementedError(f"{type(self).__name__} requires explicit copy options.")
+            source_rid = source.rid
+            destination_client = self.destination_client_for(source)
 
-        logger = logging.getLogger(type(self).__module__)
-        log_extras = {
-            "destination_client_workspace": destination_client.get_workspace(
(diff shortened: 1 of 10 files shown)


nominal/experimental/migration/migrator/base.py
```diff
@@ -55,41 +55,52 @@ def get_existing_destination_resource(self, source: Resource) -> Resource | None
...
+
+        if self.use_singleflight():
+            return self.ctx.run_singleflight(
+                resource_type=self.resource_type,
+                source_resource=source,
+                source_rid=source.rid,
+                fn=_copy,
+            )
+
+        return _copy()
```
seanmreidy: Nit: bit of a weird pattern, since we can just put the body of _copy here.

nominal/experimental/migration/migrator/context.py
```diff
@@ -25,3 +34,46 @@ def destination_client_for(self, source_resource: Any) -> NominalClient:
...
+        """Run migration work once per resource/workspace pair.
+
+        Concurrent callers for the same source RID and destination workspace wait on
+        the first in-flight migration and reuse its result.
+        """
+        destination_client = self.destination_client_for(source_resource)
+        workspace_rid = destination_client._clients.workspace_rid
+        if workspace_rid is None:
+            raise ValueError("Destination client workspace RID is required for singleflight migrations.")
+        key: tuple[str, str, str] = (resource_type.value, source_rid, workspace_rid)
```
seanmreidy: Nit: since the RID has the resource type here, it's probably redundant to key on this, but it also can't hurt.

seanmreidy (approved):

seanmreidy (approved):

PR #722, opened 2026-04-20
feat: multi-asset/run workbook migration

seanmreidy:
### Problem                                                                                                                                                                                                                                    
   
  The existing workbook migration path works by creating a template from the source workbook, copying the template to the destination, and instantiating a new workbook from it. This round-trip through the template API works well for         
  single-asset and single-run workbooks, but cannot handle workbooks whose `dataScope` references more than one asset or run. Those workbooks were silently skipped — they would never appear on the destination tenant.
                                                                                                                                                                                                                                                 
  ### Strategy                                                                                                                                                                                                                                   
  
  #### Phase 1 — Enqueue during per-asset migration                                                                                                                                                                                              
                                                            
  When `AssetMigrator` processes a workbook associated with an asset, it now branches on the workbook's scope:

  - **Single-asset / single-run workbooks** continue to use the existing template-based copy path.                                                                                                                                               
  - **Multi-asset workbooks** are enqueued as *pending* in `MigrationState`. Before enqueuing, a scope check determines whether all referenced assets are either (a) present in the current migration config or (b) already mapped in the
  migration state from a previous run. If any asset is entirely outside both, the workbook is recorded as a skip with a structured reason rather than silently dropped.                                                                          
  - **Multi-run workbooks** are always enqueued as pending (no upfront scope check; validation deferred to migration time when the full RID mapping is available).
                                                                                                                                                                                                                                                 
  #### Phase 2 — Deferred resolution                        
                                                                                                                                                                                                                                                 
  After all assets and runs have been migrated, `WorkbookMigrator.migrate_deferred_workbooks()` iterates the pending lists and attempts each workbook:

  1. Looks up every source asset/run RID in the migration state to build a find-and-replace map.                                                                                                                                                 
  2. If any required RID is missing (e.g. one asset was out of scope), records a structured skip and moves on — no partial workbooks are created.
  3. Serialises the source workbook's layout and content objects via `ConjureEncoder`, applies the RID map using `clone_conjure_objects_with_rid_overrides`, and deserialises back. This function extends the existing UUID-regeneration logic:  
  internal panel/chart UUIDs are regenerated (with cross-object coherence preserved so layout↔content panel IDs stay in sync), while explicit asset/run RID overrides are substituted in-place without being regenerated.                        
  4. Creates the new workbook on the destination via `CreateNotebookRequest` with the updated `dataScope` and remapped content, then copies labels, properties, and preview image.
                                                                                                                                                                                                                                                 
  **Key insight for multi-run workbooks:** channel compute specs reference *both* run RIDs (`runRidVariableName`) and asset RIDs (`assetRidVariableName` / `computeSpecV2.run.assetRid`). The RID override map therefore merges both the run     
  mapping *and* all migrated asset mappings from the migration state; substituting only run RIDs leaves dangling source-tenant asset references and causes channels to show no data.                                                             
                                                                                                                                                                                                                                                 
  #### Structured skip / error log                          

  `MigrationState` gains a `skipped_resources` list of `SkippedResource(resource_type, source_rid, reason)` entries. This replaces the previous silent-drop behaviour and gives operators a clear post-run audit of what was not migrated and    
  why.
                                                                                                                                                                                                                                                 
  ### Long-term strategy                                    

  The RID find-and-replace approach is strictly more general than the template-based path. Once the multi-asset/run path has accumulated enough production confidence, the intent is to route single-asset and single-run workbooks through it as
   well, retiring the template round-trip entirely. The template path is retained unchanged in this PR.
                                                                                                                                                                                                                                                 
  ### Testing                                               

  #### Automated unit tests

  - **`test_conjure_clone_utils.py`** — tests `_generate_uuid_mapping` and `_replace_uuids_in_obj` directly:                                                                                                                                     
    - UUID-keyed values are extracted into the mapping and regenerated; plain strings are untouched; RID-prefixed UUIDs preserve their prefix.
    - Asset/run RIDs under non-UUID keys are not auto-extracted but are correctly substituted when added as explicit overrides.                                                                                                                  
    - The same UUID appearing across multiple objects receives the same replacement in both (layout↔content coherence).                                                                                                                          
    - JSON-encoded string field values are unpacked and have their contents replaced.
    - Integration tests via `clone_conjure_objects_with_rid_overrides` with patched Conjure serde: overrides applied, internal UUIDs regenerated, tuple overload preserves types and cross-object coherence.                                     
                                                                                                                                                                                                                                                 
  - **`test_multi_workbook_migration.py`** — tests `MigrationState`, `WorkbookMigrator.copy_multi_asset_workbook`, `copy_multi_run_workbook`, and `migrate_deferred_workbooks`:                                                                  
    - Pending workbook entries are idempotent on re-registration and clearable.                                                                                                                                                                  
    - Skip entries accumulate without overwriting.                                                                                                                                                                                               
    - Happy path: correct RID map passed to clone, `dataScope` populated with new RIDs, mapping recorded, pending cleared, labels/properties copied.
    - Missing asset or run: returns `None`, records a skip with the missing RID in the reason, no notebook created.                                                                                                                              
    - Idempotency: already-mapped workbook returns the existing destination workbook without creating a new one.                                                                                                                                 
    - Multi-run asset RID inclusion: migrated asset RIDs are merged into the override map alongside run RIDs.
    - Deferred routing: pending asset and run workbooks are each fetched from the correct source client and dispatched to the right copy method.                                                                                                 
    - Missing source client: workbook is skipped gracefully without raising.
                                                                                                                                                                                                                                                 
  - **`test_asset_migrator.py`** (extended) — tests `_copy_asset_and_run_workbooks` routing:                                                                                                                                                     
    - Single-asset workbook → `copy_from` (template path).
    - Multi-asset workbook with all assets in current config → pending.                                                                                                                                                                          
    - Multi-asset workbook with one asset already migrated in a prior run → pending (state check, not just config check).
    - Multi-asset workbook with an asset outside both config and state → skip recorded.                                                                                                                                                          
    - Single-run workbook → `copy_from`.
    - Multi-run workbook → always enqueued; same workbook encountered via two different runs is deduplicated idempotently.                                                                                                                       
                                                                                                                                                                                                                                                 
  #### Manual validation
                                                                                                                                                                                                                                                 
  - Migrated two assets that share a multi-asset workbook from a source tenant to a destination tenant. Verified the migrated workbook loaded with data wired up correctly for both assets.                                                      
  - Migrated the same two assets where one of them also had an associated multi-run workbook. Confirmed that after the initial fix (run RIDs only), channels referencing assets showed no data — then verified the fix (merging asset RIDs into
  the override map) resolved the issue and data appeared correctly.                                                                                                                                                                              
  - Verified the partial-migration case: migrating only one of the two assets causes the shared workbook to be skipped, with a skip entry recorded in the migration state naming the missing asset.

### DIFF
--- a/nominal/experimental/migration/migration_runner.py
+++ b/nominal/experimental/migration/migration_runner.py
@@ -4,13 +4,16 @@ import json
 import logging
 import re
 from pathlib import Path
+from typing import cast
 
 from nominal.core import NominalClient
+from nominal.core._clientsbunch import ClientsBunch
 from nominal.experimental.migration.config.migration_data_config import MigrationDatasetConfig
 from nominal.experimental.migration.config.migration_resources import MigrationResources
 from nominal.experimental.migration.migration_state import MigrationState
 from nominal.experimental.migration.migrator.asset_migrator import AssetCopyOptions, AssetMigrator
 from nominal.experimental.migration.migrator.context import DestinationClientResolver, MigrationContext
+from nominal.experimental.migration.migrator.workbook_migrator import WorkbookMigrator
 from nominal.experimental.migration.migrator.workbook_template_migrator import WorkbookTemplateMigrator
 
 logger = logging.getLogger(__name__)
@@ -82,6 +85,7 @@ class MigrationRunner:
                 destination_client=self.destination_client,
                 migration_state=self.migration_state,
                 destination_client_resolver=self.destination_client_resolver,
+                source_asset_rids=frozenset(self.migration_resources.source_assets.keys()),
             )
             asset_migrator = AssetMigrator(migration_context)
             template_migrator = WorkbookTemplateMigrator(migration_context)
@@ -101,6 +105,12 @@ class MigrationRunner:
 
             for source_template in self.migration_resources.source_standalone_templates:
                 template_migrator.clone(source_template)
+
+            source_clients_by_asset_rid: dict[str, ClientsBunch] = {
+                asset_rid: cast(ClientsBunch, asset_resources.asset._clients)
+                for asset_rid, asset_resources in self.migration_resources.source_assets.items()
+            }
+            WorkbookMigrator(migration_context).migrate_deferred_workbooks(source_clients_by_asset_rid)
         finally:
             self.save_state()
         logger.info("Completed migration")
--- a/nominal/experimental/migration/migration_state.py
+++ b/nominal/experimental/migration/migration_state.py
@@ -8,12 +8,42 @@ from nominal.experimental.migration.resource_type import ResourceType
 
 
 @dataclass
+class SkippedResource:
+    resource_type: str
+    source_rid: str
+    reason: str
+
+
+@dataclass
 class MigrationState(JSONWizard):
     # resource_type -> old_rid -> new_rid
     rid_mapping: dict[str, dict[str, str]] = field(default_factory=dict)
+    # source workbook_rid -> list of source asset_rids (for deferred multi-asset migration)
+    pending_multi_asset_workbooks: dict[str, list[str]] = field(default_factory=dict)
+    # source workbook_rid -> list of source run_rids (for deferred multi-run migration)
+    pending_multi_run_workbooks: dict[str, list[str]] = field(default_factory=dict)
+    # log of resources skipped due to missing dependencies or out-of-scope references
+    skipped_resources: list[SkippedResource] = field(default_factory=list)
 
     def record_mapping(self, resource_type: ResourceType, old_rid: str, new_rid: str) -> None:
         self.rid_mapping.setdefault(resource_type.value, {})[old_rid] = new_rid
 
     def get_mapped_rid(self, resource_type: ResourceType, old_rid: str) -> str | None:
         return self.rid_mapping.get(resource_type.value, {}).get(old_rid)
+
+    def record_pending_multi_asset_workbook(self, workbook_rid: str, asset_rids: list[str]) -> None:
+        """Record a multi-asset workbook for deferred migration. Overwrites any prior entry for idempotency."""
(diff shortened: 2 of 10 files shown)


seanmreidy:
* **#722** <a href="https://app.graphite.com/github/pr/nominal-io/nominal-client/722?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/nominal-io/nominal-client/722?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

varun-nominal:
if a queued workbook can’t be re-fetched, it never leaves pending_multi_*_workbooks, and because state is always saved at the end, later runs just spin on the same entry forever - not sure how likely this is tho

seanmreidy:
> if a queued workbook can’t be re-fetched, it never leaves pending_multi\_*_workbooks, and because state is always saved at the end, later runs just spin on the same entry forever - not sure how likely this is tho

Not sure I'm seeing where we enter an infinite loop - seems like the only failure mode is if we had a pending multi-asset workbook in the migration state and _then_ we skip it because not all source assets are present.  Unlikely but I think we can get around that by removing the workbook from pending if we end up skipping.  I've made that change. Let me know if there's something else I'm missing though.

varun-nominal (approved):

seanmreidy:
### Merge activity

* **Apr 22, 3:29 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/nominal-io/nominal-client/722).
* **Apr 22, 3:29 PM UTC**: @seanmreidy merged this pull request with [Graphite](https://app.graphite.com/github/pr/nominal-io/nominal-client/722).

PR #724, opened 2026-04-21
feat: add vanilla avro writer using fastavro

drake-nominal:
It's not the fastest tool in the shed, but it exists, and doesn't require pre-compiled binaries. Adds fastavro / cramjam as dependencies.


drake-nominal:
Porting to nominal-streaming.

PR #726, opened 2026-04-22
feat: add container image tarball upload to self-hosted registry

lg08:
## Summary

- Adds `nom container-image upload -n <name> -t <tag> -f <tarball>` CLI command and a corresponding `NominalClient.upload_container_image_from_io` method.

### DIFF
--- a/nominal/cli/container_registry.py
+++ b/nominal/cli/container_registry.py
@@ -26,7 +26,10 @@ def container_registry_cmd() -> None:
 @client_options
 @global_options
 def upload(name: str, tag: str, file: Path, client: NominalClient) -> None:
-    """Upload a container image tarball to Nominal's registry."""
+    """Upload a container image tarball to Nominal's registry.
+
+    Prints the resulting container image RID on stdout for shell-script capture.
+    """
     with open(file, "rb") as f:
         image = client.upload_container_image_from_io(f, name, tag)
-    click.echo(image)
+    click.echo(image.rid)
--- a/nominal/cli/container_registry.py
+++ b/nominal/cli/container_registry.py
@@ -19,7 +19,6 @@ def container_registry_cmd() -> None:
 @click.option(
     "-f",
     "--file",
-    "file",
     required=True,
     type=click.Path(exists=True, file_okay=True, dir_okay=False, readable=True, path_type=Path),
     help="path to the container image tarball to upload",
--- a/nominal/cli/container_registry.py
+++ b/nominal/cli/container_registry.py
@@ -1,0 +1,29 @@
+from __future__ import annotations
+
+from pathlib import Path
+
+import click
+
+from nominal.cli.util.global_decorators import client_options, global_options
+from nominal.core.client import NominalClient
+from nominal.core.filetype import FileType
+
+
+@click.group(name="container-registry")
+def container_registry_cmd() -> None:
+    pass
+
+
+@container_registry_cmd.command("upload")
+@click.option("-n", "--name", required=True, help="image name (e.g. 'my-extractor')")
+@click.option("-t", "--tag", required=True, help="image tag (e.g. 'v1.2.3')")
+@click.option("-f", "--file", required=True, help="path to the container image tarball to upload")
+@client_options
+@global_options
+def upload(name: str, tag: str, file: str, client: NominalClient) -> None:
+    """Upload a container image tarball to Nominal's registry."""
+    path = Path(file)
+    file_type = FileType.from_path(path)
+    with open(path, "rb") as f:
+        image = client.upload_container_image_from_io(f, name, tag, file_type=file_type)
+    click.echo(image)
--- a/nominal/cli/container_registry.py
+++ b/nominal/cli/container_registry.py
@@ -6,7 +6,6 @@ import click
 
 from nominal.cli.util.global_decorators import client_options, global_options
 from nominal.core.client import NominalClient
-from nominal.core.filetype import FileType
 
 
 @click.group(name="container-registry")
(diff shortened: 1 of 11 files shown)


nominal/cli/container_registry.py
```diff
@@ -0,0 +1,29 @@
...
+    pass
+
+
+@container_registry_cmd.command("upload")
+@click.option("-n", "--name", required=True, help="image name (e.g. 'my-extractor')")
+@click.option("-t", "--tag", required=True, help="image tag (e.g. 'v1.2.3')")
+@click.option("-f", "--file", required=True, help="path to the container image tarball to upload")
+@client_options
+@global_options
+def upload(name: str, tag: str, file: str, client: NominalClient) -> None:
```
drake-nominal: prefer to use `click.Path` as the type for file stuff, see examples in repo

nominal/core/client.py
```diff
@@ -985,6 +986,42 @@ def get_attachment(self, rid: str) -> Attachment:
         response = self._clients.attachment.get(self._clients.auth_header, rid)
         return Attachment._from_conjure(self._clients, response)
 
+    def upload_container_image_from_io(
+        self,
+        tarball: BinaryIO,
```
drake-nominal: `path` or `image` would fit conventions of other `x_from_io()` methods across the codebase more than `tarball`

nominal/core/client.py
```diff
@@ -985,6 +986,42 @@ def get_attachment(self, rid: str) -> Attachment:
...
+        name: str,
+        tag: str,
+        *,
+        file_type: tuple[str, str] | FileType = FileTypes.TAR,
```
drake-nominal: do we support .tar.gz, or are there ever times where this _wouldn't_ be .tar?

nominal/core/container_image.py
```diff
@@ -0,0 +1,57 @@
...
+from typing import Any, Protocol
+
+from typing_extensions import Self
+
+from nominal.core._clientsbunch import HasScoutParams, RegistryService
+from nominal.core._utils.api_tools import HasRid
+from nominal.ts import IntegralNanosecondsUTC, _SecondsNanos
+
+
+class ContainerImageStatus(str, Enum):
```
drake-nominal: just inherit `Enum`, not str (the latter makes typing go weirdly / considered problematic)

nominal/core/container_image.py
```diff
@@ -0,0 +1,57 @@
...
+    UNSPECIFIED = "CONTAINER_IMAGE_STATUS_UNSPECIFIED"
+    PENDING = "CONTAINER_IMAGE_STATUS_PENDING"
+    READY = "CONTAINER_IMAGE_STATUS_READY"
+    FAILED = "CONTAINER_IMAGE_STATUS_FAILED"
+
+    @classmethod
+    def _parse(cls, raw: object) -> ContainerImageStatus:
```
drake-nominal: we typically expose a `_from_conjure(cls, api_status: <explicit conjure type>)`, but never `object`

I know this is grpc so it's a bit different, but ideally we could keep similar conventions?

nominal/core/container_image.py
```diff
@@ -0,0 +1,57 @@
...
+@dataclass(frozen=True)
+class ContainerImage(HasRid):
+    rid: str
+    name: str
+    tag: str
+    status: ContainerImageStatus
+    created_at: IntegralNanosecondsUTC
+    size_bytes: int | None
+
+    _clients: _Clients = field(repr=False)
```
drake-nominal, line 66: Ideally we get some docstrings on the class / members

nominal/core/client.py
```diff
@@ -985,6 +986,42 @@ def get_attachment(self, rid: str) -> Attachment:
...
+            "workspaceRid": self._clients.workspace_rid,
+            "name": name,
+            "tag": tag,
+            "objectPath": s3_path,
+        }
+        response = self._clients.registry.create_image(self._clients.auth_header, request)
+        image = response.get("image")
+        if not isinstance(image, dict):
+            raise NominalError(f"unexpected CreateImageResponse from registry: {response!r}")
+        return ContainerImage._from_response(self._clients, image)
```
drake-nominal: perhaps we should name this `_from_grpc` to match the `_from_conjure` convention in other classes.

nominal/core/client.py
lg08: I think we only support plain .tar right now. Obviously wouldn't be hard to add support, but it doesn't currently exist

nominal/cli/container_registry.py
```diff
@@ -0,0 +1,33 @@
...
+    pass
+
+
+@container_registry_cmd.command("upload")
+@click.option("-n", "--name", required=True, help="image name (e.g. 'my-extractor')")
+@click.option("-t", "--tag", required=True, help="image tag (e.g. 'v1.2.3')")
+@click.option(
+    "-f",
+    "--file",
+    "file",
```
alkasm:
```suggestion
```

extremely minor nit but the additional name arg isn't necessary here since it's the same as the long name

nominal/core/_clientsbunch.py
```diff
@@ -113,6 +113,26 @@ def write_prometheus_batches(self, auth_header: str, data_source_rid: str, reque
         self._request("POST", self._uri + _path, params={}, headers=_headers, data=request)
 
 
+class RegistryService(Service):
+    """HTTP client for nominal.registry.v1.RegistryService via the gRPC-gateway JSON transcoder."""
+
+    def create_image(self, auth_header: str, request: Mapping[str, Any]) -> dict[str, Any]:
```
alkasm:
```suggestion
    def create_image(self, auth_header: str, request: Mapping[str, Any]) -> Mapping[str, Any]:
```

nominal/core/_clientsbunch.py
```diff
@@ -113,6 +113,26 @@ def write_prometheus_batches(self, auth_header: str, data_source_rid: str, reque
         self._request("POST", self._uri + _path, params={}, headers=_headers, data=request)
 
 
+class RegistryService(Service):
```
alkasm, line 117: hmm this isn't a conjure service so idt we should be inheriting from `conjure_python_client.Service`

alkasm (changes requested):

nominal/core/_clientsbunch.py
drake-nominal: the benefit is that it gets all the `truststore` stuff for on-prem / self hosted-- otherwise, we have to deal with the request session overrides we made..

drake-nominal:
Closing in lieu of #842

PR #737, opened 2026-04-24
feat: unify substring_match(es) arguments

alkasm:
- deprecate `exact_match`, `exact_substring`, `name_substring`, `insensitive_match` args, all do the same thing
- use name `substring_match` instead ("exact" is awkward since its case-insensitive)
- in channel search, use name `substring_matches` (plural)
- created a protocol for "container of T, where the container is not a str" for `substring_matches` functions - a bit janky but seems to work with mypy. also have a runtime check for extra safety
- keep + document behavior that the search applies across multiple fields. the user can post-filter

Part of SDKS-42


nominal/core/asset.py
```diff
@@ -599,9 +599,15 @@ def search_data_reviews(
         "The 'include_archived' parameter for asset.search_workbooks is deprecated and will be removed in a future "
         "version of Nominal. Please use 'archive_status' instead!",
     )
+    @warn_on_deprecated_argument(
+        "exact_match",
+        "'exact_match' is deprecated and will be removed in a future version of Nominal. "
+        "Use 'substring_match' instead.",
+    )
     def search_workbooks(
```
drake-nominal, line 607: should we not have an `@override` that only allows one or the other?

alkasm: eh i just felt like using the thing that we've been successfully using here

sidenote, i feel like we could build a pretty nice lib for doing this sort of thing

alkasm:
e2e tests succeed locally:
```
=========================================== 70 passed, 2 xfailed in 182.75s (0:03:02) ===========================================
```

nominal/core/client.py
```diff
@@ -301,19 +305,39 @@ def _iter_search_users(self, query: authentication_api.SearchUsersQuery) -> Iter
...
-        query = create_search_users_query(exact_match=exact_match, search_text=search_text)
-        return list(self._iter_search_users(query))
+        effective_substring_match = substring_match if substring_match is not None else exact_match
+        query = create_search_users_query(
+            substring_match=effective_substring_match,
+            search_text=search_text,
+        )
+        return [
+            user
+            for user in self._iter_search_users(query)
```
drake-nominal, line 338: should the `_matches_name_substring` logic not go in `_iter_search_users`? It's private, sure, but the filtering seems useful there too. 

Same applies for basically all the `iter` vs `search` pairs.

nominal/core/client.py
```diff
@@ -718,14 +766,20 @@ def search_runs(
         Returns:
             All runs which match all of the provided conditions
         """
+        effective_substring_match = (
+            substring_match
+            if substring_match is not None
+            else name_substring
+            if name_substring is not None
+            else exact_match
+        )
```
drake-nominal, line 775: I bet this logic could be encapsulated as a shared helper, maybe overkill, but helps tuck away "deprecated stuff" and with proper naming could serve as a better documentation of "don't copy this pattern" to ai.

nominal/core/log.py
```diff
@@ -63,18 +63,18 @@ def _to_conjure(self) -> storage_writer_api.LogPoint:
...
         return scout_compute_api.LogFilterOperator(
             regex_filter=scout_compute_api.LogRegexFilterOperator(regex=regex_match)
         )
     else:
         return scout_compute_api.LogFilterOperator(
             exact_match_case_insensitive_filter=scout_compute_api.LogExactMatchCaseInsensitiveFilter(
-                token=insensitive_match or ""
+                token=substring_match or ""
             )
         )
```
drake-nominal, line 79: log filtration is maybe something we want to keep consistent with prior naming-- though we exposed it as `search_logs`, it's really more of a "export_logs", and `case insensitive` is the verbiage we expose it as via UI.

nominal/core/workbook.py
```diff
@@ -16,6 +16,11 @@
 
 logger = logging.getLogger(__name__)
 
+
+def _matches_title_substring(title: str, substring_match: str | None) -> bool:
```
drake-nominal, line 20: we should have a single reusable helper for string match, maybe in api_tools :)

nominal/thirdparty/pandas/_pandas.py
```diff
@@ -317,15 +325,23 @@ def datasource_to_dataframe(
...
-                exact_match=channel_exact_match or (),
+                substring_matches=effective_channel_substring_matches or (),
                 fuzzy_search_text=channel_fuzzy_search_text or "",
             )
         )
-    elif channel_exact_match is not None or channel_fuzzy_search_text is not None:
+    elif (
+        channel_substring_matches is not None
+        or channel_exact_match is not None
+        or channel_fuzzy_search_text is not None
```
drake-nominal, line 340: the `effective_substring` pattern you have elsewhere could simplify the logic in this method.

tests/core/test_datasource.py
```diff
@@ -96,3 +96,49 @@ def test_batch_add_channels_returns_missing_when_server_drops_channel(mock_datas
...
+
+
+def test_search_channels_rejects_string_substring_matches(mock_datasource: DataSource):
+    with pytest.raises(TypeError, match="substring_matches must be a sequence of strings"):
+        list(mock_datasource.search_channels(substring_matches="asdf"))  # type: ignore[arg-type]
+
+
+def test_search_channels_rejects_string_deprecated_exact_match(mock_datasource: DataSource):
+    with pytest.warns(UserWarning, match="'exact_match' is deprecated"):
+        with pytest.raises(TypeError, match="exact_match must be a sequence of strings"):
```
drake-nominal, line 108: could combine indentation and have a single with statement here.

nominal/core/client.py
alkasm: I actually did this intentionally so that the behavior was a bit more obvious/visible. Not sure if that's ideal but 🤷🏼

nominal/core/log.py
alkasm: hmm im ok not changing it for now, but "insensitive match" is a horrible name for this

nominal/thirdparty/pandas/_pandas.py
```diff
@@ -261,7 +261,7 @@ def _get_renamed_timestamp_column(channels: list[Channel]) -> str:
 
 def datasource_to_dataframe(
     datasource: DataSource,
-    channel_exact_match: Sequence[str] | None = None,
+    channel_substring_matches: Sequence[str] | None = None,
```
alkasm: should use the not-str container

nominal/core/datasource.py
```diff
@@ -210,28 +211,30 @@ def get_write_stream(
 
     def search_channels(
         self,
-        exact_match: Sequence[str] = (),
+        substring_matches: IterableNotStr[str] | None = None,
```
alkasm: ah lets do `SequenceNotStr` instead then for being 1-to-1

alkasm: where did all the deprecations go?

nominal/core/_utils/query_tools.py
```diff
@@ -235,14 +235,14 @@ def create_search_assets_query(
     search_text: str | None = None,
     labels: Sequence[str] | None = None,
     properties: Mapping[str, str] | None = None,
-    exact_substring: str | None = None,
+    substring_match: str | None = None,
```
drake-nominal, line 238: it's weird that this got renamed at this level-- the actual api arg is exact_substring / exact_match.

nominal/core/asset.py
```diff
@@ -648,9 +649,11 @@ def search_data_reviews(
             )
         )
 
+    @warn_on_deprecated_argument("exact_match", "'exact_match' is deprecated. Use 'substring_match' instead.")
     def search_workbooks(
         self,
         *,
+        substring_match: str | None = None,
         exact_match: str | None = None,
         search_text: str | None = None,
```
drake-nominal, line 658: should we `typing.overload` to prevent both from being provided?

nominal/core/client.py
```diff
@@ -356,18 +357,24 @@ def _iter_search_users(self, query: authentication_api.SearchUsersQuery) -> Iter
         for raw_user in search_users_paginated(self._clients.authentication, self._clients.auth_header, query):
             yield User._from_conjure(raw_user)
 
-    def search_users(self, exact_match: str | None = None, search_text: str | None = None) -> Sequence[User]:
+    @warn_on_deprecated_argument("exact_match", "'exact_match' is deprecated. Use 'substring_match' instead.")
+    def search_users(
```
drake-nominal, line 361: same re: overload

nominal/core/client.py
```diff
@@ -760,10 +772,13 @@ def search_runs(
             self._iter_search_runs(
                 start=start,
                 end=end,
-                name_substring=name_substring,
                 labels=labels,
                 properties=properties,
-                exact_match=exact_match,
+                substring_match=substring_match
```
drake-nominal, line 777: should this be hoisted into a helper?

drake-nominal (approved):

nominal/core/_utils/query_tools.py
alkasm: good call

nominal/core/asset.py
alkasm: ah good catch, missed this one

nominal/core/client.py
alkasm: was thinking this one is a bit more awkward with overload since we're replacing a positional arg, but on second thought i guess overload still works

nominal/core/client.py
alkasm: it would look prettier but im not sure it would realistically add any clarity...what should we name the func?

PR #744, opened 2026-05-01
feat: allow additional header passthru

alkasm:
Allow users to specify arbitrary additional (key, value) metadata to pass to all requests, statically or dynamically:

```py
client = NominalClient.from_profile(profile, extra_headers={"Some-Header", "value"})
```

```py
import random
from nominal.core import HeaderProvider

class DynamicHeaderProvider(HeaderProvider):
    def headers(self) -> dict[str, str]:
        return {"Some-Header", random.choice(["red", "green", "blue"])}

client = NominalClient.from_profile(profile, extra_headers=DynamicHeaderProvider())
```

### DIFF
--- a/nominal/core/_utils/multipart_downloader.py
+++ b/nominal/core/_utils/multipart_downloader.py
@@ -111,7 +111,14 @@ class MultipartFileDownloader:
     _closed: bool = field(default=False, repr=False)
 
     @classmethod
-    def create(cls, *, max_workers: int | None = None, timeout: float = 30.0, max_part_retries: int = 3) -> Self:
+    def create(
+        cls,
+        *,
+        max_workers: int | None = None,
+        timeout: float = 30.0,
+        max_part_retries: int = 3,
+        default_headers: Mapping[str, str] | None = None,
+    ) -> Self:
         """Factor for MultipartFileDownloader
 
         Args:
@@ -120,6 +127,7 @@ class MultipartFileDownloader:
             timeout: Maximum amount of time before considering a connection dead
             max_part_retries: Maximum amount of retries to perform per part download (IO, presigned url expiry,
                 4xx error, and source file changing mid download are all things that may cause a retry)
+            default_headers: Additional default headers to attach to every request issued by the session.
 
         Returns:
             Constructed MultipartFileDownloader prepared to begin downloading.
@@ -128,7 +136,7 @@ class MultipartFileDownloader:
             max_workers = multiprocessing.cpu_count()
             logger.info("Inferring core count as %d", max_workers)
 
-        session = create_multipart_request_session(pool_size=max_workers)
+        session = create_multipart_request_session(pool_size=max_workers, default_headers=default_headers)
         pool = ThreadPoolExecutor(max_workers=max_workers)
         return cls(max_workers, timeout, max_part_retries, _session=session, _pool=pool, _closed=False)
 
--- a/nominal/core/_utils/multipart_downloader.py
+++ b/nominal/core/_utils/multipart_downloader.py
@@ -16,7 +16,7 @@ import requests
 from typing_extensions import Self
 
 from nominal.core._utils.multipart import DEFAULT_CHUNK_SIZE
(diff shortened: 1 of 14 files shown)


nominal/core/_utils/networking.py
```diff
@@ -15,13 +17,58 @@
...
+
+def normalize_header_provider(headers: HeaderProvider | Mapping[str, str] | None) -> HeaderProvider | None:
+    if headers is None:
+        return None
+    if isinstance(headers, HeaderProvider):
+        return headers
+    return StaticHeaderProvider(headers)
+
+
+class HeaderProviderSession(requests.Session):
```
drake-nominal, line 50: Will this work with the truststore session? Or is this incompatible with corporate environments?

nominal/core/client.py
```diff
@@ -196,6 +212,7 @@ def create(
         connect_timeout: timedelta | float = DEFAULT_CONNECT_TIMEOUT,
         *,
         workspace_rid: str | None = None,
+        extra_headers: HeaderProvider | Mapping[str, str] | None = None,
```
drake-nominal, line 215: Should we make an alias for this type?

nominal/core/dataset.py
drake-nominal, line 485: Should we make a helper out of this? We use this pattern quite a few times in the client

nominal/experimental/impersonation/__init__.py
```diff
@@ -9,5 +9,13 @@ def as_user(client: NominalClient, user_rid: str) -> NominalClient:
 
     The returned client injects the on-behalf-of header for all service requests.
     """
-    clients = client._clients.with_default_request_headers({ON_BEHALF_OF_USER_RID_HEADER: user_rid})
-    return NominalClient(_clients=clients, _profile=client._profile)
+    security = client._clients._service_config.security
+    return NominalClient.from_token(
```
drake-nominal, line 13: Is there a reason we can't just directly call the init here? From_token is deprecated right?

nominal/core/_utils/multipart_downloader.py
```diff
@@ -111,7 +111,14 @@ class MultipartFileDownloader:
...
 
     @classmethod
-    def create(cls, *, max_workers: int | None = None, timeout: float = 30.0, max_part_retries: int = 3) -> Self:
+    def create(
+        cls,
+        *,
+        max_workers: int | None = None,
+        timeout: float = 30.0,
+        max_part_retries: int = 3,
+        header_provider: HeaderProvider | None = None,
```
drake-nominal, line 120: Should this take a mapping as well as part of union?

nominal/core/client.py
alkasm: lol codex initially did and then I removed it. It's yet another type to export and adds a layer of indirection as a user to understand what you can pass in..so I just decided to keep it simple. It's only used in these alternative ctors.

nominal/core/_utils/multipart_downloader.py
alkasm: This is internal so the optionality isn't required. Internally we just pass a StaticHeaderProvider

nominal/core/dataset.py
alkasm: not sure what you're referring to here

nominal/experimental/impersonation/__init__.py
alkasm: `from_token` is not deprecated--it's what all other alternative ctors delegate to.

nominal/core/_utils/networking.py
alkasm: The truststore thing is an adapter rather than a session, and it is mounted on this session, so these should work together!

nominal/core/dataset.py
alkasm: chatted synchronously - going to punt on this, but the idea was to have s3_path be an actual handle rather than a str that we have to wrap in a bunch of places. unrelated to this changeset

drake-nominal (approved):

PR #747, opened 2026-05-05
feat: get the last value from a channel within a time window

duane-eldridge:
**Summary**

Adds `Channel.get_last_value()`, a method to fetch the last `(timestamp, value)` for a numeric or string channel within a bounded time window. Motivated by the common need to read a channel's current state without pulling a full series, e.g. for status displays, alerting checks, or coupling Nominal data with external systems.

**Implementation**

- New `Channel.get_last_value(*, start, end, tags)` on `nominal/core/channel.py`. Returns a `LastValue(timestamp, value)` named tuple, or `None` if the window is empty.
- Restricted to `DOUBLE`, `INT`, and `STRING` channels; others raise `TypeError`.
- If `end` is omitted it defaults to the current UTC time; if `start` is omitted it defaults to `end - 1h`. `start > end` is rejected with `ValueError` before any API call.
- Uses `scout_compute_api.SelectValue(last_value_point=...)` against the compute service. Empty windows are detected by matching the conjure `"a union value must not be None"` `ValueError` and returning `None`.
- `LastValue` is exported from `nominal.core`.

**Testing**

- Unit tests in `tests/core/test_channel.py` cover each return-value branch, the 1-hour default window fallback, `start > end` validation, empty-window `None`, the conjure error match, unsupported channel types, missing/unhandled response variants, and the request payload sent to compute.
- End-to-end smoke test against staging exercising the DOUBLE and STRING paths, the LOG `TypeError` guard, empty-window `None` handling, and concurrent fan-out via `ThreadPoolExecutor`. INT was not covered end-to-end because seeded INT channels surfaced as DOUBLE; the int64 wire-decoding branch is covered by unit tests.

### DIFF
--- a/nominal/core/channel.py
+++ b/nominal/core/channel.py
@@ -41,10 +41,10 @@ class LastValue(NamedTuple):
     """Result of `Channel.get_last_value`.
 
     `value` is `float` for DOUBLE channels, `int` for INT channels, and `str` for STRING channels.
-    `timestamp` is timezone-aware and in UTC.
+    `timestamp` is integer nanoseconds since the unix epoch (UTC).
     """
 
-    timestamp: datetime
+    timestamp: IntegralNanosecondsUTC
     value: float | int | str
 
 
@@ -290,7 +290,7 @@ class Channel(RefreshableMixin[timeseries_channelmetadata_api.ChannelMetadata]):
         point = response.single_point
         if point is None:
             raise RuntimeError(f"Expected response type to be `single_point`, received: `{response.type}`")
-        ts = _SecondsNanos.from_api(point.timestamp).to_datetime()
+        ts = _SecondsNanos.from_api(point.timestamp).to_nanoseconds()
         v = point.value
         if v.float64_value is not None:
             return LastValue(ts, v.float64_value)
--- a/nominal/core/channel.py
+++ b/nominal/core/channel.py
@@ -280,8 +280,9 @@ class Channel(RefreshableMixin[timeseries_channelmetadata_api.ChannelMetadata]):
         try:
             response = self._clients.compute.compute(self._clients.auth_header, request)
         except ValueError as e:
-            # Empty windows return {"type":"singlePoint","singlePoint":null}; conjure rejects null
-            # unions with this exact message. Match exactly so we don't swallow unrelated ValueErrors.
+            # The generated conjure-python union constructor doesn't handle optional union members,
+            # raising this ValueError when the window is empty (fixed upstream in palantir/conjure-python#1050).
+            # Match exactly so we don't swallow unrelated ValueErrors.
             if str(e) == "a union value must not be None":
                 return None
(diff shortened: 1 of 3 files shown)


nominal/core/channel.py
```diff
@@ -229,6 +240,66 @@ def token_factory(response: scout_compute_api.ComputeNodeResponse) -> scout_comp
...
+            start=api_start,
+            end=api_end,
+            node=scout_compute_api.ComputableNode(
+                value=scout_compute_api.SelectValue(last_value_point=self._to_compute_series(tags=tags)),
+            ),
+            context=scout_compute_api.Context(frame_references={}, variables={}, function_variables={}),
+        )
+        try:
+            response = self._clients.compute.compute(self._clients.auth_header, request)
+        except ValueError as e:
```
alkasm, line 291: Can you catch a more specific conjure error here? Or is value error the only thing we get?

nominal/core/channel.py
```diff
@@ -229,6 +240,66 @@ def token_factory(response: scout_compute_api.ComputeNodeResponse) -> scout_comp
...
+            # Empty windows return {"type":"singlePoint","singlePoint":null}; conjure rejects null
+            # unions with this exact message. Match exactly so we don't swallow unrelated ValueErrors.
```
alkasm: Are you saying that conjure fails in deserializing its own message? That doesn't sound right

duane-eldridge: it's from the code gen'd union constructor in nominal_api

`  elif type_of_union == 'singlePoint':
      if single_point is None:
          raise ValueError('a union value must not be None')`
          
          
there's actually a "fix" available one minor version ahead of what we're on now, though
https://github.com/palantir/conjure-python/pull/1050

alkasm: Ah so the issue is that it fails to handle optional union members. Got it. Ugh.

nominal/core/channel.py
```diff
@@ -229,6 +240,66 @@ def token_factory(response: scout_compute_api.ComputeNodeResponse) -> scout_comp
...
+        ts = _SecondsNanos.from_api(point.timestamp).to_datetime()
+        v = point.value
+        if v.float64_value is not None:
+            return LatestValue(ts, v.float64_value)
+        if v.int64_value is not None:
+            # int64 is wire-encoded as a string to preserve precision across JSON
+            return LatestValue(ts, int(v.int64_value))
+        if v.string_value is not None:
+            return LatestValue(ts, v.string_value)
+        raise RuntimeError(f"Unexpected value variant in `single_point` response: `{type(v).__name__}`")
```
alkasm: `type(v).__name__` would just be the class name, `Value`. instead you want the union's `.type`:
```suggestion
        raise RuntimeError(f"Unexpected value variant in `single_point` response: `{v.type}`")
```

alkasm (approved):

PR #748, opened 2026-05-05
feat: added capability for downloading data automatically scoped to a run

sroustom-nominal:
Similar to `datasource_to_dataframe`, added `run_to_dataframe`. Will extract run bounds and apply them to the datasources within the run. Accepts asset, dataset, channel, label, and property filters.


nominal/core/run.py
```diff
@@ -384,6 +386,39 @@ def list_assets(self) -> Sequence["Asset"]:
         """List assets associated with this run."""
         return list(self._iter_list_assets())
 
+    def to_dataframe(
+        self,
+        *,
+        data_filters: "Dataset | Asset | Sequence[Dataset | Asset] | None" = None,
```
drake-nominal: You can just do `Dataset | Asset | Sequence[Dataset | Asset]` without the quotes fyi. 

This should probably take a `datascope_name` as a string vs. passing in assets and datasets directly.  Otherwise, the code would look like:

```
ds = asset.get_dataset("to_download")
download_run(data_filters=[asset, ds])
```
which is just extra boilerplate.

nominal/thirdparty/pandas/_pandas.py
```diff
@@ -404,3 +406,111 @@ def _export_channel_batch(channel_batch: tuple[Channel, ...]) -> pd.DataFrame:
...
+def run_to_dataframe(
+    run: Run,
+    *,
+    data_filters: Dataset | Asset | Sequence[Dataset | Asset] | None = None,
+    labels: Sequence[str] | None = None,
+    properties: Mapping[str, str] | None = None,
+    channel_exact_match: Sequence[str] | None = None,
+    channel_fuzzy_search_text: str | None = None,
+    tags: Mapping[str, str] | None = None,
+    enable_gzip: bool = True,
```
drake-nominal: we should consider if this is actually worth exposing or not vs. having an opinionated choice we always go with.

nominal/core/run.py
drake-nominal: The beauty of taking a datascope name instead of a dataset + tags combo, is that we already know the right tag filters to use when you specify the datascope name, and then you can just add additional tag filters on top.

nominal/thirdparty/pandas/_pandas.py
```diff
@@ -404,3 +406,111 @@ def _export_channel_batch(channel_batch: tuple[Channel, ...]) -> pd.DataFrame:
...
+    -------
+    ```
+    rid = "..."  # Taken from the UI or via the SDK
+    run = client.get_run(rid)
+    dfs = run_to_dataframe(run)
+    for ds_rid, df in dfs.items():
+        print(ds_rid, df.shape)
+    ```
+
+    """
```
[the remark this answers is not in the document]
drake-nominal: should also check if there's a end as well.

nominal/thirdparty/pandas/_pandas.py
```diff
@@ -404,3 +406,111 @@ def _export_channel_batch(channel_batch: tuple[Channel, ...]) -> pd.DataFrame:
...
+    if data_filters is None:
+        out_ds_rid = {ds.rid: ds for _, ds in run.list_datasets()}
+    else:
+        run_dataset_rids = {ds.rid for _, ds in run.list_datasets()}
+        filter_items: Sequence[Dataset | Asset] = (
+            [data_filters] if isinstance(data_filters, (Dataset, Asset)) else list(data_filters)
+        )
+
+        for item in filter_items:
+            if isinstance(item, Dataset):
```
drake-nominal: can use a switch statement, but I also find it slightly unnatural to pass an asset vs. just taking datascope names & erroring out if a run contains multiple assets

nominal/thirdparty/pandas/_pandas.py
```diff
@@ -404,3 +406,111 @@ def _export_channel_batch(channel_batch: tuple[Channel, ...]) -> pd.DataFrame:
...
+            start=run.start,
+            end=run.end,
+            tags=tags,
+            enable_gzip=enable_gzip,
+            num_workers=num_workers,
+            channel_batch_size=channel_batch_size,
+        )
+        for rid, dataset in out_ds_rid.items()
+        if required_labels.issubset(dataset.labels)
+        and all(dataset.properties.get(k) == v for k, v in required_properties.items())
```
drake-nominal: these are strange filters to me. I feel like we probably should stick to using just datascope names & not have this sort of clientside filtering logic with labels+properties.

nominal/thirdparty/pandas/_pandas.py
```diff
@@ -404,3 +406,111 @@ def _export_channel_batch(channel_batch: tuple[Channel, ...]) -> pd.DataFrame:
...
+            dataset,
+            channel_exact_match=channel_exact_match,
+            channel_fuzzy_search_text=channel_fuzzy_search_text,
+            start=run.start,
+            end=run.end,
+            tags=tags,
+            enable_gzip=enable_gzip,
+            num_workers=num_workers,
+            channel_batch_size=channel_batch_size,
+        )
```
drake-nominal, line 560: It's a shame that we have to serially process each datascope vs. using further parallelism. but I suppose the fanout problem could be annoying. just seems like we're leaving perf on the table if you request 1 channel from like 10 datascopes. I also find datascope name -> dataframe a more natural return type than rid -> dataframe (because now users have to do lookups to go from rid -> datascope name after calling this function)

sroustom-nominal: Added parallelism for `datasource_to_dataframe` calls

nominal/thirdparty/pandas/_pandas.py
sroustom-nominal: Decided on some opinions, open to additional discussion.

nominal/core/run.py
sroustom-nominal: Updated the code/architecture to utilize datascopes instead, open to additional feedback

nominal/thirdparty/pandas/_pandas.py
```diff
@@ -404,3 +405,92 @@ def _export_channel_batch(channel_batch: tuple[Channel, ...]) -> pd.DataFrame:
...
+        selected = {}
+        for scope in datascopes:
+            if scope not in dataset_by_datascope:
+                logger.warning(
+                    "Run %s does not have a datascope %r; available: %s",
+                    run.rid,
+                    scope,
+                    sorted(dataset_by_datascope),
+                )
+                continue
```
drake-nominal: We should probably just value error here vs. silently not returning data.

nominal/thirdparty/pandas/_pandas.py
```diff
@@ -404,3 +405,92 @@ def _export_channel_batch(channel_batch: tuple[Channel, ...]) -> pd.DataFrame:
...
+        raise RuntimeError(
+            f"Run {run.rid!r} has {len(run.assets)} assets; run_to_dataframe only supports single-asset runs."
+        )
+
+    dataset_by_datascope = {scope: ds for scope, ds in run.list_datasets()}
+
+    if datascopes is None:
+        selected = dataset_by_datascope
+    else:
+        selected = {}
```
drake-nominal: could probably make it more terse:
```
# side benefit-- protects against dupe datascope names
requested_datascopes = set(datascopes) 

selected = {
    name: dataset 
    for name, dataset in dataset_by_datascope.items() 
    if name in requested_datascopes
}

if requested_datascopes.difference(selected):
    raise ValueError("... not all scopes present ...")
```

nominal/thirdparty/pandas/_pandas.py
```diff
@@ -404,3 +405,92 @@ def _export_channel_batch(channel_batch: tuple[Channel, ...]) -> pd.DataFrame:
...
+                    scope,
+                    sorted(dataset_by_datascope),
+                )
+                continue
+            selected[scope] = dataset_by_datascope[scope]
+
+    if not selected:
+        return {}
+
+    with concurrent.futures.ThreadPoolExecutor(max_workers=len(selected)) as pool:
```
drake-nominal: len(selected) is dangerous because this could be much greater than the number of cores, and slow things down a lot. This also suffers from fan-out: if you specify 10 workers & 10 datascopes, this will make 100 workers. maybe we skip the concurrency here, or otherwise, make a wersion of `datasource_to_dataframe` that takes a ThreadPoolExecutor to schedule jobs in so that we can share one pool of workers. A bit more work, but the latter is probably better.

tests/core/test_run_to_dataframe.py
```diff
@@ -0,0 +1,155 @@
...
+def _make_dataset(clients: MagicMock, rid: str) -> Dataset:
+    return Dataset(
+        rid=rid,
+        name=rid,
+        description=None,
+        bounds=None,
+        properties={},
+        labels=[],
+        _clients=clients,
+    )
```
drake-nominal: could probably be a pytest.fixture.

tests/core/test_run_to_dataframe.py
```diff
@@ -0,0 +1,155 @@
...
+
+def test_unknown_datascope_warns_and_skips(
+    mock_run: Run, mock_dataset: Dataset, patched_export: MagicMock, caplog: pytest.LogCaptureFixture
+):
+    """An unknown datascope ref_name is skipped with a warning; valid ref_names still download."""
+    with (
+        patch.object(Run, "list_datasets", return_value=[("primary", mock_dataset)]),
+        caplog.at_level("WARNING", logger=pandas_module.__name__),
+    ):
+        result = pandas_module.run_to_dataframe(mock_run, datascopes=["primary", "nonexistent"])
```
drake-nominal: generally prefer `magicmock` to "monkeypatching", results in a bit more "real" tests.

tests/core/test_run_to_dataframe.py
```diff
@@ -0,0 +1,155 @@
...
+            channel_batch_size=15,
+        )
+
+    assert result is expected
+    mock.assert_called_once_with(
+        mock_run,
+        datascopes=["primary"],
+        channel_exact_match=["x"],
+        num_workers=2,
+        channel_batch_size=15,
```
drake-nominal: I find these "white box" tests to be fragile / not too useful vs. something that's testing "behaviors" and treat our code as much as a black box as possible. Just testing that we're forwarding args mechanically isn't so useful, but will break every time we make a substantial change. A _great_ test will keep working even if you swap out the implementation details 100%.

nominal/thirdparty/pandas/_pandas.py
sroustom-nominal: Ya, I noticed this as well and left it in there to discuss. Thanks for the suggestion, will work that in there

nominal/thirdparty/pandas/_pandas.py
sroustom-nominal: Added `ValueError`

nominal/thirdparty/pandas/_pandas.py
sroustom-nominal: Will revert the concurrency as there are planned updates to `datasource_to_dataframe`, and we can update to a shared pool of workers if we decide to do so.

nominal/thirdparty/pandas/_pandas.py
sroustom-nominal: Incorporated

tests/core/test_run_to_dataframe.py
sroustom-nominal: Agreed. Removed the white box test

PR #750, opened 2026-05-05
chore: update release-please to include uv.lock version bump

sroustom-nominal:
Updates the release-please process to include a step bumping the `uv.lock` version such that it no longer pollutes new branches.

### DIFF
--- a/.github/workflows/release-please.yml
+++ b/.github/workflows/release-please.yml
@@ -23,6 +23,8 @@ jobs:
         if: ${{ steps.release.outputs.pr }}
         uses: actions/checkout@v4
         with:
+          # Using fromJSON since github refnames don't provide the release-please
+          # branch, which is what this next commit needs to be based off of
           ref: ${{ fromJSON(steps.release.outputs.pr).headBranchName }}
           token: ${{ secrets.NOMBOT_CLASSIC_PERSONAL_ACCESS_TOKEN }}
 
--- a/.github/workflows/release-please.yml
+++ b/.github/workflows/release-please.yml
@@ -14,6 +14,33 @@ jobs:
     runs-on: depot-ubuntu-24.04-small
     steps:
       - uses: googleapis/release-please-action@v4
+        id: release
         with:
           token: ${{ secrets.NOMBOT_CLASSIC_PERSONAL_ACCESS_TOKEN }}
           release-type: python
+
+      - name: Check out release PR branch
+        if: ${{ steps.release.outputs.pr }}
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ fromJSON(steps.release.outputs.pr).headBranchName }}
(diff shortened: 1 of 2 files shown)


sroustom-nominal:
A normal version bump correctly update the `uv.lock` file (tested on a separate branch to not pollute the release tags and `main`). The following cases will also properly update the `uv.lock` file:
- Changing a core dependency
- Changing a dev dependency
- Changing an optional dependency

.github/workflows/release-please.yml
```diff
@@ -14,6 +14,35 @@ jobs:
...
+        id: release
         with:
           token: ${{ secrets.NOMBOT_CLASSIC_PERSONAL_ACCESS_TOKEN }}
           release-type: python
+
+      - name: Check out release PR branch
+        if: ${{ steps.release.outputs.pr }}
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ fromJSON(steps.release.outputs.pr).headBranchName }}
```
drake-nominal, line 28: there should be an existing github-provided variable that gives this more simply.

.github/workflows/release-please.yml
```diff
@@ -14,6 +14,35 @@ jobs:
...
+          set -euo pipefail
+          uv lock
+          if git diff --quiet uv.lock; then
+            echo "uv.lock already in sync with pyproject.toml"
+            exit 0
+          fi
+          git config user.name NomBot
+          git config user.email 117409387+nominal-bot@users.noreply.github.com
+          git add uv.lock
+          git commit -m "chore: update uv.lock"
```
drake-nominal: could also do `--author` flag for git commit to not have to do git config first

.github/workflows/release-please.yml
```diff
@@ -14,6 +14,35 @@ jobs:
...
+        uses: astral-sh/setup-uv@v6
+        with:
+          version: "0.8.x"
+
+      - name: Sync uv.lock to release PR
+        if: ${{ steps.release.outputs.pr }}
+        run: |
+          set -euo pipefail
+          uv lock
+          if git diff --quiet uv.lock; then
```
drake-nominal: A safer check might be doing `uv lock --check` to determine if the lockfile is up to date, rather than doing a `git diff` based approach. feels maybe a bit more _correct_ idk.

.github/workflows/release-please.yml
drake-nominal: e.g. maybe `${{ github.ref_name }}`, though, maybe you avoided this because it didn't work.

sroustom-nominal: According to Claude there isn't one because this is based on the release-please branch generated name, not the PR generating branch. We could hard-code the release-please prefix (since it's mostly consistent), but I'd prefer leaving it this way in case release-please ever changes anything

drake-nominal (approved):

PR #752, opened 2026-05-06
feat: warn when NominalClient is created without a workspace RID

seanmreidy:
## Summary                                                                                                                                                                                                                                     
                                                                                                                                                                                                                                                 
  - Adds a `UserWarning` in `NominalClient.from_token` when no `workspace_rid` is provided                                                                                                                                                       
  - Warning message: "NominalClient will soon require a workspace RID. Any client which doesn't
    have a workspace RID specified will fail."                                                                                                                                                                                                   
                                                                                                                                                                                                                                                 
  ## Background
                                                                                                                                                                                                                                                 
  `get_default_workspace` returns a workspace if the user belongs to exactly one workspace, or if                                                                                                                                              
  the org has a default workspace configured. This is the first step toward making        
  `workspace_rid` a required parameter.                                                                                                                                                                                                          
                                       
  ## Test plan                                                                                                                                                                                                                                   
  - [ ] Warning emitted when calling `NominalClient.from_token(token, url)` without `workspace_rid`                                                                                                                                            
  - [ ] No warning when `workspace_rid` is provided                                                                                                                                                                                              
  - [ ] `NominalClient.create(url, token)` also triggers the warning (routes through `from_token`)                                                                                                                                             
  - [ ] `NominalClient.from_profile` triggers the warning when the profile has no `workspace_rid`


nominal/core/client.py
```diff
@@ -177,6 +178,12 @@ def from_token(
                 certifi's trust store is used.
             connect_timeout: Request connection timeout.
         """
+        if workspace_rid is None:
+            warnings.warn(
+                "NominalClient will soon require a workspace RID. Any client which doesn't have a workspace RID specified will fail.",
+                UserWarning,
+                stacklevel=2,
+            )
         trust_store_path = certifi.where() if trust_store_path is None else trust_store_path
```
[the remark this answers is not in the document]
seanmreidy: Seems a bit excessive for just a warning message

nominal/core/client.py
```diff
@@ -171,6 +180,7 @@ def from_token(
         connect_timeout: timedelta | float = DEFAULT_CONNECT_TIMEOUT,
         extra_headers: HeaderProvider | Mapping[str, str] | None = None,
         _profile: str | None = None,
+        _workspace_warning_emitted: bool = False,
```
alkasm: don't all creation methods delegate down to `from_token`? we don't need to pass this parameter. just warn here, and not in the other methods.

tests/e2e/test_workspace_resolution.py
```diff
@@ -52,6 +53,7 @@ def test_unconfigured_client_uses_workspace_service_default(client: NominalClien
     assert unconfigured_client._workspace_rid_for_search(WorkspaceSearchType.DEFAULT) == expected_workspace_rid
 
 
+@pytest.mark.filterwarnings("ignore::UserWarning")
```
alkasm: can we either assert the warning (if we're explicitly testing that the warning is raised) or explicitly pass it in the tests, rather than ignoring the warning in the tests?

seanmreidy:
@claude

seanmreidy:
@claude

PR #775, opened 2026-05-08
feat: consolidate workbook migration between single and multi asset/run

seanmreidy:
## Summary                                                                                                                                                                                                                                     
                                                            
  The workbook migrator previously had two strategies for copying workbooks: a template-based path for single-asset/run workbooks, and a find/replace path for multi-asset/run workbooks. We were planning to consolidate these anyway, but a concrete failure case made it the right time to do it now.
                                                                                                                                                                                                                                                 
**The problem with the template path**                    

Using templates introduces another potential failure point, especially if RIDs are still hardcoded in the template's JSON.  Using the template path ignores any hardcoded RIDs that should be re-mapped.  We ran into this issue when hydrating demo workbooks into a another account, where we were still pointing to the old RID when populating channels.

Now that we have a path that uses the find/replace strategy (multi-asset/run workbooks), we should use that and bypass this issue.
                                                                                                                                                                                                                                                 
### The fix                                                                                                                                                                                                                                        
  
  The find/replace path (clone_conjure_objects_with_rid_overrides) does an unconditional value replacement across the entire serialized JSON — it replaces any string value matching a known RID regardless of which field it appears in. This correctly handles both the symbolic "assetRid" style (no replacement needed, travels through unchanged) and the hardcoded RID style (gets remapped to the destination environment).
                                                                                                                                                                                                                                                 
Single-asset/run workbooks now use the same find/replace approach as multi-asset/run workbooks. The copy-immediately vs defer-to-end-of-migration distinction is preserved exactly as before — the routing in asset_migrator.py is unchanged.  
  
As part of the same change, the shared logic across all three copy paths (content extraction, clone, create request, record mapping, update metadata, migrate preview image) is extracted into a _copy_workbook helper, reducing the three methods to just their distinct concerns: building rid_overrides and data_scope.
                                                                                                                                                                                                                                                 
## Test plan                                                 

  - All existing workbook unit tests pass (43/43)
  - ruff format, ruff check, and mypy clean on changed file
  - Performed a manual end-to-end test and verified that workbooks successfully migrated.

### DIFF
--- a/nominal/experimental/migration/migrator/workbook_migrator.py
+++ b/nominal/experimental/migration/migrator/workbook_migrator.py
@@ -1,17 +1,18 @@
 from __future__ import annotations
 
+import json
 import logging
 import re
-from dataclasses import dataclass
+from dataclasses import dataclass, field
 from typing import Any, Mapping, Sequence, cast
 
+from conjure_python_client._serde.decoder import ConjureDecoder
+from conjure_python_client._serde.encoder import ConjureEncoder
 from nominal_api import api as nominal_api
 from nominal_api import scout_notebook_api, scout_workbookcommon_api
 
 from nominal.core import NominalClient
 from nominal.core._clientsbunch import ClientsBunch
-from nominal.core.asset import Asset
-from nominal.core.run import Run
 from nominal.core.workbook import Workbook
 from nominal.experimental.id_utils.id_utils import UUID_RE
 from nominal.experimental.migration.migrator.attachment_migrator import AttachmentMigrator
@@ -26,8 +27,8 @@ ATTACHMENT_RID_PATTERN = re.compile(rf"ri\.attachments\.[^.]+\.attachment\.{UUID
 
 @dataclass(frozen=True)
 class WorkbookCopyOptions(ResourceCopyOptions):
-    destination_asset: Asset | None = None
-    destination_run: Run | None = None
+    source_to_destination_asset_rid_mapping: Mapping[str, str] = field(default_factory=dict)
+    source_to_destination_run_rid_mapping: Mapping[str, str] = field(default_factory=dict)
     new_labels: Sequence[str] | None = None
     new_properties: Mapping[str, str] | None = None
 
@@ -38,6 +39,7 @@ class WorkbookMigrator(Migrator[Workbook, WorkbookCopyOptions]):
         return ResourceType.WORKBOOK
 
     def clone(self, source: Workbook) -> Workbook:
+        """Not supported — workbooks must be copied with an explicit RID mapping via copy_from."""
         raise NotImplementedError("Workbook clone is unsupported; use copy_from with destination asset/run.")
 
     def default_copy_options(self) -> WorkbookCopyOptions | None:
@@ -51,24 +53,31 @@ class WorkbookMigrator(Migrator[Workbook, WorkbookCopyOptions]):
         if existing_workbook is not None:
             return existing_workbook
 
-        if (options.destination_asset is None) == (options.destination_run is None):
-            raise ValueError("Exactly one of destination_asset or destination_run must be provided.")
-
         source_clients = cast(ClientsBunch, source._clients)
         raw_notebook = source_clients.notebook.get(source_clients.auth_header, source.rid)
 
-        asset_rid_map = dict(self.ctx.migration_state.rid_mapping.get(ResourceType.ASSET.value, {}))
+        source_run_rids = raw_notebook.metadata.data_scope.run_rids or []
+        source_asset_rids = raw_notebook.metadata.data_scope.asset_rids or []
+        rid_overrides: dict[str, str] = {
+            **options.source_to_destination_asset_rid_mapping,
+            **options.source_to_destination_run_rid_mapping,
+        }
 
-        if options.destination_run is not None:
-            dest_run_rid = options.destination_run.rid
-            source_run_rids = raw_notebook.metadata.data_scope.run_rids or []
-            rid_overrides = {**asset_rid_map, **{r: dest_run_rid for r in source_run_rids}}
-            data_scope = scout_notebook_api.NotebookDataScope(run_rids=[dest_run_rid], asset_rids=None)
+        if source_run_rids:
+            missing = [r for r in source_run_rids if r not in options.source_to_destination_run_rid_mapping]
+            if missing:
+                raise ValueError(f"Run RIDs not provided for workbook {source.rid}: {missing}")
+            data_scope = scout_notebook_api.NotebookDataScope(
+                run_rids=[options.source_to_destination_run_rid_mapping[r] for r in source_run_rids], asset_rids=None
+            )
         else:
-            dest_asset_rid = options.destination_asset.rid  # type: ignore[union-attr]
-            source_asset_rids = raw_notebook.metadata.data_scope.asset_rids or []
-            rid_overrides = {**asset_rid_map, **{r: dest_asset_rid for r in source_asset_rids}}
-            data_scope = scout_notebook_api.NotebookDataScope(run_rids=None, asset_rids=[dest_asset_rid])
+            missing = [r for r in source_asset_rids if r not in options.source_to_destination_asset_rid_mapping]
+            if missing:
+                raise ValueError(f"Asset RIDs not provided for workbook {source.rid}: {missing}")
+            data_scope = scout_notebook_api.NotebookDataScope(
+                run_rids=None,
+                asset_rids=[options.source_to_destination_asset_rid_mapping[r] for r in source_asset_rids],
+            )
 
         return self._copy_workbook(
             source,
@@ -88,9 +97,12 @@ class WorkbookMigrator(Migrator[Workbook, WorkbookCopyOptions]):
         labels: Sequence[str] | None = None,
         properties: Mapping[str, str] | None = None,
     ) -> Workbook:
+        """Create a new workbook in the destination by find/replacing RIDs in the source content.
+
+        Handles RID substitution, content attachment migration, preview image migration,
+        and metadata (labels/properties) propagation.
+        """
         content_v2 = raw_notebook.content_v2
-        if content_v2 is not None and not isinstance(content_v2, scout_workbookcommon_api.UnifiedWorkbookContent):
-            raise ValueError(f"Unexpected content_v2 type for workbook {source.rid}")
         content = (content_v2.workbook if content_v2 is not None else None) or raw_notebook.content
         if content is None:
             raise ValueError(f"Missing content for workbook {source.rid}")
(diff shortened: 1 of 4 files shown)


seanmreidy:
* **#775** <a href="https://app.graphite.com/github/pr/nominal-io/nominal-client/775?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/nominal-io/nominal-client/775?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

nominal/experimental/migration/migrator/workbook_migrator.py
```diff
@@ -52,74 +47,90 @@ def _get_existing_destination_resource(self, destination_client: NominalClient,
...
-        """This method copies content from an old workbook to a new workbook by use of templates, in order to
-        modify hardcoded variables in workbook content. We do this by creating a template in the source
-        client, copying the template to the destination client, creating a new workbook from the template in the
-        destination client, and then archiving the template in both clients.
-        """
         existing_workbook = self.get_existing_destination_resource(source)
         if existing_workbook is not None:
             return existing_workbook
 
         if (options.destination_asset is None) == (options.destination_run is None):
```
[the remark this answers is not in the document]
seanmreidy: Lol come on man we didn't even touch this part of the code

nominal/experimental/migration/migrator/workbook_migrator.py
```diff
@@ -52,74 +47,90 @@ def _get_existing_destination_resource(self, destination_client: NominalClient,
         return destination_client.get_workbook(mapped_rid)
 
     def _copy_from_impl(self, source: Workbook, options: WorkbookCopyOptions) -> Workbook:
-        """This method copies content from an old workbook to a new workbook by use of templates, in order to
-        modify hardcoded variables in workbook content. We do this by creating a template in the source
-        client, copying the template to the destination client, creating a new workbook from the template in the
-        destination client, and then archiving the template in both clients.
-        """
         existing_workbook = self.get_existing_destination_resource(source)
```
drake-nominal, line 60: probably want to keep a docstring :)

nominal/experimental/migration/migrator/workbook_migrator.py
```diff
@@ -52,74 +47,90 @@ def _get_existing_destination_resource(self, destination_client: NominalClient,
...
-
-        source_metadata = source._get_latest_api().metadata
-        labels = options.new_labels if options.new_labels is not None else source_metadata.labels
-        properties = options.new_properties if options.new_properties is not None else source_metadata.properties
-        new_workbook.update(labels=labels, properties=properties)
+        source_clients = cast(ClientsBunch, source._clients)
+        raw_notebook = source_clients.notebook.get(source_clients.auth_header, source.rid)
 
-        self._migrate_preview_image(source, new_workbook)
+        asset_rid_map = dict(self.ctx.migration_state.rid_mapping.get(ResourceType.ASSET.value, {}))
```
drake-nominal: minor nit: Having a double layer of dicts is kinda gross vs. dedicated dicts per type. You can also just _use_ the enum as a dict key.

nominal/experimental/migration/migrator/workbook_migrator.py
```diff
@@ -52,74 +47,90 @@ def _get_existing_destination_resource(self, destination_client: NominalClient,
...
 
-        new_template.archive()
-        source_template.archive()
-        return new_workbook
+        if options.destination_run is not None:
```
drake-nominal: what's the point of this method if there _isn't_ a destination run? I would assume we need to handle multi-run too based on pr description?

nominal/experimental/migration/migrator/workbook_migrator.py
```diff
@@ -52,74 +47,90 @@ def _get_existing_destination_resource(self, destination_client: NominalClient,
...
+            dest_run_rid = options.destination_run.rid
+            source_run_rids = raw_notebook.metadata.data_scope.run_rids or []
+            rid_overrides = {**asset_rid_map, **{r: dest_run_rid for r in source_run_rids}}
+            data_scope = scout_notebook_api.NotebookDataScope(run_rids=[dest_run_rid], asset_rids=None)
+        else:
+            dest_asset_rid = options.destination_asset.rid  # type: ignore[union-attr]
+            source_asset_rids = raw_notebook.metadata.data_scope.asset_rids or []
+            rid_overrides = {**asset_rid_map, **{r: dest_asset_rid for r in source_asset_rids}}
```
drake-nominal: this logic can silently go from 5 assets in a workbook to collapsing down to 1, which seems dangerous.

nominal/experimental/migration/migrator/workbook_migrator.py
```diff
@@ -52,74 +47,90 @@ def _get_existing_destination_resource(self, destination_client: NominalClient,
...
+        self.ctx.migration_state.record_mapping(ResourceType.WORKBOOK, source.rid, new_workbook.rid)
+
+        source_metadata = raw_notebook.metadata
+        new_workbook.update(
+            labels=labels if labels is not None else source_metadata.labels,
+            properties=properties if properties is not None else source_metadata.properties,
+        )
+
+        self._migrate_preview_image(source, new_workbook)
+        return new_workbook
```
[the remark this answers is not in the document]
seanmreidy: I think the issue here was that multi-asset/run workbook migration didn't handle attachments in the first place. That was an existing bug that increased in blast radius with this change. Fixing.

nominal/experimental/migration/migrator/workbook_migrator.py
```diff
@@ -52,74 +47,90 @@ def _get_existing_destination_resource(self, destination_client: NominalClient,
...
-                asset=options.destination_asset,
-                title=source.title,
-                is_draft=source.is_draft(),
-            )
-
-        destination_run = options.destination_run
-        if destination_run is None:
-            raise ValueError("Exactly one of destination_asset or destination_run must be provided.")
+        content_v2 = raw_notebook.content_v2
+        if content_v2 is not None and not isinstance(content_v2, scout_workbookcommon_api.UnifiedWorkbookContent):
```
drake-nominal: This seems like LLM code that doesn't properly know about the conjure types. `content_v2` is either None or UnifiedWorkbookContent, the isinstance duckcheck is probably unecessary.

nominal/experimental/migration/migrator/workbook_migrator.py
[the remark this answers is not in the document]
drake-nominal: i like the idea of the bot, but I'd love us to just explicitly handle attachments as its own step vs. being a wishy-washy "if it's attached to the workbook, migrate it". ignore if this is the case and teh claude comment is bad.

nominal/experimental/migration/migrator/workbook_migrator.py
```diff
@@ -186,44 +197,12 @@ def copy_multi_asset_workbook(self, source: Workbook, source_asset_rids: list[st
...
-        content_v2 = raw_notebook.content_v2
-        if content_v2 is not None and not isinstance(content_v2, scout_workbookcommon_api.UnifiedWorkbookContent):
-            raise ValueError(f"Unexpected content_v2 type for workbook {source.rid}")
-        content = (content_v2.workbook if content_v2 is not None else None) or raw_notebook.content
-        if content is None:
-            raise ValueError(f"Missing content for workbook {source.rid}")
-
-        new_layout, new_content = clone_conjure_objects_with_rid_overrides(
-            (raw_notebook.layout, content), rid_overrides=rid_map
+        data_scope = scout_notebook_api.NotebookDataScope(
```
drake-nominal: I see this logic a few times, could we make a helper for building notebook data scope? maybe unecessary / extra bloat.

nominal/experimental/migration/migrator/workbook_migrator.py
seanmreidy: Could be a destination_asset. I think that can be made clearer.

nominal/experimental/migration/migrator/workbook_migrator.py
seanmreidy: Yeah, it's a good point. I think we were saved here as long as we call from the asset_migrator, which will only call `_copy_from_impl` if we have a single asset/run workbook. However, I don't like the fact that we rely on that, since calling workbook_migrator with a multi-asset workbook will break.  This actually prompted me to do a refactor of the code to simplify logic even further to have `copy_from_impl` work with both single and multi-asset workbooks and use the same logic, putting the burden of creating the mapping on the caller.

nominal/experimental/migration/migrator/workbook_migrator.py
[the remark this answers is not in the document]
seanmreidy: Now what we've done is call migrate_attachments as part of the workbook.  I think it makes sense to have that hook at the direct parent level so we can make sure we link it to the right destination live. That also works better for partial migrations where you may want to selectively migrate some resources, which would in turn restrict the attachments.

nominal/experimental/migration/migrator/workbook_migrator.py
seanmreidy: This would require a change to migration state, seems a bit out of scope for this.

nominal/experimental/migration/migrator/workbook_migrator.py
seanmreidy: Updated.

nominal/experimental/migration/migrator/workbook_migrator.py
seanmreidy: This is all consolidated now.

nominal/experimental/migration/migrator/workbook_migrator.py
```diff
@@ -52,74 +49,133 @@ def _get_existing_destination_resource(self, destination_client: NominalClient,
...
+        rid_map: dict[str, str] = {}
+        for old_rid in attachment_rids:
+            new_attachment = attachment_migrator.migrate_by_rid(source_clients, old_rid)
+            rid_map[old_rid] = new_attachment.rid
+            logger.debug("Migrated content attachment %s -> %s", old_rid, new_attachment.rid)
+
+        def _replace_rid(match: re.Match[str]) -> str:
+            return rid_map.get(match.group(0), match.group(0))
+
+        content_json = ATTACHMENT_RID_PATTERN.sub(_replace_rid, content_json)
```
drake-nominal, line 168: this might be cleaner as a lambda vs. a nested inner function.

drake-nominal (approved):

nominal/experimental/migration/migrator/workbook_migrator.py
seanmreidy: Will tackle in a follow-up!

seanmreidy:
### Merge activity

* **May 11, 3:01 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/nominal-io/nominal-client/775).
* **May 11, 3:02 PM UTC**: @seanmreidy merged this pull request with [Graphite](https://app.graphite.com/github/pr/nominal-io/nominal-client/775).

PR #784, opened 2026-05-12
chore: revamp AI review harness

drake-nominal:
Be far more explicit with the way we setup claude for reviews in the repository.


.github/workflows/claude-review.yml
```diff
@@ -0,0 +1,108 @@
...
+            - This repository uses strict mypy for enforcing type safety. We have existing automations that will fail
+              a review if it detects any issues with type safety: instead of merely pointing out issues that are already
+              obvious, consider carefully if there is insight you can deliver about bigger picture updates to the issue
+              in particular to provide a nuanced alternative opinion. If not, don't bother commenting, but if so, give a terse yet
+              informative comment with a suggestion, but word it as such.
+            - Don't duplicate observations already raised by a prior reviewer (bot or human). 
+              If either already made a point you'd make, thumbs-up their comment instead of restating.
+            - Be terse. Flag real issues only — skip style nits that ruff already enforces, and skip filler praise.
+            - Nudge for tests when new logic isn't a thin wrapper around an existing endpoint.
+            - Never comment inline with the intent of praising or stating that something is good: this is wasteful filler.
```
seanmreidy: Another way to write this (maybe less prescriptive and more future-proof) would be - only write a comment if you expect an action to be taken. This can include if we think there's a break in intended behavior (outside the description of the PR), a bug/regression, a violation of best practices, etc.

.github/workflows/claude-review.yml
```diff
@@ -0,0 +1,108 @@
+name: Claude Code Review
+
+# Only run on PRs that are made against the main branch
+on:
+  pull_request:
+    types: [opened, ready_for_review]
```
seanmreidy, line 12: I actually like "ready_for_review" to exclusively gate the run of this workflow.   Sometimes a PR is drafted but not ready for comments yet.

drake-nominal: I assumed `ready_for_review` only triggers on transition from draft -> full PR, and not when you open a pr directly in a "ready for review" state.

Leundai (approved):

.github/workflows/claude-review.yml
alkasm: oh, are these ANDed then? I assumed OR, maybe Sean did as well

seanmreidy: I also believed it was ORd.

drake-nominal: they're OR'd

drake-nominal:
```
The docs pages are large and the fetch summarizer isn't extracting the one-line action definitions, but I can answer this definitively from well-established GitHub Actions behavior (the docs define these activity types as):

  - opened — "A pull request was created." Fires on creation regardless of draft status.
  - ready_for_review — "A draft pull request was marked as ready for review." Fires only on the draft → ready transition.

  Direct answer

  ready_for_review only fires on the transition from draft to ready. It does not fire when a PR is created directly in a ready-for-review (non-draft) state. A PR opened directly as non-draft emits only the opened action — never ready_for_review.

  Why this matters for your change

  If you drop opened and keep only ready_for_review, you'll lose review coverage for the most common case: PRs opened directly as non-draft. Those would emit only opened, which you'd no longer be listening for, so Claude would never run on them. The only PRs that would get
  reviewed are ones that start as draft and are later flipped to ready.

  Your current config already handles all three cases correctly:

  ┌─────────────────────────────────┬──────────────────┬──────────────────────────────────┐
  │            Scenario             │  Action emitted  │         Current behavior         │
  ├─────────────────────────────────┼──────────────────┼──────────────────────────────────┤
  │ PR opened directly as non-draft │ opened           │ ✅ runs (passes draft == false)  │
  ├─────────────────────────────────┼──────────────────┼──────────────────────────────────┤
  │ PR opened as draft              │ opened           │ ⏭️  skipped by if: draft == false │
  ├─────────────────────────────────┼──────────────────┼──────────────────────────────────┤
  │ Draft → marked ready            │ ready_for_review │ ✅ runs                          │
  └─────────────────────────────────┴──────────────────┴──────────────────────────────────┘

  The draft == false guard on the job is what prevents opened from running on drafts, so opened + ready_for_review together give you exactly "review every non-draft PR, once it's non-draft" without double-reviewing.
```

PR #786, opened 2026-05-14
feat: transport provider for smartcard auth

adithya-nominal:
Introduces a TransportProvider abstraction that threads an optional custom transport configuration from NominalClient construction through conjure service clients, multipart upload/download sessions, and gRPC channels. The concrete base class defines three per-transport methods (create_http_adapter(), create_multipart_adapter(), and create_grpc_channel_credentials()) each with a default backed by a ThreadSafeSSLContext (HTTP/multipart) or grpc.ssl_channel_credentials() (gRPC), so subclasses override only the transports they need to customize.

This scaffolding means the smartcard module in #788 just needs to subclass TransportProvider and override create_http_adapter() and create_grpc_channel_credentials(). Multipart uploads to S3 inherit the default since presigned URLs authenticate via query parameters and don't need a client certificate. Nominal Client users pass an instance into NominalClient.from_profile() or NominalClient.from_token() to get mTLS with a hardware-backed key across all API calls, file transfers, and gRPC channels.


nominal/smartcard/_session.py
```diff
@@ -0,0 +1,86 @@
...
+        backend = self._backend_factory(module_path)
+        certificate = select_piv_authentication_certificate(
+            backend.list_certificate_candidates(),
+            self._config,
+        )
+        pin = self._pin_provider(self._config.pin_prompt)
+        try:
+            backend.login(certificate, pin)
+        finally:
+            del pin
```
alkasm: don't rely on `del` here - simpler to just not assign the variable at all

adithya-nominal: Brought the new smartcard provider functionality to #788, to keep this one focused on just enabling new ssl providers. Will address this there.

nominal/core/_utils/networking.py
```diff
@@ -31,6 +31,17 @@ class HeaderProvider(ABC):
     def headers(self) -> Mapping[str, str]: ...
 
 
+class SslContextProvider(ABC):
+    """Provides an ssl.SSLContext for transport-level mTLS auth.
+
+    Not tied to requests or gRPC. When the client migrates to gRPC, a ``create_grpc_credentials()``
+    method can be added here without changing existing implementations.
```
alkasm: hmm as in, we'll be able to provide a default implementation based on `create_ssl_context()`? or as in, adding an abstractmethod? if the latter, then adding the abstractmethod would be a breaking change, and wouldn't automatically just work when we add grpc.

adithya-nominal: Dropped some details [here](https://www.notion.so/nmnl/Nominal-Test-Env-with-CAC-Auth-3559462a2d2e80079909dfaebfadfef7?pvs=97#3619462a2d2e80938db9de5dcfcbca69). There would be a default implementation and we'd override it for smartcards. I'll get a more concrete outline of the smartcard override in next week before merging the PRs in but I don't expect the shape of the approach to change.

nominal/core/_utils/networking.py
```diff
@@ -206,7 +224,8 @@ def create_conjure_service_client(
         return_none_for_unknown_union_types: If true, returns None instead of raising an exception when an unknown
             union type is encountered during decoding API responses.
         header_provider: Additional default headers to attach to each request.
-        enable_keep_alive: If true, enable keep alive in connections with the service.
```
alkasm, line 209: 😅

alkasm (dismissed):

nominal/core/_utils/networking.py
adithya-nominal: Added abstract def

adithya-nominal:
@cursor review

nominal/core/_clientsbunch.py
```diff
@@ -257,13 +260,15 @@ def client_factory(service_class: type[TService]) -> TService:
                 user_agent=agent,
                 service_config=cfg,
                 header_provider=header_provider,
+                ssl_context_provider=ssl_context_provider,
```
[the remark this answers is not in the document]
adithya-nominal: The code to create the smartcard context provider in https://github.com/nominal-io/nominal-client/pull/788 only creates one

nominal/core/_utils/networking.py
```diff
@@ -31,6 +31,26 @@ class HeaderProvider(ABC):
...
+    @abstractmethod
+    def create_grpc_channel_credentials(
+        self,
+        *,
+        root_certificates: bytes | None = None,
+        certificate_chain_pem: bytes | None = None,
+    ) -> Any:
+        """Return grpc.ChannelCredentials for smartcard-backed mTLS over gRPC.
+
+        The return type is ``Any`` so that grpcio is not a required import at definition time.
```
alkasm: we can add it ad a dev dependency and put it in a type checking guard for now, and still have the accurate type hint

grpc will be a hard req within a few weeks anyways

adithya-nominal: Added

PR #793, opened 2026-05-20
fix: apache 2.0 the package metadata and subproject

alkasm:
```sh
% diff packages/nominal-tdms/LICENSE LICENSE 
% diff packages/nominal-tdms/NOTICE NOTICE  
```
the licenses and notices are identical between the root package and subpackages

### DIFF
--- a/packages/nominal-tdms/NOTICE
+++ b/packages/nominal-tdms/NOTICE
@@ -1,0 +1,3 @@
+nominal-client
+This product includes software developed by Nominal, Inc.
+Copyright (c) 2026 Nominal, Inc.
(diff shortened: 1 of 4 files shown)


packages/nominal-tdms/NOTICE
```diff
@@ -0,0 +1,3 @@
+nominal-client
```
alkasm, line 1:
```suggestion
nominal-tdms
```

probably?

alkasm: well, `nominal-client` is the repo name, not the pypi name..

alkasm: so i guess it should stay?

victoryNap (approved):

PR #794, opened 2026-05-22
feat: grpc support for smartcard auth

adithya-nominal:
Implements create_grpc_channel_credentials() on SmartcardTransportProvider, completing the gRPC portion of smartcard mTLS from #788.

The core addition is SmartcardPrivateKeySigner, a PKCS#11 signing callback compatible with gRPC's ssl_channel_credentials_with_custom_signer. create_grpc_channel_credentials() reuses the same SmartcardSessionManager as the HTTP path for certificate and slot discovery, derives the PEM certificate chain from the card's DER data, and caches the resulting ChannelCredentials behind the provider lock. A close() method releases the PKCS#11 session when the channel is torn down. gRPC on Windows uses the same PKCS#11 path rather than Schannel since gRPC's BoringSSL stack bypasses the OS certificate store entirely, so the custom-signer approach is the right solution on all platforms.


adithya-nominal:
@cursor review

PR #803, opened 2026-05-30
feat: cleanup dependencies

alkasm:
- remove python>=3.10 markers for nominal-streaming and nominal-video (nominal already is >=3.10)
- set slightly stricter boundaries for wildcard version deps (e.g. pandas, polars, pyyaml)
- remove unused packages (cachetools, types-cachetools)
- remove dataclass-wizard (and reimpl functionality without it)
- remove unused dev deps
- add [mis] extra for openpyxl, and add an import guard for it
- removed some comments that were duplicative of information (platforms that are supported)
- set urllib to >= rather than ==, and used the next version up (2.7.0) so that deps.dev reports things correctly


nominal/experimental/migration/migration_state.py
```diff
@@ -25,6 +25,27 @@ class MigrationState(JSONWizard):
...
+            pending_multi_run_workbooks=data.get("pending_multi_run_workbooks", {}),
+            skipped_resources=skipped_resources,
+        )
+
+    @classmethod
+    def from_json(cls, data: str) -> MigrationState:
+        return cls.from_dict(json.loads(data))
+
+    def to_json(self) -> str:
+        return json.dumps(asdict(self))
```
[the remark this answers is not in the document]
alkasm: Yes, this was intentional, but good catch. This is only an issue if someone upgrades mid-migration. Also this code is mostly internal-only usage. So I think it should be ok.

pyproject.toml
```diff
@@ -25,35 +25,26 @@ classifiers = [
...
-    # - Windows x86_64 (AMD64)
-    # - macOS arm64
-    # - Linux x86_64, arm64, and armv7l
+    "pandas>=2.0.0",
+    "polars>=1.0.0",
+    "python-dateutil>=2.8.2",
+    "pyyaml>=6.0",
+    "requests>=2.32.0",
+    # urllib3>=2.7.0 fixes CVE-2026-21441, CVE-2026-44431
+    "urllib3>=2.7.0",
```
alkasm, line 39: Bumped from 2.6.3, and made `>=` rather than `==`. I'll note the CVE says the issue was patched in 2.6.3, which is what we were previously pinned to, but deps.dev seems to think 2.6.3 is affected. In any case I don't think it is going to be an issue to just say `>=2.7.0`, so went with that.

pyproject.toml
```diff
@@ -25,35 +25,26 @@ classifiers = [
...
     "truststore>=0.10.4",
-    "types-cachetools>=6.0.0.20250525",
     "typing-extensions>=4,<5",
-    # Any version is okay for these dependencies
-    "pandas>=0.0.0",
-    "polars>=0.0.0",
-    "python-dateutil>=0.0.0",
-    "pyyaml>=0.0.0",
-    "requests>=0.0.0",
-    "openpyxl>=0.0.0",
```
alkasm, line 46: Moved to `[mis]` extra. This is a minor breaking change but I believe acceptable. Code now has an try/except around the import for guidance.

pyproject.toml
```diff
@@ -25,35 +25,26 @@ classifiers = [
...
-    "dataclass-wizard>=0.39.1",
     "exceptiongroup>=1.3.0",
-    "ffmpeg-python>=0.2.0",
     "nominal-api==0.1224.0",
-    "rich>=14.1.0",
-    "tabulate>=0.9.0,<0.10",
     "truststore>=0.10.4",
-    "types-cachetools>=6.0.0.20250525",
     "typing-extensions>=4,<5",
-    # Any version is okay for these dependencies
```
alkasm, line 40: updated to give a slightly-better than complete wildcard for these

pyproject.toml
```diff
@@ -25,35 +25,26 @@ classifiers = [
 
 requires-python = ">=3.10,<4"
 dependencies = [
-    "cachetools>=6.1.0",
-    "click>=8,<9",
     "conjure-python-client>=3.1.0,<4",
-    "dataclass-wizard>=0.39.1",
     "exceptiongroup>=1.3.0",
-    "ffmpeg-python>=0.2.0",
```
alkasm, line 33: Moved to the `[video]` extra. This is a minor breaking change but I believe acceptable. Code now has an try/except around the import for guidance.

pyproject.toml
```diff
@@ -25,35 +25,26 @@ classifiers = [
...
     "nominal-api==0.1224.0",
-    "rich>=14.1.0",
-    "tabulate>=0.9.0,<0.10",
     "truststore>=0.10.4",
-    "types-cachetools>=6.0.0.20250525",
```
alkasm, line 38: unused (also should have been a dev dep in the first place)

pyproject.toml
```diff
@@ -25,35 +25,26 @@ classifiers = [
 
 requires-python = ">=3.10,<4"
 dependencies = [
-    "cachetools>=6.1.0",
```
alkasm, line 28: unused

pyproject.toml
```diff
@@ -25,35 +25,26 @@ classifiers = [
...
-    "click>=8,<9",
     "conjure-python-client>=3.1.0,<4",
-    "dataclass-wizard>=0.39.1",
```
alkasm, line 31: functionality re-written using `dataclasses.asdict()` and `json.dumps()`

stefanv (approved):
Nice cleanup. Failures are also present on main.

PR #804, opened 2026-06-01
feat: allow for excluded workbooks in asset migration

seanmreidy:
During the sandbox hydration migration, we discovered that the migration was inadvertently copying draft workbooks from the source asset into the destination workspace. While these drafts were never being promoted to the demo workbooks list
(that logic was already correct), they were still being created in the destination, cluttering the sandbox with incomplete workbooks that customers shouldn't see.

The root cause is that `AssetMigrator._copy_asset_and_run_workbooks` calls `search_workbooks(include_drafts=True)` and blindly copies every workbook it finds on the source asset, regardless of whether it belongs to the curated demo set. The right
fix is to give callers a way to restrict which workbooks get migrated rather than cleaning up the mess post-hoc.

Rather than making nominal-client aware of the concept of "demo workbooks" (which belongs in internal-tools), we introduced a generic workbook allowlist: AssetResources now carries an optional source_workbook_rids: frozenset[str] | None
field. When None, behavior is unchanged and all workbooks are migrated. When populated, only workbooks whose RIDs appear in the set are copied — both at the asset level and for run-scoped workbooks. AssetCopyOptions gains a matching
workbook_rids_allowlist field that MigrationRunner passes through from AssetResources. This pattern is intentionally symmetric so it can be replicated for other resource types (runs, checklists) in the future by following the same shape.

On the internal-tools side, when set_to_demo_workbook is true, we now fetch the source demo workbook RIDs from the API once upfront and use them to populate source_workbook_rids on every AssetResources before migration begins. This
pre-fetched set is reused for validation (so that draft workbooks with missing metadata no longer cause spurious validation failures) and passed directly into _update_demo_workbooks to avoid a redundant API call on each destination profile.

Test coverage was added at both levels. Four unit test cases were added to `test_asset_migrator.py` covering the None (all-pass), populated allowlist, empty allowlist, and run-workbook filtering scenarios. The existing
test_migrate_asset_maximal e2e test was extended to exercise this path in a real migration: a third workbook is created on the source asset and intentionally omitted from source_workbook_rids, with a final assertion confirming it has no
mapping in the migration state after the run.

### DIFF
--- a/nominal/experimental/migration/config/migration_resources.py
+++ b/nominal/experimental/migration/config/migration_resources.py
@@ -11,6 +11,7 @@ from nominal.core.workbook_template import WorkbookTemplate
 class AssetResources:
     asset: Asset
     source_workbook_templates: Sequence[WorkbookTemplate]
+    source_workbook_rids: frozenset[str] | None = None
 
 
 @dataclass(frozen=True)
--- a/nominal/experimental/migration/migration_runner.py
+++ b/nominal/experimental/migration/migration_runner.py
@@ -108,6 +108,7 @@ class MigrationRunner:
                         include_video=self.asset_inclusion_config.include_video,
                         include_checklists=self.asset_inclusion_config.include_checklists,
                         include_workbooks=self.asset_inclusion_config.include_workbooks,
+                        workbook_rids_allowlist=asset_resources.source_workbook_rids,
                     ),
                 )
 
--- a/nominal/experimental/migration/migrator/asset_migrator.py
+++ b/nominal/experimental/migration/migrator/asset_migrator.py
@@ -9,6 +9,7 @@ from nominal_api import scout_asset_api
 from nominal.core import NominalClient
 from nominal.core._event_types import SearchEventOriginType
 from nominal.core.asset import Asset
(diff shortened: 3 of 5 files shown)


seanmreidy:
* **#804** <a href="https://app.graphite.com/github/pr/nominal-io/nominal-client/804?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/nominal-io/nominal-client/804?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

abvthecity (approved):

alkasm (approved):

seanmreidy:
### Merge activity

* **Jun 2, 1:24 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/nominal-io/nominal-client/804).
* **Jun 2, 1:24 PM UTC**: @seanmreidy merged this pull request with [Graphite](https://app.graphite.com/github/pr/nominal-io/nominal-client/804).

PR #809, opened 2026-06-09
feat(grpc): expose Role Service via a shared-channel ClientsBunch gRPC pattern

varun-nominal:
## Summary

Adds a gRPC client path to the SDK — one shared channel plus a stub factory that mirrors the existing conjure stub factory — and uses it to expose the Role Service. Replaces #835.

## Functional changes

- **`nominal/core/_utils/grpc_tools.py`** (new): builds one shared, fully-configured `grpc.Channel` and a `create_grpc_stub_factory(StubClass) -> stub` closure — the gRPC analogue of `create_conjure_client_factory`.
- **`ClientsBunch.roles`**: the generated `RoleServiceStub`, built in `from_config` and exposed directly (no wrapper class).
- **`dataset_utils`**: dataset-owner lookup now calls `clients.roles.GetResourceRoles(...)`, replacing the hand-rolled per-call channel.
- **deps**: `nominal-api-protos == 0.1282.0` is now a base dependency (the `protos` extra is dropped); write-stream proto imports moved to the `nominal.protos.*` namespace.
- **docs**: `docs/src/networking-tls.md` + the `from_token`/`from_profile`/`create` docstrings document TLS on corporate networks.

## Key conjure → gRPC porting decisions

- **One shared channel, not per-call** — a single long-lived, multiplexed channel mirrors conjure's shared `requests` session; every stub binds to it.
- **Cross-cutting concerns via interceptors** — bearer auth + `HeaderProvider` headers and a default per-call deadline are injected by client interceptors, so call sites never pass `metadata=`/`timeout=` (the analogue of `HeaderProviderSession` + conjure's per-request timeouts).
- **Retry** — native gRPC `retryPolicy`, applied uniformly to every method and derived from the same `ServiceConfiguration` conjure uses (`max_num_retries`→`maxAttempts`, `backoff_slot_size`→`initialBackoff`, 503/429→`UNAVAILABLE`/`RESOURCE_EXHAUSTED`). Two documented divergences from `RetryWithJitter`: the first retry can't be 0 (gRPC requires `initialBackoff > 0`), and gRPC only retries *uncommitted* RPCs.
- **TLS** — gRPC (BoringSSL) can't consult the OS trust store on demand like `truststore`, so the root bundle is materialized statically as `trust_store_path ∪ OS roots`, stdlib-only: Windows/Linux auto-include enterprise CAs; macOS uses `trust_store_path` (documented escape).
- **Generated stubs exposed directly** on `ClientsBunch`; service-specific logic stays in feature modules, not the transport layer.

## Testing

`uv run pytest tests/` — 233 passed / 10 skipped · `uv run mypy -p nominal` — clean · `uv run ruff check` — clean.


nominal/__init__.py
```diff
@@ -0,0 +1,87 @@
+import warnings
+from pkgutil import extend_path
+from typing import Any
+
+__path__ = extend_path(__path__, __name__)  # type: ignore[name-defined]
```
varun-nominal: because nominal-client already has its own regular Python package named nominal, while the new proto wheel installs generated code under nominal.protos.

drake-nominal (approved):

PR #813, opened 2026-06-11
feat: independent version bumping for release namespaces

seanmreidy:
## Description

This PR allows for independent build and bump of packages based on where changes were made, starting with nominal and nominal-tdms.

Files changed:

- release-please-config.json (new) — monorepo config with both packages; separate-pull-requests: false keeps a single grouped release PR so the existing uv.lock sync step still works
- .release-please-manifest.json (new) — seeds release-please with the true state: nominal at 1.145.1, nominal-tdms at 1.143.0 (its last successfully published version)
- release-please.yml — swaps release-type: python for config-file/manifest-file; everything else unchanged
- publish-pypi.yml — removes the branches: main trigger (publishing should only happen on release tags), adds nominal-tdms-v*._._ tag pattern, detects the package from the tag name and builds only that package
- build-check-test.yml — adds a build-packages job that runs uv build --package for each workspace package on every PR

## Testing

- Added a new build-check-test CI workflow which verifies build
- Added a dry run step and json validation step
- Manually checked dry run and validated that no changes were picked up for `nominal-tdms`: `✔ No commits for path: packages/nominal-tdms, skipping`
- Will check once this PR is merged that release-please doesn't trigger a new publish of the tdms namespace


seanmreidy:
* **#813** <a href="https://app.graphite.com/github/pr/nominal-io/nominal-client/813?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/nominal-io/nominal-client/813?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

.github/workflows/build-check-test.yml
```diff
@@ -47,3 +47,41 @@ jobs:
       - name: Run unit tests
         run: |
           just test
+
+  build-packages:
+    runs-on: depot-ubuntu-24.04-small
```
alkasm, line 52: should we change the others to -small as well? i dont know if it's better to keep them all aligned for caching and so on or not, but im sure whatever `small` resources are sufficient for doing the other tasks too.

if its not really an issue for caching or whatever, fine to leave as is too, just thinking out loud

.github/workflows/build-check-test.yml
```diff
@@ -47,3 +47,41 @@ jobs:
...
+  validate-release-config:
+    runs-on: depot-ubuntu-24.04-small
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - name: Validate release config JSON
+        run: |
+          python3 -m json.tool release-please-config.json
+          python3 -m json.tool .release-please-manifest.json
```
alkasm, line 74: this seems kinda unnecessary but also fine

.github/workflows/publish-pypi.yml
```diff
@@ -1,10 +1,14 @@
...
+  workflow_dispatch:
+    inputs:
+      tag:
+        description: 'Tag to simulate (e.g. nominal-tdms-v1.143.1)'
+        required: true
   push:
     tags:
       - "v*.*.*"
-    branches:
-      - main
```
alkasm, line 7: should we keep this in? seems like a mostly extraneous safeguard

.github/workflows/publish-pypi.yml
```diff
@@ -21,7 +25,17 @@ jobs:
           version: "0.8.x"
           enable-cache: false
           cache-dependency-glob: "uv.lock"
-      - name: Build and Publish Package to PYPI
+      - name: Determine package from tag
+        id: package
```
alkasm, line 29: what is `id` for?

alkasm: oh its because below it's referenced as `steps.package`

.github/workflows/publish-pypi.yml
```diff
@@ -21,7 +25,17 @@ jobs:
...
         run: |
-          uv build --all-packages
-          uv publish --trusted-publishing always
+          TAG="${{ github.event.inputs.tag || github.ref_name }}"
+          if [[ "$TAG" == v* ]]; then
+            echo "name=nominal" >> "$GITHUB_OUTPUT"
+          else
+            echo "name=${TAG%-v*}" >> "$GITHUB_OUTPUT"
+          fi
```
alkasm, line 36: maybe instead of using two separate schemes of:
- `v*` tag implies `nominal` package
- `whatever-v*` tag implies `whatever` package

we should just use the latter and explicitly use `nominal-v*`?

release-please-config.json
```diff
@@ -0,0 +1,12 @@
+{
+  "separate-pull-requests": false,
+  "packages": {
+    ".": {
+      "release-type": "python"
```
alkasm, line 5: should this also have `component`?

.github/workflows/build-check-test.yml
seanmreidy: Agreed — leaving runner sizes as-is, want to keep this scoped to the namespace changes.

.github/workflows/build-check-test.yml
seanmreidy: We do need it — release-please walks the full commit history to compute version bumps, and the default shallow clone (depth 1) cuts that off. Added a comment to the workflow to make that explicit.

.github/workflows/publish-pypi.yml
seanmreidy: Keeping it — it's the mechanism for manually testing the build logic without pushing a real tag (step 2 of our verification plan). Without it there's no way to validate the package detection code short of a real release.

.github/workflows/publish-pypi.yml
seanmreidy: Yep exactly, needed to reference it as `steps.package.outputs.name` in the build step.

.github/workflows/publish-pypi.yml
seanmreidy: I'd push back on this one. The `v*.*.*` convention for the root package is standard semver tagging and matches what release-please uses by default for a single-package repo. Switching to `nominal-v*.*.*` would mean migrating all existing tags, creating a new `nominal-v1.145.1` anchor (same situation we just fixed for tdms), and updating any external tooling watching for `v*` tags. The asymmetry is a small cost for avoiding that churn.

release-please-config.json
seanmreidy: Related to the tag naming question above — if we added `component: nominal` here, release-please would switch to `nominal-v*.*.*` tags for the root package, which has the migration cost described there. Without `component`, it keeps the existing `v*.*.*` convention. Happy to add it if we decide to go the uniform route, but I'd leave it out for now.

.github/workflows/build-check-test.yml
alkasm: My comment was referring to the separate step for json validation

.github/workflows/publish-pypi.yml
alkasm: Why would we need to migrate previous tags? either way, fine with the convention of `v* -> nominal`, not a biggie.

release-please-config.json
alkasm: yea ok figured

.github/workflows/publish-pypi.yml
alkasm: no i meant should we keep `branches: [main]` around or not

alkasm (approved):

.github/workflows/build-check-test.yml
seanmreidy: Ah. Yeah, that's more for just the initial validation, but can't hurt to keep in there

.github/workflows/publish-pypi.yml
seanmreidy: Intentionally removed — it was triggering the publish workflow on every push to main, not just releases. Since the version only bumps on release-please PRs, every other main push would try to re-upload an already-existing version and fail with a 400. Tag-only triggering is the right model here since release-please always creates a tag when cutting a release.

alkasm: oooh so its an _or_ condition not an _and_?

alkasm: copilot says otherwise:

> When both tags: and branches: are specified under the same push: trigger, they are AND — the push must match both a tag pattern and a branch pattern to fire.

(I didn't remember this being an issue previously so that's why I was taking a look)

seanmreidy: Confirmed — looking at the actions history, `branches` and `tags` under `push` are OR, not AND. Both fire independently: branch pushes match `branches`, tag pushes match `tags`, and each triggers a separate workflow run. You can see this clearly in the history — every release shows two runs, one for the main branch push and one for the tag.

That said, I overstated the case earlier. The release pushes are *also* failing — the branch push for a release uploads `nominal-X.Y.Z` successfully but then fails on `nominal_tdms-1.143.0` (already exists on PyPI), and the tag push fails on both. Non-release merges to main (e.g. `chore: bump nominal-api`) then pile on with the same tdms failure.

So there are two distinct problems: `branches: main` causes spurious publish attempts on every non-release merge, and the nominal-tdms version being stuck at 1.143.0 causes every publish to fail regardless. The monorepo release-please config fixes the versioning; removing `branches: main` eliminates the spurious triggers.

seanmreidy: Err this bot posted the comment prematurely - disregard and standby (this is human Sean)

alkasm: i think copilot lied to me 😱

(literally the _embedded into github_ copilot lied to me about github behavior)

https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#example-including-branches-and-tags

seanmreidy:
### Merge activity

* **Jun 16, 1:43 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/nominal-io/nominal-client/813).
* **Jun 16, 1:43 PM UTC**: @seanmreidy merged this pull request with [Graphite](https://app.graphite.com/github/pr/nominal-io/nominal-client/813).

PR #826, opened 2026-06-16
feat(experimental): runtime decorator for authoring containerized extractors

varun-nominal:
Adds `nominal.experimental.extractor` — an in-container runtime SDK that lets authors write a Nominal Hosted containerized extractor as a single decorated function, instead of hand-wiring env vars and hand-writing `manifest.json`.

### The decorator

```python
from nominal.experimental.extractor import extractor, ExtractorContext, IngestType

@extractor
def split(ctx: ExtractorContext) -> None:
    table = read_parquet(ctx.input())                 # the mounted input file (Path)
    parts = ctx.param("PARTS", int, default=2)         # typed, lazy, default
    for i, chunk in enumerate(chunks_of(table, parts)):
        out = ctx.output_dir / f"part_{i}.parquet"
        write_parquet(chunk, out)
        ctx.add_output(out, ingest_type=IngestType.TABULAR)

if __name__ == "__main__":
    split.run()
```

`ExtractorContext` resolves everything from the environment the ingest pipeline injects — no declarations:
- `input()` / `inputs` (discovered from the `/input` mount) and `input("ENV_VAR")` by name
- `param(name, type=..., default=..., required=...)` (read on demand, coerced)
- `output_dir` (from `$OUTPUT_DIR`)
- `add_output(path, *, ingest_type, tag_columns, channel_prefix, timestamp_metadata)` — records a file you wrote (no I/O itself)
- `build_manifest()` — the manifest document, if you want to inspect/test it directly

### What `run()` does
- **Manifest mode** (image registered MANIFEST → `MANIFEST_FILE_NAME` present): writes `manifest.json` automatically from your `add_output` calls, in scout's exact `ExtractorManifest` shape.
- **Single-file mode**: enforces exactly one output, writes no manifest.
- **Mode mismatch**: emitting multiple files for a single-file-registered image fails *in-container* with a clear message — instead of the cryptic downstream `MultipleFilesFound` (the failure mode we actually hit testing the manifest image).
- Any exception → traceback to stderr + non-zero exit, so the ingest job fails cleanly.

### Per-output timestamp metadata

A manifest output can declare its own timestamp field via `timestamp_metadata`, mirroring scout's new `ManifestOutput.timestampMetadata`. This lets outputs of different formats in one job use different timestamp fields (e.g. a CSV keyed on `time` and a JSON_L file keyed on `ts`), instead of all sharing the single job-level timestamp metadata.

```python
from nominal.experimental.extractor import TimestampMetadata, EpochTimeUnit

ctx.add_output(
    out,
    ingest_type=IngestType.JSON_L,
    timestamp_metadata=TimestampMetadata("ts", EpochTimeUnit.MICROSECONDS),
)
```

- `TimestampMetadata(series_name, epoch_time_unit)` + `EpochTimeUnit` (SECONDS/MILLISECONDS/MICROSECONDS/NANOSECONDS) mirror scout's `ManifestTimestampMetadata` / `ManifestEpochTimeUnit`.
- Emitted as `{"seriesName", "epochTimeUnit"}` only when set; when omitted, the job-level timestamp metadata is used (existing manifests are unchanged).
- **Only numeric epoch timestamps** are expressible (matching what the backend supports). For `JSON_L` outputs each line must still contain a `MESSAGE` field — that path is log ingest.
- **Cross-repo coupling:** the field is only *honored* once the scout side merges (`ManifestOutput.timestampMetadata` + the JSON_L timestamp-field fix) and combined-service + the Yeeter uploader are redeployed on that build. Until then the SDK emits a field scout ignores — forward-compatible, just not yet active.

### Notes
- **Stdlib-only.** No `nominal_api`, `protos`, or grpc imports, so it stays light inside a minimal extractor image and starts fast. Format I/O (pyarrow, etc.) is the author's own dependency.
- **Registering** the built image with Nominal is a separate, explicit step (the Nominal Hosted extractor APIs) — deliberately decoupled from the container code, since conflating them is what produces output-format mismatches.

### Testing
`tests/experimental/test_extractor.py` (12 tests): single-file passthrough, manifest generation (incl. `channelPrefix`/`tagColumns`/`timestampMetadata`, and omission when unset), the single-file>1 and manifest-zero guardrails, parameter coercion/defaults/required, input-by-env-name, output-outside-OUTPUT_DIR rejection, and non-zero exit on failure. `just check-types` clean, `ruff format`/`check` clean.

### DIFF
--- a/nominal/experimental/__init__.py
+++ b/nominal/experimental/__init__.py
@@ -1,1 +1,17 @@
-from nominal.experimental.impersonation import as_user as as_user
+from __future__ import annotations
+
+from typing import TYPE_CHECKING, Any
+
+if TYPE_CHECKING:
+    from nominal.experimental.impersonation import as_user as as_user
+
+
+# Lazy so importing a sibling package (e.g. nominal.experimental.extractor, which must stay
+# stdlib-only for use inside minimal extractor container images) doesn't eagerly pull in
+# nominal.core.client and its nominal_api/conjure_python_client dependency graph.
+def __getattr__(name: str) -> Any:
+    if name == "as_user":
+        from nominal.experimental.impersonation import as_user
+
+        return as_user
+    raise AttributeError(f"module {__name__!r} has no attribute {name!r}")
--- a/nominal/experimental/extractor/__init__.py
+++ b/nominal/experimental/extractor/__init__.py
@@ -1,21 +1,19 @@
 from nominal.experimental.extractor._extractor import (
-    EpochTimeUnit as EpochTimeUnit,
+    EpochTimeUnit,
+    Extractor,
+    ExtractorContext,
+    ExtractorError,
+    IngestType,
+    TimestampMetadata,
+    extractor,
 )
-from nominal.experimental.extractor._extractor import (
-    Extractor as Extractor,
-)
-from nominal.experimental.extractor._extractor import (
-    ExtractorContext as ExtractorContext,
-)
-from nominal.experimental.extractor._extractor import (
-    ExtractorError as ExtractorError,
-)
-from nominal.experimental.extractor._extractor import (
-    IngestType as IngestType,
-)
-from nominal.experimental.extractor._extractor import (
-    TimestampMetadata as TimestampMetadata,
-)
-from nominal.experimental.extractor._extractor import (
-    extractor as extractor,
-)
+
+__all__ = [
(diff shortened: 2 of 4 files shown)


varun-nominal:
in the future, it almost seems as if we want this decorator to **_define_** the extractor rather than just implement it somehow

PR #830, opened 2026-06-18
feat: adding/updating dataset/video/connection getters for run.py and asset.py

george-nmnl:
This change allows users to get datasets, videos, or connections by ref name directly from run objects in the SDK, where before this functionality was limited to assets only. Also altered logic to avoid unnecessary iterations and complexity within getters on assets. Added end to end tests and requisite helpers to standardize access to objects.


nominal/core/asset.py
```diff
@@ -465,28 +465,37 @@ def create_run(
         )
 
     def get_dataset(self, data_scope_name: str) -> Dataset:
-        """Retrieve a dataset by data scope name, or raise ValueError if one is not found."""
-        dataset = self.get_data_scope(data_scope_name)
-        if isinstance(dataset, Dataset):
-            return dataset
-        else:
-            raise ValueError(f"Data scope {data_scope_name} on asset {self.rid} is not a dataset")
+        """Retrieve a dataset by data scope name, or raise KeyError if one is not found."""
```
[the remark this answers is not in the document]
george-nmnl: Returning to ValueErrors which are slightly more true-to-form for what this error is anyway. This also addresses the following suggestion.

nominal/core/asset.py
```diff
@@ -465,28 +465,67 @@ def create_run(
...
+
+        Raises:
+            ValueError: If no dataset data scope exists with the provided name.
+        """
+        dataset_rids_by_scope_name = self._scope_rids("dataset")
+
+        if data_scope_name not in dataset_rids_by_scope_name:
+            raise ValueError(f"No dataset with data scope name '{data_scope_name}' found for this asset")
+
+        dataset_rid = dataset_rids_by_scope_name[data_scope_name]
```
drake-nominal: Slightly simpler pattern we can do:

```
dataset_rid = dataset_rids_by_scope_name.get(data_scope_name)
if dataset_rid is None:
    raise ...

return Dataset._from_conjure(...)
```
saves us a line x6 :)

tests/e2e/test_core.py
```diff
@@ -118,6 +132,35 @@ def test_add_dataset_to_run_and_list_datasets(client: NominalClient, csv_data, a
...
+    assert run.get_connection(CONNECTION_REF).rid == connection.rid
+    assert run.get_video(VIDEO_REF).rid == video.rid
+
+    with pytest.raises(ValueError, match=f"No connection with ref name '{DATASET_REF}' found for this run"):
+        run.get_connection(DATASET_REF)
+    with pytest.raises(ValueError, match=f"No video with ref name '{CONNECTION_REF}' found for this run"):
+        run.get_video(CONNECTION_REF)
+    with pytest.raises(ValueError, match=f"No dataset with ref name '{VIDEO_REF}' found for this run"):
+        run.get_dataset(VIDEO_REF)
+
```
drake-nominal: can we split into 3 tests, one per modality?

tests/conftest.py
```diff
@@ -49,3 +54,70 @@ def _make(counts: list[int | None]):
...
+    return Connection(
+        rid="connection-rid-1",
+        name="Test Connection",
+        description=None,
+        _clients=mock_clients,
+    )
+
+
+@pytest.fixture
+def mock_video(mock_clients):
```
drake-nominal: I think there's 2 new ones we're adding here, but we can probably leave the ones that existed where they were and skip moving to conftest.py now that we got rid of the unit tests.

george-nmnl: Looks like Claude and you agree.

tests/e2e/test_core.py
george-nmnl: Will do

nominal/core/asset.py
george-nmnl: This is nice. Will change.

drake-nominal (approved):

PR #833, opened 2026-06-18
fix(video): adaptive, safe frame-rate normalization in normalize_video

drake-nominal:
## Summary

`normalize_video` re-encoded video without controlling output frame timing. For sources whose container duration is inflated relative to their true frame count, the muxed `avg_frame_rate` (= frames / duration) came out low, so downstream consumers that read `avg_frame_rate` saw the wrong rate; concatenated outputs also accumulated timing drift. This reworks `normalize_video` into a safe, general-purpose adaptive normalizer.

## Approach

**Clean-encode layer (every source):** selected codec (`h264` / `h264_nvenc` / `hevc` / `hevc_nvenc`), `yuv420p`, no B-frames (`-bf 0`), keyframe interval.

**Adaptive timing** — chosen per source by a tiered classifier (cheap `r_frame_rate` vs `avg_frame_rate`; per-frame timestamps sampled only when ambiguous):

- **CFR-heal** — re-grid at the source's true rate with a **numerator-derived timescale** (exact for 23.976 / 29.97 / 59.94, not just ~30 fps), duplicating the previous frame to fill gaps left by dropped/corrupt frames, with a guarded duration clamp so gap-fill can't overshoot.
- **passthrough** — preserve timing for genuine variable-frame-rate or indeterminate-rate sources (never re-grids → never drops real frames).
- **hard `NormalizationError`** for timestamp discontinuities/wraps, a missing video stream, or probe/encode failure.

**Post-encode invariant check** re-probes the output and raises rather than emit a file that lost frames or drifted in duration — no silent corruption.

## API / behavior

- New `frame_rate_mode: "auto" | "cfr" | "passthrough"` (default `auto`).
- `NormalizationError` raised on every unrecoverable case; re-exported from the package.
- `frame_count` now counts decoded frames (`nb_read_frames`) rather than packets.
- Requires ffmpeg ≥ 5.1 at runtime (for `-fps_mode`).

## Testing

Pure unit tests cover the decision logic — rate parsing, tiered classification, timing/encode kwarg assembly, and the post-encode invariant checks — by feeding canned inputs to small pure functions. **The suite invokes no real ffmpeg/ffprobe**, so it runs anywhere; end-to-end behavior was verified manually.

### DIFF
--- a/check_normalize_video.py
+++ b/check_normalize_video.py
@@ -1,0 +1,269 @@
+#!/usr/bin/env python3
+"""Diagnostic for normalize_video.
+
+Runs nominal's `normalize_video` on one or more source videos and reports whether each ran
+safely and correctly: it probes the source and the normalized output and checks the guarantees
+the normalizer is supposed to provide.
+
+  * Output is h264/hevc, yuv420p, with no B-frames.
+  * No real frames are dropped (output frame count >= source frame count).
+  * Output duration stays aligned to the source.
+  * The muxed avg_frame_rate reads correctly (constant-rate output reports its true rate).
+
+A source that the normalizer *rejects* with a clear error (e.g. timestamp discontinuity, no video
+stream) is reported as SAFE-REJECTED -- that is correct behavior, not a failure.
+
+Requires ffmpeg/ffprobe on PATH (this actually re-encodes the file). Note: counting frames decodes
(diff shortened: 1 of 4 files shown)


check_normalize_video.py
alkasm, line 1: can you put this in a scripts/ directory or similar?

PR #834, opened 2026-06-20
feat: add Run.search_events method

brianburrous-nominal:
## Summary
Adds a `Run.search_events(...)` method, following the existing `Asset.search_events` pattern, scoped to all assets associated with the run.

- New `Run.search_events` delegates to the shared `_search_events` helper with `asset_rids=list(self.assets)`. Because the query builder ANDs asset clauses, this matches events associated with **all** of the run's assets — symmetric with `Run.create_event`, which associates new events with all the run's assets.
- Exposes the same filters as `Asset.search_events` (`search_text`, `after`/`before`, `labels`, `properties`, `created_by_rid`, `workbook_rid`, `data_review_rid`, `assignee_rid`, `event_type`, `origin_types`, `archive_status`).
- No plumbing change needed — `Run._Clients` already exposes the `event` service.

## Test plan
- Added `tests/core/test_run.py` with unit tests asserting the run's asset rids and `archive_status` are forwarded to `_search_events`.
- `uv run pytest tests/core/test_run.py` passes.
- `ruff check` passes.

### DIFF
--- a/nominal/core/_utils/query_tools.py
+++ b/nominal/core/_utils/query_tools.py
@@ -24,6 +24,17 @@ from nominal._utils.deprecation_tools import _NotProvided
 from nominal.ts import IntegralNanosecondsUTC, _SecondsNanos
 
 
+class AssetMatch(Enum):
+    """How a set of asset rids should be matched when searching for events.
+
+    Use ALL to require an event to reference every given asset (rids ANDed together),
+    or ANY to match events referencing at least one of the given assets (rids ORed together).
+    """
+
+    ALL = "ALL"
+    ANY = "ANY"
+
+
 class ArchiveStatusFilter(Enum):
     """Filter for archive status in search methods.
 
@@ -502,6 +513,7 @@ def create_search_events_query(  # noqa: PLR0912
     after: str | datetime | IntegralNanosecondsUTC | None = None,
     before: str | datetime | IntegralNanosecondsUTC | None = None,
     asset_rids: Iterable[str] | None = None,
+    asset_match: AssetMatch = AssetMatch.ALL,
     labels: Iterable[str] | None = None,
     properties: Mapping[str, str] | None = None,
     created_by_rid: str | None = None,
@@ -520,8 +532,13 @@ def create_search_events_query(  # noqa: PLR0912
     if before is not None:
         queries.append(event.SearchQuery(before=_SecondsNanos.from_flexible(before).to_api()))
     if asset_rids:
-        for asset in asset_rids:
-            queries.append(event.SearchQuery(asset=asset))
+        if asset_match is AssetMatch.ANY:
+            queries.append(
+                event.SearchQuery(assets=event.AssetsFilter(assets=list(asset_rids), operator=api.SetOperator.OR))
+            )
+        else:
+            for asset in asset_rids:
+                queries.append(event.SearchQuery(asset=asset))
     if labels:
         for label in labels:
             queries.append(event.SearchQuery(label=label))
--- a/nominal/core/event.py
+++ b/nominal/core/event.py
@@ -13,7 +13,7 @@ from nominal.core._event_types import EventType as EventType  # noqa: PLC0414
 from nominal.core._event_types import SearchEventOriginType as SearchEventOriginType  # noqa: PLC0414
 from nominal.core._utils.api_tools import HasRid, RefreshableMixin, rid_from_instance_or_string
 from nominal.core._utils.pagination_tools import search_events_paginated
-from nominal.core._utils.query_tools import ArchiveStatusFilter, create_search_events_query
+from nominal.core._utils.query_tools import ArchiveStatusFilter, AssetMatch, create_search_events_query
 from nominal.ts import IntegralNanosecondsDuration, IntegralNanosecondsUTC, _SecondsNanos, _to_api_duration
 
 
@@ -161,6 +161,7 @@ def _search_events(
     after: str | datetime | IntegralNanosecondsUTC | None = None,
     before: str | datetime | IntegralNanosecondsUTC | None = None,
     asset_rids: Iterable[str] | None = None,
+    asset_match: AssetMatch = AssetMatch.ALL,
     labels: Iterable[str] | None = None,
     properties: Mapping[str, str] | None = None,
     created_by_rid: str | None = None,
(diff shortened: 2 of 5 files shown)


nominal/core/run.py
```diff
@@ -244,6 +244,43 @@ def create_event(
...
+        """Search for events associated with all assets of this run.
+
+        See nominal.core.event._search_events for details.
+        """
+        return _search_events(
+            self._clients,
+            search_text=search_text,
+            after=after,
+            before=before,
+            asset_rids=list(self.assets),
```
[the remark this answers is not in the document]
brianburrous-nominal: Good catch — fixed in 4b5cc8a. Note the suggested `list(self.assets) or None` wouldn't fully fix it: passing `None` also adds no asset filter, so it'd still return all events. Instead `search_events` now short-circuits and returns `[]` when the run has no assets, which is the correct result under the union (ANY) semantics. Covered by `test_search_events_empty_assets_returns_no_events`.

nominal/core/run.py
```diff
@@ -244,6 +244,43 @@ def create_event(
...
+        data_review_rid: str | None = None,
+        assignee_rid: str | None = None,
+        event_type: EventType | None = None,
+        origin_types: Iterable[SearchEventOriginType] | None = None,
+        archive_status: ArchiveStatusFilter = ArchiveStatusFilter.NOT_ARCHIVED,
+    ) -> Sequence[Event]:
+        """Search for events associated with all assets of this run.
+
+        See nominal.core.event._search_events for details.
+        """
```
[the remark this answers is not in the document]
brianburrous-nominal: Addressed in 4b5cc8a. Note the semantics changed since this review: assets are now matched as a **union** (ANY), not ANDed — a multi-asset run matches events on *any* of its assets, which is the intuitive behavior you flagged. The docstring is now full google-style (Args/Returns) matching `create_event` and states the union semantics explicitly.

tests/core/test_run.py
```diff
@@ -0,0 +1,51 @@
...
+
+
+def test_search_events_passes_archive_status(mock_run):
+    """Run.search_events forwards archive_status to the shared event search helper."""
+    with patch("nominal.core.run._search_events", return_value=[]) as mock_search_events:
+        result = mock_run.search_events(archive_status=ArchiveStatusFilter.ANY)
+
+    assert result == []
+    mock_search_events.assert_called_once()
+    assert mock_search_events.call_args.kwargs["archive_status"] == ArchiveStatusFilter.ANY
```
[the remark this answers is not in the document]
brianburrous-nominal: Agreed — removed the trivial archive_status passthrough test. Replaced it with `test_search_events_empty_assets_returns_no_events` (real behavior) and added `test_create_search_events_query_asset_match_{all,any}` in test_query_tools to pin the actual query contract (ALL → one clause per asset; ANY → a single OR `AssetsFilter`). 4b5cc8a.

nominal/core/_utils/query_tools.py
```diff
@@ -520,8 +532,13 @@ def create_search_events_query(  # noqa: PLR0912
...
     if asset_rids:
-        for asset in asset_rids:
-            queries.append(event.SearchQuery(asset=asset))
+        if asset_match is AssetMatch.ANY:
+            queries.append(
+                event.SearchQuery(assets=event.AssetsFilter(assets=list(asset_rids), operator=api.SetOperator.OR))
+            )
+        else:
+            for asset in asset_rids:
+                queries.append(event.SearchQuery(asset=asset))
```
drake-nominal, line 541: I don't think this works in practice-- all of the queries get ANDed together whereas we probably need to build up a query that ORs together asset rids that gets thrown into the bigger AND tree.

notably, I think the _only_ correct behavior for `run.search` is ANY, and we can avoid the additional class by just taking in `asset_rids: Sequence[str]` (if we don't already) and handling them as we do with the ANY branch.

unless an event can be present on multiple assets at once, which complicates things.

drake-nominal: a bit of a roller coaster of a comment, maybe we can just have `asset_rids_all_match` vs `asset_rids_any_match` args which are mutually exclusive. idk. this is a bit of an annoying one to expose.

tests/core/test_run.py
```diff
@@ -0,0 +1,52 @@
...
+        run_number=1,
+        assets=["asset-rid-1", "asset-rid-2"],
+        created_at=0,
+        _clients=mock_clients,
+    )
+
+
+def test_search_events_ors_run_assets(mock_run):
+    """Run.search_events matches events on any of the run's assets (ANY)."""
+    with patch("nominal.core.run._search_events", return_value=[]) as mock_search_events:
```
drake-nominal: no monkeypatch allowed-- use `MagicMock` / `mock_clients` fixture to mock the API layer.

tests/core/test_run.py
```diff
@@ -0,0 +1,52 @@
...
+
+    assert result == []
+    mock_search_events.assert_called_once()
+    assert mock_search_events.call_args.kwargs["asset_rids"] == ["asset-rid-1", "asset-rid-2"]
+    assert mock_search_events.call_args.kwargs["asset_match"] == AssetMatch.ANY
+
+
+def test_search_events_empty_assets_returns_no_events(mock_run):
+    """A run with no associated assets returns no events instead of searching all events."""
+    object.__setattr__(mock_run, "assets", [])
```
drake-nominal: never use `__setattr__` in this repo.

alkasm:
```suggestion
    mock_run.assets = []
```

nominal/core/event.py
```diff
@@ -161,6 +161,7 @@ def _search_events(
     after: str | datetime | IntegralNanosecondsUTC | None = None,
     before: str | datetime | IntegralNanosecondsUTC | None = None,
     asset_rids: Iterable[str] | None = None,
+    asset_match: AssetMatch = AssetMatch.ALL,
```
alkasm, line 164: I worry about exposing this lever in this way; every other parameter can be combined in an OR or AND fashion as well, singling this param out for the purpose is not scalable to other parameters and sets a non-ideal precedent. added to the fact that this function already has way too many parameters and the documentation is going to be more confusing as we add these levers. we should probably bite the bullet and expose a query builder type instead.

tests/core/test_run.py
alkasm: I see that CI failed after applying this suggestion. That implies that this _isn't_ a "mock run" despite the name (i.e. it's not a `unittest.Mock` instance).

drake-nominal (approved):

PR #839, opened 2026-06-23
feat(grpc): migrate comments, units, and workspace services to gRPC

drake-nominal:
## Summary

Continues the conjure→gRPC migration started for the Role Service in #809. This migrates the three lowest-usage `ClientsBunch` services that have a backend gRPC stub available — **comments**, **units**, and **workspace** — to their gRPC variants, reusing the shared-channel `grpc_factory` pattern (auth/retry/TLS/deadlines already handled by the channel interceptors).

Migration order was chosen by measuring call-site usage across the SDK and overlaying which services actually have a gRPC stub; these three were the cleanest, lowest-blast-radius starting set.

## What changed

- **comments** → `CommentsServiceStub`: `Run.add_comment` builds a proto `CreateCommentRequest` and converts via a new `Comment._from_proto`.
- **units** → `UnitsServiceStub`: `get_all_units` / `get_unit` / `get_commensurable_units` and the `_available_units` / `_error_on_invalid_units` helpers; new `Unit._from_proto` (the multi-type `Unit._from_conjure` is retained for channel/run unit parsing). `get_unit` now returns `None` for an unrecognized symbol via proto field presence (`HasField`) instead of catching an HTTP error.
- **workspace** → `WorkspaceServiceStub`: `get_workspace` / `list_workspaces` plus the internal `ClientsBunch` resolve path (`resolve_workspace` / `_fetch_default_workspace` / cached `_default_workspace`), which now carries the proto `Workspace` type. New `Workspace._from_proto`. The "no default workspace" case still raises `NominalConfigError`.
- **gRPC error translation**: new `translate_grpc_errors()` context manager applied at every gRPC call site (including the pre-existing `roles` one, fixing a raw-`grpc.RpcError` leak). It maps `PERMISSION_DENIED` / `UNAUTHENTICATED` / `NOT_FOUND` / `INVALID_ARGUMENT` to four new `NominalError` subclasses and falls back to the base `NominalError` for other codes. Translation lives at the call boundary (not the stub) deliberately, so individual methods can still catch specific status codes.
- **SandboxWorkspaceService** (no gRPC equivalent yet) promoted to a first-class conjure `ClientsBunch` field, removing the previous transport-internals-borrowing hack in the migration CLI.
- `cli/util/verify_connection.py` updated to catch the now-gRPC-translated `NominalError` from `get_workspace` (previously caught `ConjureHTTPError`).

## Compatibility

Transport swap only — public signatures and return types (`Comment`, `Unit`, `Workspace`) are unchanged. The new exception subclasses extend `NominalError`, so existing `except NominalError` handlers keep working.

## Testing

- `uv run pytest tests/core tests/cli -q` → **175 passed, 10 skipped, 0 failed**
- `uv run mypy -p nominal` → clean (only the pre-existing `nominal.tdms: import-untyped`)
- `uv run ruff check nominal/ tests/` → clean

Built via TDD task-by-task with per-task review plus a whole-branch review (which caught and fixed the `verify_connection.py` consumer of the migrated `get_workspace`).


nominal/core/_utils/grpc_tools.py
```diff
@@ -286,3 +294,28 @@ def factory(stub_class: type[TStub]) -> TStub:
...
+_GRPC_STATUS_TO_EXCEPTION: dict[grpc.StatusCode, type[NominalError]] = {
+    grpc.StatusCode.PERMISSION_DENIED: NominalPermissionDeniedError,
+    grpc.StatusCode.UNAUTHENTICATED: NominalAuthenticationError,
+    grpc.StatusCode.NOT_FOUND: NominalNotFoundError,
+    grpc.StatusCode.INVALID_ARGUMENT: NominalInvalidArgumentError,
+}
+
+
+@contextmanager
+def translate_grpc_errors() -> Iterator[None]:
```
alkasm, line 308: can leave for a follow-on PR, but we can do this with interceptors rather than wrapping every API call

nominal/core/unit.py
```diff
@@ -39,6 +41,11 @@ def _from_conjure(cls, api_unit: api.Unit | scout_units_api.Unit | scout_run_api
         name = "" if api_unit.name is None else api_unit.name
         return cls(name=name, symbol=api_unit.symbol)
 
+    @classmethod
+    def _from_proto(cls, unit: units_pb2.Unit) -> Self:
```
alkasm, line 45: can we delete `Unit._from_conjure` or not yet?

tests/core/test_client_units.py
```diff
@@ -0,0 +1,83 @@
...
+from nominal.protos.units.v1 import units_pb2
+
+
+def test_available_units_flattens_units_by_property() -> None:
+    """get_all_units flattens the units-by-property map into a single flat list of units."""
+    stub = MagicMock()
+    resp = units_pb2.GetAllUnitsResponse()
+    resp.units_by_property["electric"].value.extend(
+        [units_pb2.Unit(name="coulomb", symbol="C"), units_pb2.Unit(name="ampere", symbol="A")]
+    )
```
alkasm:
```suggestion
    resp = units_pb2.GetAllUnitsResponse(
        units_by_property={
            "electric": [
                units_pb2.Unit(name="coulomb", symbol="C"),
                units_pb2.Unit(name="ampere", symbol="A"),
            ]
        }
    )
```
always construct protos in-place rather than initialize and mutate.

we should write a lint rule for this

tests/core/test_clientsbunch.py
```diff
@@ -50,12 +51,8 @@ def _make_clients_bunch(*, workspace_rid: str | None) -> ClientsBunch:
...
 
 
-def _raw_workspace(rid: str) -> MagicMock:
-    workspace = MagicMock()
-    workspace.rid = rid
-    workspace.id = rid.rsplit(".", 1)[-1]
-    workspace.org = "test-org"
-    return workspace
+def _ws(rid: str) -> workspaces_pb2.Workspace:
+    return workspaces_pb2.Workspace(rid=rid, id=rid.rsplit(".", 1)[-1], org="test-org")
```
alkasm: not that it matters but workspace IDs are not the locator component of the rid; the ID is the "name" of the workspace

nominal/core/unit.py
drake-nominal: still used in other places sadly, can't delete quite yet.

nominal/core/_utils/grpc_tools.py
drake-nominal: I originally had this as an interceptor, but decided to do as a context manager to allow various routes freedom to handle errors in a bespoke way if needed-- for example, maybe one of these errors means something specific to one api or another. down the road, would be easy to convert to a interceptor if that ends up not being important / done.

alkasm: hmm but isn't the only difference that at the callsite you'd be catching `grpc.RpcError` vs `NominalInvalidArgumentError` or similar? the exception is still thrown at the same line. and you can always grab the `__cause__` if it's a wrapped exception like these

alkasm (approved):

PR #842, opened 2026-06-25
feat: v2 hosted containerized extractors + image upload/registration

drake-nominal:
Adds canonical support for registering container images and containerized extractors, and ingesting with them, via the v2 gRPC surface (`nominal.ingest.v2.ContainerizedExtractorService` + `nominal.registry.v2.RegistryService`).

## The v2 model

An extractor carries **identity** (name / description / archived); its **execution contract** — inputs, parameters, output format, default timestamp metadata — lives on the container images registered against it, exactly one of which is active:

```python
extractor = client.create_containerized_extractor("my-parser")
image = extractor.register_image(
    "extractor.tar",              # `docker save` tarball, uploaded to Nominal's registry
    tag="v1",
    inputs=[FileExtractionInput(name="Input file", environment_variable="INPUT_FILE")],
    default_timestamp_column="timestamp",
    default_timestamp_type="iso_8601",
)
extractor.set_active_image(image)  # polls until READY by default, then activates
job = dataset.add_containerized(extractor, sources={"INPUT_FILE": "./data.bin"})
files = job.as_files_ingested()   # block until the extractor's outputs are ingested
```

## What's in here

**Commit 1 — `refactor(core)`:** splits `RefreshableMixin` into `RefreshableConjureMixin` / `RefreshableGrpcMixin` so proto-backed resources reuse the in-place refresh machinery.

**Commit 2 — `feat(grpc)`:**
- `ContainerizedExtractor` (`nominal/core/containerized_extractor.py`): create/get/search, partial `update()`, archive/unarchive, `register_image()`, `set_active_image(poll_until_ready=True)`.
- `ContainerImage` (`nominal/core/container_image.py`): refresh/delete/`poll_until_ready()`, string-valued enums with mypy-exhaustive proto conversion, forward-compatible decoding of unknown enum values.
- `Dataset.add_containerized` validates required inputs against the **active image** and returns the tracking `IngestionJob` (containerized extraction is async and multi-file).
- `nominal/ts`: polymorphic `_to_proto()` on timestamp types + exhaustive (`assert_never`) proto→SDK decoding.
- gRPC search pagination follows the conjure conventions (`paginate_grpc` sibling of `paginate_rpc`; queries built in `query_tools`).
- `register_image` rejects output formats the backend cannot currently ingest (`REGISTERABLE_OUTPUT_FORMATS`) **before** uploading the tarball.

## Breaking changes — deliberately shipped as a minor

Removes the v1 conjure containerized-extractor API: `DockerImageSource`, `UserPassAuth`, `TagDetails`; `add_containerized` loses `tag` and returns `IngestionJob` instead of `DatasetFile`; `create_containerized_extractor` / `search_containerized_extractors` signatures rewritten. **Telemetry showed zero customer usage of the removed API**, so this intentionally ships as `feat:` (minor) rather than burning a major version on a no-op break.

## Notes

- Requires a backend with the v2 services deployed; the tag-less containerized ingest resolves against the extractor's active image (verified against the current backend, including MANIFEST fan-out with per-output timestamp overrides).
- `default_timestamp_metadata` is required at registration (backend `buf.validate`); per-ingest and per-manifest-output overrides take precedence — resolution order documented on `ContainerImage.default_timestamp_metadata`.

## Backend follow-ups (tracked from review)

- [ ] `CreateImage`: server-side validation rejecting non-ingestible output formats (defense-in-depth for non-SDK clients).
- [ ] `SearchImages`: add `ExtractorFilter` to the `SearchFilter` oneof (images are keyed by extractor but can't be searched by it); then expose `ContainerizedExtractor.list_images()`.
- [ ] Optional: relax `workspace_rid` on registry Get/Delete requests (server resolves from rid) — would let `ContainerImage` drop its stored workspace and use plain `RefreshableGrpcMixin`.
- [ ] Decide whether default-less images should be legal (`CreateImageRequest.default_timestamp_metadata` required today); if relaxed, `register_image` gains both-or-neither overloads.


varun-nominal (approved):

drake-nominal:
Evaluated the structural finding (collapse `_DatasetWrapper.add_containerized`'s timestamp branch into one unconditional forward): **rejected — the collapse doesn't type-check.**

`Dataset.add_containerized` expresses the both-or-neither contract through its overloads: one variant with no timestamp params, one requiring both non-None. The wrapper holds `timestamp_column: str | None`, so the proposed unconditional forward fails mypy:

```
error: Argument "timestamp_column" to "add_containerized" of "Dataset" has incompatible type "str | None"; expected "str"  [arg-type]
```

The wrapper's two branches exist to narrow the Optionals so each delegate call matches an overload, and the `else` raise is likewise forced — an XOR'd pair cannot be forwarded through the typed boundary at all, so the wrapper must reject it locally. Collapsing would require either dropping the inner overloads (weakening the public type contract) or `cast`/`type: ignore` at the call site. The duplication is the cost of expressing both-or-neither in the type system, mirroring the same overload shape on `Dataset.add_containerized` itself.

PR #843, opened 2026-06-25
chore(deps): bump nominal-api & nominal-api-protos to 0.1287.0

drake-nominal:
## What

Bumps `nominal-api` and `nominal-api-protos` from `0.1282.0` → `0.1287.0`, and cleans up gRPC `# type: ignore` comments that the new stubs make unnecessary.

## Why the type-ignore cleanup

`nominal-api-protos==0.1287.0` now ships `*_pb2_grpc.pyi` type stubs (and a `py.typed` marker) for the gRPC route stubs — e.g. `WorkspaceServiceStub.GetWorkspace` is typed as `grpc.UnaryUnaryMultiCallable[GetWorkspaceRequest, GetWorkspaceResponse]`.

On their own those stubs change nothing for mypy: they're typed in terms of `grpc.*`, and `grpc` was untyped here, so the subscripts collapsed to `Any`. Adding **`types-grpcio`** (dev dependency) makes `grpc` concrete, which lets the route-stub return types resolve and renders several ignores unused. mypy (`strict = true`, so `warn_unused_ignores` is on) then flagged exactly these:

| File | Removed |
|------|---------|
| `nominal/core/_clientsbunch.py` | `# type: ignore[no-any-return]` on the two `WorkspaceServiceStub` calls |
| `nominal/core/_utils/grpc_tools.py` | `# type: ignore[import-untyped]` on `import grpc` |
| `nominal/core/_utils/grpc_tools.py` | `# type: ignore[misc]` on 5 grpc base-class subclassings (`ClientCallDetails`, the 4 interceptor bases) |

Two now-inaccurate "untyped gRPC stub" comments were refreshed in the process.

### Intentionally kept

`grpc_tools.py` — the `[call-arg]` ignore on the stub factory stays. It's not about grpc typing: `stub_class` is a generic `type[TStub]` whose constructor mypy can't resolve. mypy confirms it's still needed.

## Verification

- `uv run mypy` — clean (`Success: no issues found in 143 source files`)
- `uv run pytest` — 251 passed, 10 skipped
- `uv run ruff format --check` / `uv run ruff check` — clean
- `uv lock --check` — passes
- Enum-safety spot check: `IngestStatusV2` in `0.1287.0` has 8 union members, all handled by `IngestStatus._from_conjure` (the spot fixed in #838) — no new unhandled members.

### DIFF
--- a/nominal/core/_utils/grpc_tools.py
+++ b/nominal/core/_utils/grpc_tools.py
@@ -1,8 +1,8 @@
 """gRPC transport plumbing for the Nominal client.
 
-This module is the gRPC analogue of the conjure stub factory in `_utils.networking`: it builds a single,
-shared, fully-configured `grpc.Channel` and returns a `create_grpc_stub_factory(StubClass) -> stub`
-closure, so each backend gRPC service is exposed as a generated stub bound to that one channel.
+This module is the gRPC analogue of the conjure transport in `_utils.networking`: it builds a single,
+shared, fully-configured `grpc.Channel` (`create_grpc_channel`) that each backend gRPC service binds its
+generated stub to, so every stub shares that one channel.
 
 The channel is configured to track the conjure HTTP transport as closely as gRPC allows:
 
@@ -22,7 +22,7 @@ from collections import namedtuple
 from contextlib import contextmanager
 from functools import lru_cache
 from pathlib import Path
-from typing import Any, Callable, Iterator, TypeVar
+from typing import Any, Iterator, Protocol, TypeVar
 from urllib.parse import urlparse
 
 import grpc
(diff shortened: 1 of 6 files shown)


nominal/core/_utils/grpc_tools.py
```diff
@@ -291,7 +291,8 @@ def create_grpc_stub_factory(
     )
 
     def factory(stub_class: type[TStub]) -> TStub:
-        return stub_class(channel)  # type: ignore[call-arg]  # grpc stub constructors are untyped
+        # stub_class is a generic ``type[TStub]``; mypy can't see the channel-accepting stub constructor.
+        return stub_class(channel)  # type: ignore[call-arg]
 
     return factory
```
alkasm: If `TStub` is covariant, then you can do something like:

```py
from typing import Protocol

class GRPCStub(Protocol[TStub]):
    def __call__(self, channel: grpc.Channel) -> TStub: ...

def create_grpc_stub_factory(...) -> Callable[[GRPCStub[TStub]], TStub]:
```

and drop the ignore

alkasm: also this is super minor but i'd probably inline this factory in the same place that the conjure factory is inlined, and just have the above logic to create a channel in here?

pyproject.toml
```diff
@@ -74,6 +74,7 @@ dev = [
     "pytest-cov>=5.0.0,<6",
     "pytest>=8.3.2,<9",
     "ruff==0.12.1",
+    "types-grpcio>=1.0.0.20260614,<2",
```
alkasm, line 77: what a weird versioning scheme
- you've heard of semver
- you've heard of calver

...what if...we had..both?

drake-nominal: ❇️ google ❇️

alkasm (approved):

PR #846, opened 2026-06-26
feat(migration): support migrating standalone checklists

seanmreidy:
## Summary
Adds the ability to migrate **standalone checklists** (not tied to assets) given a list of checklist RIDs, mirroring the existing `standalone_workbook_template_rids` flow.

Only the checklist *definition* is cloned (via `ChecklistMigrator.copy_from` — `clone()` is unsupported for checklists). No run or execution is involved. Channel references resolve at execution time against whatever run the checklist is later run against, provided that run exposes matching data-source **ref names** and **channel names** (and tags, if any). This is the same portability mechanism checklists already rely on, so no new constraint is introduced.

## Changes
- `config/migration_resources.py` — add `source_standalone_checklists`
- `migration_cli.py` — parse `standalone_checklist_rids`; surface the key in the generated `prep` config; include count in log
- `parallel_migration_runner.py` (the path the CLI uses) and `migration_runner.py` (sequential, for parity with e2e tests) — clone via `copy_from(checklist, ChecklistCopyOptions())`
- `sample_migration_config.yml` + `README.md` — document the new key and the resolution caveat
- `tests/e2e/migration/test_migration.py` — `test_migrate_standalone_checklist`

## Usage
```yaml
migration:
  standalone_checklist_rids:
    - ri.scout.main.checklist.<uuid>
```

## Verification
- `just check-format`, `just check-imports`, `just check-types` (mypy strict, py3.10–3.14) — pass
- `just test` (unit) — 348 passed
- Migration e2e runs in CI (needs staging credentials)
- Ran a manual test checklist clone to verify checklist cloning without data still resolves an execution once data is available, via a test from demo-sandbox to staging-sandbox.

### DIFF
--- a/nominal/experimental/migration/README.md
+++ b/nominal/experimental/migration/README.md
@@ -75,6 +75,8 @@ migration:
     - asset_rid: ri.scout.main.asset.<uuid>
   standalone_workbook_template_rids:
     - ri.scout.main.template.<uuid>
+  standalone_checklist_rids:     # optional: clone checklist definitions, no asset/run required
+    - ri.scout.main.checklist.<uuid>
   impersonation:                 # optional: create resources as mapped destination users
     enabled: true
     unmapped_source_user_behavior: service_user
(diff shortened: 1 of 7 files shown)


seanmreidy:
Addressed both inline comments in f048ff1:

- **Dead guard / friendly error** — wrapped `get_checklist` in `try/except` so a bad RID surfaces as a `UsageError` instead of a raw conjure error (confirmed `get_checklist` raises rather than returning falsy).
- **Type** — `_load_standalone_checklists` now returns `list[Checklist]`.
- **Test** — `test_migrate_standalone_checklist` now sets and asserts `description`/`labels`/`properties` round-trip (not just `name`), so a metadata-dropping regression is caught. (Left check/variable assertions out since the fixture checklist has none; the metadata round-trip mirrors `test_migrate_standalone_template`.)

Kept the pre-existing `_load_standalone_templates` / `_parse_asset_entry` dead guards and `list[Any]` returns as-is to keep this PR scoped — they share the same pattern and are worth a separate cleanup. The structural dedup suggestion (a generic `_load_standalone` + task-builder) is noted; holding off until a fourth standalone type actually lands.

seanmreidy:
@claude

varun-nominal (approved):

PR #853, opened 2026-06-30
fix(migration): allow empty/absent source_asset_rids

seanmreidy:
## Problem

In the migration runner (`nominal/experimental/migration`), `_load_asset_resources` required `migration.source_asset_rids` to be a non-empty list. A config that omits it or sets it to an empty list raised a `UsageError`, even when the intent was to migrate only `standalone_workbook_template_rids`.

## Fix

Treat an absent or empty `source_asset_rids` as "no assets to migrate" — return an empty asset map so the migration skips assets and proceeds to `standalone_workbook_template_rids`. A non-list value (e.g. a bare string) is still rejected as a config error.

## Tests

Added `tests/core/test_migration_source_assets.py` covering: `None`, empty list, non-list rejection, and the both-provided error.

### DIFF
--- a/nominal/experimental/migration/migration_cli.py
+++ b/nominal/experimental/migration/migration_cli.py
@@ -113,8 +113,12 @@ def _load_asset_resources(
         raise click.UsageError("Provide only one of 'migration.source_asset_rids' or 'migration.source_assets'.")
 
     if source_assets is None:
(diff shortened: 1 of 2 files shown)


ckniffin-nominal (approved):
lgtm

PR #872, opened 2026-07-08
feat(migration): offline `nom migrate summary` command + shared dry-run log constants

tarun-prakash-nmnl:
Follow-up to review comments on [internal-tools PR #45](https://github.com/nominal-io/internal-tools/pull/45): the migration summary tool moves into this repo so it lives next to the dry-run code it parses, benefits users of this repo alone, and cannot drift from the migrator log statements. Companion PR consuming this: [nominal-io/internal-tools#49](https://github.com/nominal-io/internal-tools/pull/49) — **this PR must merge and release first** (predicted `1.152.0`).

**What's here:**

- **`nominal/experimental/migration/dry_run.py`** (new) — shared dry-run logging vocabulary: `DRY_RUN_PREFIX`, `would_create_message(resource_type)` (builds the exact `[DRY RUN] Would create <label> '%s' (source: %s)` format string), and `dry_run_create_pattern()` (the parsing regex built from the same constants and `ResourceType` labels). Labels are alternated longest-first, which also fixes a real bug in the internal-tools parser: `workbook template` lines were miscounted as `workbook`.
- **All 16 dry-run log sites** in the migrators/runner now build their messages from the shared constants. Wording is byte-identical to before.
- **`nominal/experimental/migration/migration_summary.py`** (new) + **`nom migrate summary`** subcommand — emits a markdown resource-count table from one of three sources: `--from-config <yml>` (repeatable; offline size check), `--from-log <log>` (tallies dry-run would-create lines), `--from-state <json>` (counts `rid_mapping` entries via `MigrationState.from_json`). `--output` appends to a file such as `$GITHUB_STEP_SUMMARY`. Fully offline — no profiles or tokens required (`@global_options` only).
- `load_migration_block` moved out of `migration_cli.py` into the new module for reuse (import direction: cli → summary, no cycle).
- Sample config and README updated; the sample now carries the never-store-tokens warning previously only in internal-tools.

**From review:**
- `summarize_config` mirrors all of `copy`'s config validation — both-asset-forms, non-list `source_asset_rids`, and non-mapping/empty `source_assets` are rejected with `copy`'s exact error messages instead of silently reporting 0 assets (Cursor Bugbot + Claude reviewer).
- New string-level `format_resource_label()` in `resource_type.py`; `resource_label()` delegates to it and `summarize_state` uses it, so state-summary labels come from the same formatter as log labels (Claude reviewer).
- On why two asset forms exist (`source_asset_rids` list vs `source_assets` map): both are pre-existing `copy` schema — see the [thread](https://github.com/nominal-io/nominal-client/pull/872#discussion_r3544749147); consolidation would be a breaking config change, offered as a follow-up.

**Tests:** new `tests/core/test_migration_summary.py` (round-trips every `ResourceType` through message-builder → parser; multi-word-label regression; non-create lines ignored; config validation rejections; snake/camelCase state; `DRY_RUN_PREFIX` lock; label-formatter delegation) and `tests/cli/test_migrate_summary_cli.py` (each mode end-to-end via `CliRunner` with no profile configured; `--output` append; usage errors). Existing literal assertion in `test_asset_migrator.py` now derives from the constants.

**Verification:** full suite passes under Python 3.13, `ruff format --check`/`ruff check` clean, `mypy -p nominal.experimental.migration` strict clean. Devin Review: no issues found.

https://claude.ai/code/session_01Fc2sY63WBF9rruL6xLEWpM

### DIFF
--- a/nominal/experimental/migration/migration_summary.py
+++ b/nominal/experimental/migration/migration_summary.py
@@ -23,6 +23,7 @@ import yaml
 
 from nominal.experimental.migration.dry_run import dry_run_create_pattern
 from nominal.experimental.migration.migration_state import MigrationState
+from nominal.experimental.migration.resource_type import format_resource_label
 
 
 def load_migration_block(raw: Any) -> dict[str, Any]:
@@ -59,10 +60,14 @@ def summarize_config(path: Path) -> tuple[str, dict[str, int]]:
 
     source_asset_rids = migration.get("source_asset_rids")
     source_assets = migration.get("source_assets")
-    # Mirror _load_asset_resources: `nom migrate copy` rejects configs with both forms, so the
-    # size check must not silently pick one and report misleading counts.
+    # Mirror _load_asset_resources: whatever `nom migrate copy` rejects, the size check must
+    # reject too rather than silently reporting misleading counts.
     if source_asset_rids is not None and source_assets is not None:
         raise click.UsageError("Provide only one of 'migration.source_asset_rids' or 'migration.source_assets'.")
+    if source_asset_rids is not None and not isinstance(source_asset_rids, list):
+        raise click.UsageError("'migration.source_asset_rids' must be a list.")
+    if source_assets is not None and (not isinstance(source_assets, dict) or not source_assets):
+        raise click.UsageError("'migration.source_assets' must be a non-empty mapping.")
     template_total = 0
     if isinstance(source_asset_rids, list):
         counts["assets"] = len(source_asset_rids)
@@ -108,7 +113,7 @@ def summarize_state(path: Path) -> tuple[str, dict[str, int]]:
     """Count old->new RID mappings per resource type from a migration-state JSON file."""
     state = MigrationState.from_json(path.read_text(encoding="utf-8"))
     counts = {
-        resource_type.lower().replace("_", " "): len(mapping)
+        format_resource_label(resource_type): len(mapping)
         for resource_type, mapping in state.rid_mapping.items()
         if isinstance(mapping, dict)
     }
--- a/nominal/experimental/migration/migration_summary.py
+++ b/nominal/experimental/migration/migration_summary.py
@@ -59,6 +59,10 @@ def summarize_config(path: Path) -> tuple[str, dict[str, int]]:
 
     source_asset_rids = migration.get("source_asset_rids")
     source_assets = migration.get("source_assets")
+    # Mirror _load_asset_resources: `nom migrate copy` rejects configs with both forms, so the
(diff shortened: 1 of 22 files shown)


nominal/experimental/migration/migration_summary.py
```diff
@@ -0,0 +1,147 @@
...
+def summarize_config(path: Path) -> tuple[str, dict[str, int]]:
+    """Static counts from a config YAML, without contacting any tenant."""
+    with path.open("r", encoding="utf-8") as f:
+        raw: Any = yaml.safe_load(f)
+    migration = load_migration_block(raw)
+
+    counts: dict[str, int] = {}
+
+    source_asset_rids = migration.get("source_asset_rids")
+    source_assets = migration.get("source_assets")
```
seanmreidy, line 62: I think this should only be source_asset_rids?

nominal/experimental/migration/migration_summary.py
```diff
@@ -0,0 +1,143 @@
...
+    template_total = 0
+    if isinstance(source_asset_rids, list):
+        counts["assets"] = len(source_asset_rids)
+        entries: list[Any] = source_asset_rids
+    elif isinstance(source_assets, dict):
+        counts["assets"] = len(source_assets)
+        entries = list(source_assets.values())
+    else:
+        counts["assets"] = 0
+        entries = []
```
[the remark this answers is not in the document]
seanmreidy: I'm actually confused why both keys exist

seanmreidy (approved):
One minor nit/question

nominal/experimental/migration/migration_summary.py
tarun-prakash-nmnl: Both forms are pre-existing `nom migrate copy` schema (this PR didn't add them — see `_load_asset_resources` in migration_cli.py, which enforces they're mutually exclusive): `source_asset_rids` is the simple list form that `nom migrate prep` generates, while the `source_assets` map form exists to attach per-asset `workbook_template_rids` keyed by asset RID. Both are documented in `sample_migration_config.yml`.

The summary has to accept exactly what `copy` accepts — if it only counted `source_asset_rids`, it would report 0 assets for a valid map-form config, which is the drift class this PR is meant to eliminate. Just pushed 9df7581, which also mirrors `copy`'s remaining validation here (non-list `source_asset_rids` / non-mapping `source_assets` now error with the same messages instead of counting 0).

If consolidating the config schema down to a single form is preferred, that's a breaking config change to `copy` itself — happy to file it as a separate follow-up.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #873, opened 2026-07-08
fix(migration): persist migration state incrementally and flush it on SIGINT/SIGTERM

tarun-prakash-nmnl:
A cancelled or killed `nom migrate copy` run previously lost **all** migration state, making a half-done migration unresumable even though the resources were already created on the destination. Found live: [internal-tools run 28962064479](https://github.com/nominal-io/internal-tools/actions/runs/28962064479) was cancelled mid-migration and wrote no state file, so the [resume run](https://github.com/nominal-io/internal-tools/actions/runs/28962331251) correctly failed with "no prior state exists".

**Root cause:** state was only written in a `finally` after *every* task completed (`run_parallel_migration`) — and that `finally` is unreachable on cancellation: the KeyboardInterrupt unwind blocks in the executor's `shutdown(wait=True)` waiting for in-flight copies, and GitHub Actions hard-kills the process ~7.5s after SIGINT, long before they finish. RID mappings for everything already migrated die with the process.

**Fixes (layered — every catchable exit path now saves, and uncatchable kills lose at most in-flight work):**
- **Incremental saves** — `run_concurrent` gains an `on_task_complete` callback; the parallel runner saves state after every settled task (success *and* failure). The sequential `MigrationRunner.run_migration` likewise saves after each asset/template/checklist. State JSON is tiny, so the extra writes are negligible. This is the only cover for SIGKILL/OOM/runner death.
- **Signal flush** — new `_flush_state_on_termination` context manager: on SIGINT/SIGTERM it saves state immediately (well within Actions' grace window, covering cancellation mid-task during long dataset-file copies), then restores the original handler and re-raises so exit semantics are unchanged. No-op off the main thread.
- **Reachable final save** — on any raised exit (incl. KeyboardInterrupt/SystemExit) the executor is shut down with `wait=False, cancel_futures=True`, so the `finally` save runs immediately instead of blocking behind in-flight copies; the success path still waits for a clean shutdown. This gives local Ctrl-C the classic try/finally guarantee without relying on the signal handler.
- **Atomic writes** — `save_state` writes via temp-file + `os.replace`, so a kill mid-write can never leave a truncated (unresumable) state file. Matters more now that saves are frequent.
- **Locked, reentrant serialization** — `ThreadSafeMigrationState.to_json` takes the mutation lock (serialization now runs while workers are still recording), and the lock is an `RLock` so the signal handler can't deadlock if it fires while the main thread is mid-save (per Bugbot review).

**Tests:** new `tests/core/test_migration_state_persistence.py` (11 tests) — callback fires once per task including failed ones; SIGINT inside the context saves a resumable state file and still raises `KeyboardInterrupt`; handlers restored; interrupt reaches the final save without waiting on an in-flight (Event-gated) task; atomic write leaves no `.tmp` residue and round-trips through `MigrationState.from_json`; reentrant `to_json` under a held lock.

**Verification:** full suite passes under Python 3.13 (451 passed), `ruff format --check`/`ruff check` clean, `mypy -p nominal.experimental.migration` strict clean.

Companion change in internal-tools adds a loud `::warning::` when a non-preview workflow run ends with no state file, and normalizes the client's versioned state files (`state_v2.json`, …) to the uploaded artifact so resume-of-resume keeps the newest state.

https://claude.ai/code/session_01Fc2sY63WBF9rruL6xLEWpM

### DIFF
--- a/nominal/experimental/migration/migration_runner.py
+++ b/nominal/experimental/migration/migration_runner.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import logging
+import os
 import re
 from pathlib import Path
 from typing import cast
@@ -117,15 +118,18 @@ class MigrationRunner:
                         workbook_rids_allowlist=asset_resources.source_workbook_rids,
                     ),
                 )
+                self.save_state()
 
             for source_template in self.migration_resources.source_standalone_templates:
                 template_migrator.clone(source_template)
+                self.save_state()
 
             for source_checklist in self.migration_resources.source_standalone_checklists:
                 # ChecklistMigrator.clone() raises NotImplementedError; use copy_from() to clone the
                 # checklist definition. No run/execution is involved — series resolve at execution time
                 # against whatever run the checklist is later run against.
                 checklist_migrator.copy_from(source_checklist, ChecklistCopyOptions())
+                self.save_state()
 
             source_clients_by_asset_rid: dict[str, ClientsBunch] = {
                 asset_rid: cast(ClientsBunch, asset_resources.asset._clients)
@@ -142,4 +146,8 @@ class MigrationRunner:
             logger.info(f"{DRY_RUN_PREFIX} Skipping migration state write to %s", self.migration_state_path)
             return
         self.migration_state_path.parent.mkdir(parents=True, exist_ok=True)
-        self.migration_state_path.write_text(self.migration_state.to_json(), encoding="utf-8")
+        # Write-then-rename so a process killed mid-write can never leave a truncated state file
+        # behind — state is saved incrementally and a corrupt file would break resume.
+        tmp_path = self.migration_state_path.with_name(f"{self.migration_state_path.name}.tmp")
+        tmp_path.write_text(self.migration_state.to_json(), encoding="utf-8")
+        os.replace(tmp_path, self.migration_state_path)
--- a/nominal/experimental/migration/migration_runner.py
+++ b/nominal/experimental/migration/migration_runner.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import itertools
 import logging
 import os
 import re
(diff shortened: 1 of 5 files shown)


nominal/experimental/migration/migration_runner.py
```diff
@@ -117,15 +118,18 @@ def run_migration(self) -> None:
                         workbook_rids_allowlist=asset_resources.source_workbook_rids,
                     ),
                 )
+                self.save_state()
```
seanmreidy, line 127: This will be shallow I think. Any child migrations that are done within the asset won't be saved incrementally. Maybe adding save_state to base.py would be helpful

nominal/experimental/migration/parallel_migration_runner.py
```diff
@@ -47,6 +51,45 @@ def fn() -> None:
...
+
+    originals: dict[int, object] = {}
+
+    def _handler(signum: int, frame: FrameType | None) -> None:
+        logger.warning(
+            "Received signal %d — saving migration state to %s before exiting", signum, runner.migration_state_path
+        )
+        runner.save_state()
+        signal.signal(signum, originals[signum])  # type: ignore[arg-type]
+        signal.raise_signal(signum)
```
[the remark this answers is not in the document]
seanmreidy: Seems legit

nominal/experimental/migration/migration_runner.py
tarun-prakash-nmnl: Agreed — it was shallow. Fixed in 3b8bd91, with one twist on the suggestion: the hook lives on the state object rather than `Migrator.base`, because several migrators record mappings directly via `ctx.migration_state.record_mapping(...)` (e.g. data scopes, deferred workbooks, dataset files) and would bypass a base-class hook — every mutation funnels through the `MigrationState` mutators, so that's the true choke point.

Every mutator (`record_mapping`, pending-workbook records/clears, `record_skip`) now triggers a persist, **debounced to at most one write per second**: serialization is O(state size) and file-heavy assets record a mapping per dataset file, so saving on every mapping would be quadratic over a big migration. Net effect: a hard kill (SIGKILL/OOM) now loses at most ~1s of child mappings instead of the whole in-flight asset; all catchable exits still save unconditionally (per-task callback, signal flush, `finally`), so they lose nothing regardless of the debounce.

---
_Generated by [Claude Code](https://claude.ai/code)_

seanmreidy (approved):

PR #876, opened 2026-07-10
fix(migration): route mcap and dataflash through native ingest

ckniffin-nominal:
Routes migrated MCAP and ArduPilot Dataflash files through `Dataset.add_mcap` and `Dataset.add_ardupilot_dataflash` instead of generic tabular ingestion. The source object is streamed to a temporary file so native ingestion receives the original filename and file tags, then the temporary file is cleaned up after ingestion completes.

Adds explicit `FileType` predicates for MCAP and Dataflash while reusing the canonical `FileTypes.DATAFLASH` value. Journal, MCAP, and Dataflash migration now share the same temporary-file ingestion helper.

Tests cover native routing, filename/tag preservation, polling, and cleanup.

Linear: https://linear.app/nominal-io/issue/CON-2757/route-mcap-and-dataflash-migration-through-native-ingest

Link to Devin session: https://app.devin.ai/sessions/20d77beca15049d681c806f0b41c66d3
Requested by: @ckniffin-nominal

### DIFF
--- a/nominal/experimental/migration/utils/file_utils.py
+++ b/nominal/experimental/migration/utils/file_utils.py
@@ -9,7 +9,7 @@ from typing import BinaryIO, cast
 
 import requests
 
-from nominal.core import Dataset, DatasetFile, FileType, FileTypes
+from nominal.core import Dataset, DatasetFile, FileType
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


nominal/experimental/migration/utils/file_utils.py
```diff
@@ -67,6 +74,27 @@ def copy_file_to_dataset(
...
+    with tempfile.TemporaryDirectory() as tmp_dir:
+        tmp_path = Path(tmp_dir) / file_name
+        with response:
+            with tmp_path.open("wb") as tmp:
+                shutil.copyfileobj(response.raw, tmp)
+        new_file = ingest_fn(tmp_path)
+        new_file.poll_until_ingestion_completed()
+        return new_file
+
+
```
drake-nominal, line 97: We can do this wholly in memory as well using io streams and the whole `_from_io` methods instead of needing to even involve disks, even preserving input filenames.

seanmreidy: +1 to this

drake-nominal: In fact, what is now the tabular path _does_ use this mechanism. The only reason the journal json path doesn't do this is likely because we haven't added a method for it yet. Ideally instead we do:

* add `add_journal_json_from_io`
* have one single helper that basically does what `_ingest_from_temp_file` does, but using io pipes

PR #878, opened 2026-07-11
ci(e2e): file-upload special-char e2e tests on S3 + Azure Blob

seanmreidy:
Adds a `file-upload-e2e` matrix job that runs the special-character filename tests against each object-store backend — **aws** (gov/S3) and **azure** (Blob).

## What it covers

`test_special_char_filename_round_trips` uploads a file named with **every special character the client allows** and asserts it ingests and round-trips its stored name — run for real against **both** backends. This is a cross-backend **drift canary**: unit tests mock the upload client, so they can only verify our client-side logic; only a real upload can catch a character that passes client-side validation but fails **server-side** (in scout or the cloud storage layer). If AWS, Azure, or the backend ever starts rejecting or mangling an allowed character, the matrix leg for that backend goes red immediately.

The unsafe characters (`/ \ ? % { } '`) are rejected client-side *before any upload*, so they keep a single smoke check — e2e can't observe backend behavior for them since they never leave the client.

## Scope & rationale

**Intentionally narrow** (only the filename tests): the broader Azure staging suite has unrelated flaky failures today (e.g. a file-download test), which would make a full-suite Azure job red for reasons unrelated to this change. This is a stopgap — **once a general Azure staging e2e suite lands, it will run these checks and this dedicated file-upload job will be folded into / replaced by it.**

Each matrix leg **fails loudly** if its secret is missing (no silent skips). AWS reuses the existing staging secrets; Azure uses `E2E_NOMINAL_AZURE_STAGING_API_KEY` / `E2E_NOMINAL_AZURE_STAGING_WORKSPACE_RID`.

## Note on CI

This workflow only triggers on PRs targeting `main`, so it won't run on this stacked PR until it's retargeted to `main` (after the base PR merges). The tests were verified locally against both backends (**14 passed each**: 13 allowed characters round-trip + 1 unsafe character rejected).

Stacked on #877.

### DIFF
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -1,5 +1,7 @@
 # End-to-end tests against Nominal staging.
 # Requires the E2E_NOMINAL_STAGING_API_KEY repository secret.
+# The file-upload-e2e Azure leg additionally requires E2E_NOMINAL_AZURE_STAGING_API_KEY /
+# E2E_NOMINAL_AZURE_STAGING_WORKSPACE_RID.
 
 name: E2E Tests (Staging)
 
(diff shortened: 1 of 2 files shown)


alkasm (approved):

seanmreidy:
Superseded by #885, which merged to `main` as `e94173b`. This PR auto-closed when its base branch (`fix/upload-filename-encoding`, #877) was deleted on merge; #885 is the same change re-cut cleanly against `main`.

PR #883, opened 2026-07-14
feat: experimental derived series

adithya-nominal:
Supports derived series creation/modification as an experimental feature.

Tested on a staging environment. Confirmed that creation/retrieval works as expected and that data is derived correctly.


nominal/experimental/compute/README.md
```diff
@@ -96,6 +96,12 @@ In addition to the Python-internal DSL above, this library can execute expressio
 pip install 'nominal[compute]'
 ```
 
+### Computing a series
+
+`compute_series` applies a `nominal_compute` expression to concrete channels and returns the computed values as a
+`pandas.Series` indexed by timestamp. References in the expression are bound to channels at execution time via
+`inputs`—nothing is persisted back to Nominal:
```
david-nominal: nit: the ':' seems missplaced

nominal/experimental/compute_as_code/_series.py
```diff
@@ -0,0 +1,127 @@
...
+    request = _build_export_request(
+        compute_node=_to_conjure_series(expr),
+        column_name=name,
+        inputs=inputs,
+        start=start,
+        end=end,
+        enable_gzip=enable_gzip,
+    )
+    response = client._clients.dataexport.export_channel_data(client._clients.auth_header, request)
+    df = pd.read_csv(
```
david-nominal, line 82: how does this work with tags?

adithya-nominal: We didn't need tags for our impl use cases but good call, will add that in 🫡. Pushing changes in a few.

david-nominal (approved):

nominal/experimental/compute_as_code/_series.py
```diff
@@ -0,0 +1,136 @@
...
+    import nominal_compute
+
+_FlexibleTimestamp = str | datetime | ts.IntegralNanosecondsUTC
+
+
+def _to_conjure_series(
+    series: nominal_compute.NumericSeries | nominal_compute.CategoricalSeries,
+) -> scout_compute_api.Series:
+    """Convert a ``nominal_compute`` series expression into the ``scout_compute_api.Series`` the compute API expects."""
+    wire_json = series.to_json()  # type: ignore[union-attr]
```
alkasm, line 29: this ignore shouldn't be necessary

nominal/experimental/compute_as_code/_series.py
```diff
@@ -0,0 +1,136 @@
...
+"""
+
+from __future__ import annotations
+
+import json
+from datetime import datetime
+from typing import TYPE_CHECKING, Any, BinaryIO, Mapping, cast
+
+import pandas as pd
+from conjure_python_client._serde.decoder import ConjureDecoder
```
alkasm: why are we importing from an internal path?

nominal/experimental/compute_as_code/_series.py
```diff
@@ -0,0 +1,136 @@
...
+
+_FlexibleTimestamp = str | datetime | ts.IntegralNanosecondsUTC
+
+
+def _to_conjure_series(
+    series: nominal_compute.NumericSeries | nominal_compute.CategoricalSeries,
+) -> scout_compute_api.Series:
+    """Convert a ``nominal_compute`` series expression into the ``scout_compute_api.Series`` the compute API expects."""
+    wire_json = series.to_json()  # type: ignore[union-attr]
+    if type(series).__name__ == "NumericSeries":
```
alkasm: use an `isinstance` check

nominal/experimental/compute_as_code/_series.py
```diff
@@ -0,0 +1,136 @@
...
+
+
+def compute_series(
+    client: NominalClient,
+    expr: nominal_compute.NumericSeries | nominal_compute.CategoricalSeries,
+    inputs: Mapping[str, Channel],
+    start: _FlexibleTimestamp | None = None,
+    end: _FlexibleTimestamp | None = None,
+    *,
+    tags: Mapping[str, Mapping[str, str]] | None = None,
```
alkasm: wdyt about `tags_by_reference_name` or something a bit more self-documenting? or maybe updating the docstring is sufficient? either way i dont find the docstring here super clear on its own

nominal/experimental/compute/README.md
```diff
@@ -96,6 +96,38 @@ In addition to the Python-internal DSL above, this library can execute expressio
...
+
+from nominal.core import NominalClient
+from nominal.experimental.compute import compute_series
+
+client = NominalClient.from_profile("...")
+dataset = client.get_dataset("...")
+
+# Author an expression over named references, then bind each reference to a concrete channel.
+expr = nc.NumericSeries.Reference("a") - nc.NumericSeries.Reference("b")
+series = compute_series(client, expr, inputs={"a": dataset.get_channel("a"), "b": dataset.get_channel("b")})
```
alkasm: nit but can you just add a more realistic example here since the three repeated strings otherwise make this a bit more confusing than it realistically is

nominal/experimental/compute_as_code/_series.py
adithya-nominal: Removing it fails mypy checks: 
```
nominal/experimental/compute_as_code/_series.py:29: error: Item "NumericSeries" of "NumericSeries | CategoricalSeries" has no attribute "to_json"  [union-attr]
nominal/experimental/compute_as_code/_series.py:29: error: Item "CategoricalSeries" of "NumericSeries | CategoricalSeries" has no attribute "to_json"  [union-attr]
```

Could change nominal-compute's stub generation to add to_json but otherwise I do think we need it.

alkasm (approved):

PR #896, opened 2026-07-28
feat(video): add dataset video ingest and typed video dataset files

varun-nominal:
Adds first-class support for video stored as a **channel on a dataset**, so video is uploaded and managed the same way tabular and MCAP data already are:

```python
video = dataset.add_video("front-camera.mp4", channel="camera/front", start=recording_start)
video.poll_until_ingestion_completed()
video.update(starting_timestamp=corrected_start)   # fix the alignment after the fact

for file in dataset.list_video_files():
    print(file.name, file.num_frames, file.media_frame_rate)
```

### What's included

**`VideoDatasetFile`** — a `DatasetFile` subtype for video, exposing the aggregates segmentation produces (`num_frames`, `num_segments`, `media_duration_seconds`, `media_frame_rate`, `scale_factor`). It inherits everything a dataset file already supports: `refresh()`, `download()`, `delete()`, `poll_until_ingestion_completed()`, tags, bounds, ingest status.

**Uploads** — `Dataset.add_video` / `add_video_from_io` (start timestamp or per-frame timestamps) and `add_mcap_video` / `add_mcap_video_from_io` (timestamps from an MCAP topic).

**Retrieval** — `Dataset.list_video_files()` and `get_video_file()`. Video rows are also specialized on the *generic* paths (`list_files`, `get_dataset_file`, `search_files`, `IngestionJob.dataset_files()`), so a video file comes back as `VideoDatasetFile` everywhere; narrow with `isinstance`. The subtype survives `refresh()` and polling, so the aggregates populate as segmentation completes.

**`update()`** — change a file's name, start timestamp, or timing scale (exactly one of `ending_timestamp` / `true_frame_rate` / `scale_factor`, expressed via `typing.overload`), mirroring the legacy `VideoFile.update` signature so migrating callers don't relearn it. Updates are addressed by the file's own `dataset_rid` (backend support shipped in scout#17688; pins bumped to `nominal-api` 0.1352.0), so `update()` works on any `VideoDatasetFile` handle — including files read back from the catalog — with no channel bookkeeping in the client.

**`ChannelDataType.VIDEO`** — video channels previously surfaced as `UNKNOWN` through `search_channels` / `get_channels`, since the enum had no member for a type the backend already reports.

### Not included

- No deprecations of `Video` / `VideoFile` / `create_video`. Legacy video is untouched and keeps working; this is purely additive.
- No per-file `description` (video channels don't carry one) and no per-file archive.
- Asset-backed video channels can't be updated — the backend supports datasource-backed channels only for these mutations.

### Testing plan

Unit tests across `tests/core/test_dataset.py`, `test_video_dataset_file.py`, and `test_video_ingest.py`, covering: subtype dispatch over real catalog rows (video vs plain), aggregate extraction with and without segment metadata, timestamp-mode exclusivity and text-stream rejection (parametrized over the entry points), MCAP extension rejection, the ingest-response resolution paths (direct file id, job-lookup fallback, and the zero/multiple failure cases), one real-request test asserting the exact `IngestRequest` that `add_video_from_io` submits, and `update()` validation and dataset-RID addressing.

Full suite passes (584), `mypy` and `ruff` clean.

**Known gap:** coverage is unit-level only. The equivalent legacy methods get real upload→segmentation coverage from `tests/e2e/`; this surface has no e2e counterpart yet, so the upload → segmentation → aggregates-populate loop is unproven against a live instance.


nominal/core/video_dataset_file.py
```diff
@@ -0,0 +1,134 @@
...
+                that for every second of media, two seconds have elapsed in absolute time.
+
+        Returns:
+            This file, refreshed with its new bounds and recomputed segment metadata.
+
+        Raises:
+            ValueError: no field was provided, more than one of {ending_timestamp, true_frame_rate,
+                scale_factor} was provided, or this file's video channel could not be resolved.
+
+        NOTE: only one of {ending_timestamp, true_frame_rate, scale_factor} may be present at one time.
```
drake-nominal, line 96: Can we use `typing.overload` to express this?

nominal/core/video_dataset_file.py
```diff
@@ -0,0 +1,134 @@
...
+        NOTE: video channels do not carry per-file descriptions, so unlike the legacy
+            `VideoFile.update` there is no `description` parameter.
+        """
+        scale_inputs = (ending_timestamp, true_frame_rate, scale_factor)
+        if sum(value is not None for value in scale_inputs) > 1:
```
drake-nominal: sum of bools is a little wonky but i guess it does work.

nominal/core/video.py
```diff
@@ -455,24 +455,34 @@ def _build_video_file_timestamp_manifest(
...
+    if start is not None:
         # TODO(drake): expose scale parameter to users
         return scout_video_api.VideoFileTimestampManifest(
             no_manifest=scout_video_api.NoTimestampManifest(
                 starting_timestamp=_SecondsNanos.from_flexible(start).to_api()
             )
         )
-    else:
-        raise ValueError("One of 'start' or 'frame_timestamps' must be provided")
+    raise AssertionError("unreachable: exactly one timestamp mode was validated above")
```
drake-nominal: instead of raising an assertion error, we should raise a normal value error or other standard exception.

nominal/core/video.py
```diff
@@ -455,24 +455,34 @@ def _build_video_file_timestamp_manifest(
...
-    elif start is not None:
+    if start is not None:
         # TODO(drake): expose scale parameter to users
         return scout_video_api.VideoFileTimestampManifest(
             no_manifest=scout_video_api.NoTimestampManifest(
                 starting_timestamp=_SecondsNanos.from_flexible(start).to_api()
             )
         )
-    else:
-        raise ValueError("One of 'start' or 'frame_timestamps' must be provided")
```
drake-nominal, line 475: I think the elif else structure was preferable

nominal/core/video.py
```diff
@@ -455,24 +455,34 @@ def _build_video_file_timestamp_manifest(
     upload_client: upload_api.UploadService,
     start: datetime | IntegralNanosecondsUTC | None = None,
     frame_timestamps: Sequence[IntegralNanosecondsUTC] | None = None,
+    mcap_topic: str | None = None,
     header_provider: HeaderProvider | None = None,
```
drake-nominal, line 459: We should use `typing.override` ideally to express that it can only be one or the other, but maybe overkill because this is a helper. Perhaps it should be expressed in `update()`.

nominal/core/dataset_file.py
```diff
@@ -252,51 +258,69 @@ def get_file_size(self) -> int | None:
...
-        timestamp_type = None
-        if dataset_file.timestamp_metadata is not None:
-            timestamp_column = dataset_file.timestamp_metadata.series_name
-            timestamp_type = _catalog_timestamp_type_to_typed_timestamp_type(
-                dataset_file.timestamp_metadata.timestamp_type
-            )
+        return cls(**_parse_common_file_fields(clients, dataset_file))
+
+
+def _parse_common_file_fields(clients: DatasetFile._Clients, dataset_file: scout_catalog.DatasetFile) -> dict[str, Any]:
```
drake-nominal: this is a bit weird of a method to make things DRY. I like how it does the internals once, however, I really do think it would be cleaner to just route to each individual methods. 

One hacky idea to keep things DRY AND typesafe-- use `DatasetFile._from_conjure` in `VideoDatasetFile._from_conjure` and then pull typed fields off of that when constructing a VideoDatasetFile.

nominal/core/_video_ingest.py
```diff
@@ -0,0 +1,138 @@
...
+    the channel (one series is created per originating ingest job).
+    """
+    return scout_video_api.VideoChannelSeries(
+        data_source=scout_video_api.VideoDataSourceChannel(
+            data_source_rid=dataset_rid,
+            channel=channel,
+            tags={**(tags or {})},
+        )
+    )
+
```
drake-nominal: some of these helpers aren't really buying a ton, just adding layers but not really simplifying plumbing.

nominal/core/_video_ingest.py
```diff
@@ -0,0 +1,138 @@
...
+                page_size=None,
+                prefix=None,
+            ),
+        )
+        for channel_metadata in response.results:
+            if channel_metadata.data_type == api.SeriesDataType.VIDEO:
+                yield channel_metadata.name
+        if response.next_page_token is None:
+            break
+        next_page_token = response.next_page_token
```
drake-nominal: This should use our classic search machinery vs. having paging etc. implemented separately.

nominal/core/_video_ingest.py
```diff
@@ -0,0 +1,138 @@
...
+            end=_SecondsNanos.from_nanoseconds(bounds.end).to_api(),
+        )
+    )
+    for channel in iter_video_channel_names(clients, dataset_rid):
+        if _channel_contains_file(clients, dataset_rid, channel, dataset_file_id, probe_bounds):
+            return channel
+    raise ValueError(
+        f"could not determine the video channel for dataset file {dataset_file_id!r} on dataset "
+        f"{dataset_rid!r}: no video channel on the dataset contains it"
+    )
```
drake-nominal: this is super unperformant & will make a ton of sequential (rate limited) network requests.

First, search all channels. Then for each channel individually, list the files on that channel, then look to see if that file matches any of those files. 

Instead, this should be stored / associated in the backend & we shouldn't expose functionality that needs the channel of a video file yet.

nominal/core/dataset.py
```diff
@@ -130,6 +146,33 @@ def _handle_ingest_response(self, response: ingest_api.IngestResponse) -> Datase
...
+            # The backend writes `metadata.video` when it creates the dataset-file row, before segmentation is
+            # dispatched — so a video ingest's row is video-typed from the moment it exists. A non-video result
+            # here means the id points at a genuinely non-video file, not a not-yet-populated one.
+            raw = self._clients.catalog.get_dataset_file(
+                self._clients.auth_header, details.dataset_rid, details.dataset_file_id
+            )
+            file = _dataset_file_from_conjure(self._clients, raw)
+            if not isinstance(file, VideoDatasetFile):
+                raise NominalIngestError(f"ingested file {details.dataset_file_id!r} is not a video dataset file")
+            return file
```
drake-nominal: If we flip this it is more defensive against future modalities (e.g. SpatialDatasetFile). I think we should do `if isinstance(file, VideoDatasetFile):... else: raise`

nominal/core/dataset.py
```diff
@@ -457,6 +500,300 @@ def add_mcap_from_io(
...
+        video: BinaryIO,
+        name: str,
+        *,
+        channel: str,
+        start: datetime | IntegralNanosecondsUTC | None = None,
+        frame_timestamps: Sequence[IntegralNanosecondsUTC] | None = None,
+        file_type: tuple[str, str] | FileType = FileTypes.MP4,
+        tags: Mapping[str, str] | None = None,
+        overwrite_overlapping: bool = False,
+    ) -> VideoDatasetFile:
```
drake-nominal, line 608: we use `file_name` as the name of the argument elsewhere in `dataset.py` vs. `name`-- can we update on the two from io methods?

drake-nominal:
## Test review: 37 added tests → ~18

The behavioral coverage here is good, but a fair chunk of the suite asserts the call graph rather than behavior. Proposed cuts, grouped by reason.

### Delete outright

- **`test_ingest_options_builds_video_v2`** and **`test_ingest_options_none_tags_becomes_empty_and_no_overwrite_is_none`** (`test_video_ingest.py`) — `build_video_ingest_options` is a pure constructor; every assert restates a line of `_video_ingest.py`. Trivial 1:1 conversion. The one non-obvious bit (`overwrite_overlapping or None` collapsing `False → None`) belongs on the real ingest test, not on a dedicated test of a formatter.
- **`test_update_maps_each_scale_input_to_its_union_arm`** — uses `getattr(scale, arm)`, and is a bare `for` loop rather than `@pytest.mark.parametrize`, so a failure won't say which arm broke. What it covers is a three-branch `if/elif` mapping a kwarg to a same-named union field.
- **`test_timestamp_manifest_excluded_from_repr_and_equality`** — asserts that `dataclasses.field(repr=False, compare=False)` does what the stdlib documents.

### Delete, but only alongside one replacement

These follow `ds = MagicMock(); Dataset.method(ds, ...); assert ds.other_method.called`. The system under test is a mock, the assertion transcribes the call site, and any pure refactor (inlining `_ingest_video`, say) breaks them with no behavior change.

- `test_add_video_forwards_start_mode`
- `test_add_video_forwards_frame_timestamps_mode`
- `test_add_video_from_io_delegates_to_ingest_video`
- `test_add_mcap_video_from_io_delegates_with_mcap_topic`
- `test_get_video_file_returns_video_subtype` — two-line method; asserts the mock returns the mock.
- `test_ingest_video_uploads_and_submits_video_v2` — `build_opts.assert_called_once_with(...)` is a verbatim copy of the call site, and it patches away `build_video_ingest_options`, the one object worth inspecting.
- `test_list_files_specializes_video_rows` — the comment apologizing for the mock setup ("patching `Dataset._list_files` has no effect here") is the tell that the test is fighting its own scaffolding; it verifies a one-word source diff already covered by the dispatch tests.

**Replacement:** one test driving a *real* `Dataset` (the `mock_dataset` fixture already in `test_dataset.py`) through `add_video_from_io`, asserting the actual `ingest_api.IngestRequest` handed to `clients.ingest.ingest` — same convention as the neighboring `test_write_logs_*`. That covers the manifest arm, options mapping, s3 path, and the whole delegation chain in one test, and survives refactors. Without it these deletions do lose coverage.

### Collapse rather than delete

- The four timestamp-mode validation tests (`test_add_video_{,from_io_}requires_a_timestamp_mode` / `..._rejects_both_timestamp_modes`) → one parametrized test over (entry point × bad args). `test_manifest_requires_exactly_one_mode` is a third copy of the same invariant but is the load-bearing one — keep it.
- `test_add_video_from_io_rejects_text_stream` + `test_add_mcap_video_from_io_rejects_text_stream` → parametrize over the two entry points.
- The three `_dataset_file_from_conjure` dispatch tests → one parametrized test over real `scout_catalog.DatasetFile` beans asserting `isinstance(result, VideoDatasetFile)` vs `type(result) is DatasetFile`, instead of patching both factories and comparing `"video-file"` / `"base-file"` sentinels. Same branch coverage, no patching, and it would catch a broken `_from_conjure`.
- `test_from_conjure_populates_aggregates_from_segment_metadata` + `test_from_conjure_leaves_aggregates_none_without_segment_metadata` → one parametrized test. The body under test is `None if segment is None else segment.X` five times, and both patch `DatasetFile._from_conjure` to return a hand-built base instance, coupling them to the subclass factory's field-by-field construction order.

### Keep

All five `test_handle_video_response_*` (real branch and error paths in `_resolve_ingested_video_file`), `test_add_mcap_video_rejects_non_mcap_path`, `test_get_video_file_raises_type_error_for_non_video`, the three `update` validation/addressing tests, and the four `test_manifest_*` tests. The `update` and `manifest` ones inspect the real request/union object and assert no call was made on invalid paths — that's the shape the rest should take.

### Nits

- The five `test_handle_video_response_*` names are stale; the method is `_resolve_ingested_video_file`.
- `test_update_addresses_the_files_dataset` asserts `request.channel_series is None`, an unset optional the source never touches.

nominal/core/_video_ingest.py
```diff
@@ -0,0 +1,29 @@
...
+def build_video_ingest_options(
+    target_rid: str,
+    *,
+    channel: str,
+    tags: Mapping[str, str] | None,
+    s3_path: str,
+    timestamp_manifest: scout_video_api.VideoFileTimestampManifest,
+    overwrite_overlapping: bool,
+) -> ingest_api.IngestOptions:
+    """Build IngestOptions for a VideoOptsV2 ingest into an existing dataset channel."""
```
drake-nominal: we should probably just make this a private method in dataset.py

drake-nominal (approved):
approve with one nit

PR #898, opened 2026-07-29
feat(experimental): two-layer upload backoff, tuned defaults, production hardening

drake-nominal:
Stacked on #848 — everything below the final commit is that PR. The reviewable delta here is the single commit on top (`c01d825`).

## What this adds on top of #848

**Two-layer throttle backoff** replaces the adaptive rate pacer in the experimental uploader:
- **Local layer**: the shared conjure client's own short jittered transport retries absorb isolated throttles. The uploader's dedicated upload client is gone — it consumes `clients.upload` again, and `upload_client=` remains as an injection seam.
- **Global layer**: a fixed-width admission lane bounds how many requests may touch the Nominal API concurrently (structural pressure bound, no rate estimation), plus one shared storm damper that doubles on sustained throttling and decays on success, with full jitter so retries don't re-herd.
- Deadline semantics preserved: lane-admission waits and backoff sleeps spend the same per-request budget as the request itself; a backoff sleep holds no lane ticket; short-deadline callers (the abort path) fail fast behind a stormed lane.

**Benchmark-tuned defaults**
- `max_workers=10` concurrent PUT streams — aggregate throughput rises with streams only until the network path saturates; past that, additional streams degrade every stream.
- Driver pool (files in flight) defaults to half the part pool.
- Nominal admission lane of 5 — fills the server's admission budget without provoking refusals; wider lanes finished many-small-file batches no sooner while re-sending a large fraction of request bodies.
- Single-shot small-file route **on by default at 1 MiB** (`small_file_route_max_bytes=None` disables it; 0-byte files always take multipart).

**Hardening from review**
- A cancelling close now settles *and waiter-notifies* every future `enqueue_file` ever returned before it returns, so `concurrent.futures.wait`/`as_completed` on dropped files can never hang; future issuance is lock-serialized against close, which also turns the enqueue-vs-shutdown race into a clean `RuntimeError`.
- `IngestBuilder.submit()` is single-use — success or failure — and a second call raises instead of re-uploading and double-ingesting; documented on the class and method.
- DEBUG logging across the upload lifecycle: resolved pool config, per-file route decisions, per-file completion timings, throttle events with the damper's current delay, and close.

**Cleanup**
- The benchmark instrumentation module is removed, and the core conjure-transport status-retry/pool parameterization it existed to exercise is reverted — relative to #848, this PR no longer touches `core/_utils/networking.py` at all.

## Testing

78 tests in `tests/experimental/ingest/` (injected fake clocks, event rendezvous — no real sleeps), including regression locks for the waitable-after-cancelling-close guarantee, single-use submit on both success and failure paths, and the small-file route's default/at-threshold/0-byte/disabled edges. `ruff` clean; `mypy` unchanged.


nominal/experimental/ingest/_multipart_uploader.py
```diff
@@ -0,0 +1,581 @@
...
+)
+
+logger = logging.getLogger(__name__)
+
+# Files at or below the uploader's `small_file_route_max_bytes` are uploaded single-shot via the
+# backend `upload_file` endpoint (one request) instead of the multipart flow. That endpoint holds
+# a server request thread for the whole transfer, so the opt-in threshold is hard-capped: measured
+# throughput of the two routes reaches parity around this size, above which multipart is strictly
+# better (its bytes stream directly to storage and its parts parallelize). Experimental.
+MAX_SMALL_FILE_ROUTE_BYTES = 4 * 1024 * 1024  # 4 MiB
```
drake-nominal: this default should be 1MiB, not 4Mib, for small files

nominal/experimental/ingest/_multipart_uploader.py
```diff
@@ -0,0 +1,581 @@
...
+
+# Default single-shot threshold. Below this size the route's single request beats the multipart
+# flow's three; comfortably inside the hard ceiling above so there is headroom to raise it.
+DEFAULT_SMALL_FILE_ROUTE_MAX_BYTES = 1024 * 1024  # 1 MiB
+
+# Benchmark-tuned part-pool default. Aggregate upload throughput rises with concurrent PUT
+# streams only until the network path saturates; past that point more streams actively degrade
+# every stream (they congest a shared queue and recover slowly from the synchronized losses),
+# so the default is a moderate fixed width rather than something scaled to CPU count.
+DEFAULT_MAX_WORKERS = 10
```
drake-nominal: The comment here should be less "you _will_ see degradations" and more of a cautious "raising this may degrade performance-- here's the tangible scenario in which you should raise this".

nominal/experimental/ingest/_multipart_uploader.py
```diff
@@ -0,0 +1,581 @@
...
+    # a future appended under it is guaranteed to be covered by close's cancel pass.
+    _issued_futures: list[Future[str]] = field(default_factory=list, repr=False)
+    _issue_lock: threading.Lock = field(default_factory=threading.Lock, repr=False)
+
+    @classmethod
+    def create(
+        cls,
+        clients: Any,
+        *,
+        workspace_rid: str | None = None,
```
drake-nominal: we should create one with a `NominalClient`. We should use the configured workspace_rid from that client.

Let's make the names more descriptive:

* `max_storage_workers`
* `max_small_file_workers`
* `max_files_in_flight`
* `max_nominal_concurrency`
* `per_request_retry_timeout`
* `max_backoff_duration`

We shouldn't need to take in a separate `upload_client` in `create`-- we should take in the `NominalClient` and do the rest ourselves w.r.t. setup.

nominal/experimental/ingest/_multipart_uploader.py
```diff
@@ -0,0 +1,581 @@
...
+                # upload open server-side).
+                for future in self._issued_futures:
+                    future.cancel()  # queued files drop; running or settled ones are unaffected
+                self._driver_pool.shutdown(wait=True)
+                self._small_pool.shutdown(wait=True)
+                self._part_pool.shutdown(wait=True)
+            else:
+                self._driver_pool.shutdown(wait=True)  # drivers submit parts: part pool must outlive them
+                self._part_pool.shutdown(wait=True)
+                self._small_pool.shutdown(wait=True)
```
drake-nominal: couldn't we have a if block pre-amble that cancels futures / does the non-waiting shutdown, and then both branches can share the waiting shutdown code.

nominal/experimental/ingest/_multipart_uploader.py
```diff
@@ -0,0 +1,581 @@
...
+            )
+            logger.debug(
+                "completed multipart upload for %s (%d bytes, %d parts) in %.2fs",
+                safe_filename,
+                pending.total_size,
+                len(etags),
+                time.monotonic() - started,
+            )
+            return location
+        except BaseException as e:
```
drake-nominal, line 494: can this just `except Exception`?

nominal/experimental/ingest/_upload_pacing.py
```diff
@@ -0,0 +1,206 @@
...
+def _is_throttle_error(exc: BaseException) -> bool:
+    """True if `exc` is the server refusing the request because the caller is over budget.
+
+    `RetryError` is the primary signal: the transport's own short retry ladder was exhausted,
+    meaning the throttling is sustained rather than momentary. Raw 429/503 are kept for
+    injected transports that do not retry statuses locally.
+    """
+    if isinstance(exc, requests.exceptions.RetryError):
+        return True
+    status = getattr(exc, "status_code", None)
```
drake-nominal: never use getattr

nominal/experimental/ingest/_upload_pacing.py
```diff
@@ -0,0 +1,206 @@
...
+
+    def __init__(
+        self,
+        *,
+        max_concurrency: int = NOMINAL_MAX_CONCURRENCY,
+        deadline_seconds: float = DEFAULT_THROTTLE_DEADLINE_S,
+        backoff: _GlobalBackoff | None = None,
+        clock: Callable[[], float] = time.monotonic,
+        sleep: Callable[[float], None] = time.sleep,
+        jitter: Callable[[float], float] | None = None,
```
drake-nominal: I can't help but wonder if default values for a private class like this make sense / if it makes it more likely for drift to happen. I also can't help but feel like we added a bunch of unecessary initializer arguments for the sake of testing.

nominal/experimental/ingest/__init__.py
```diff
@@ -0,0 +1,7 @@
+"""Experimental multi-file ingest via a single MULTI ingest job (see `IngestBuilder`)."""
+
+from __future__ import annotations
+
+from nominal.experimental.ingest._ingest_builder import IngestBuilder
+
+__all__ = ["IngestBuilder"]
```
drake-nominal: Let's expose multipart builder too

nominal/experimental/ingest/_multipart_uploader.py
drake-nominal: The 1 MiB default landed as `DEFAULT_SMALL_FILE_ROUTE_MAX_BYTES` (a few lines below) — this 4 MiB constant is the hard *ceiling* on the `small_file_route_max_bytes` knob, not the default. e06e9fe reorders the two so the default reads first and relabels both comments to make DEFAULT vs CEILING unambiguous.

nominal/experimental/ingest/_multipart_uploader.py
drake-nominal: Rephrased in e06e9fe as cautious guidance: raising it *may* degrade throughput past the path's saturation point, and the tangible raise-it scenario is spelled out — a measured benchmark on a high-bandwidth egress showing N streams sustaining more aggregate throughput than 10. The create() docstring got the same treatment.

nominal/experimental/ingest/_multipart_uploader.py
drake-nominal: Done in e06e9fe: `create(client: NominalClient, ...)` derives the upload client, auth, headers, and the client's configured workspace itself — the separate `upload_client` and `workspace_rid` params are gone (tests inject through the client object now). Knobs renamed as listed: `max_storage_workers`, `max_small_file_workers`, `max_files_in_flight`, `max_nominal_concurrency`, `per_request_retry_timeout` (was throttle_deadline), and `max_backoff_duration` is now a real knob capping the storm damper's delay (was a module-private constant).

tests/experimental/ingest/test_multipart_uploader.py
```diff
@@ -0,0 +1,746 @@
...
+        # a queued file can be cancelled and never touches the network
+        assert futures[-1].cancel() is True
+        session.put_release.set()
+        for fut in futures[:-1]:
+            fut.result(timeout=10)
+        with pytest.raises(CancelledError):
+            futures[-1].result(timeout=1)
+        up.close()
+        assert service.calls.count("initiate") == 9  # the cancelled file never initiated
+
```
drake-nominal, line 293: Testing has a few conventions in this repository:

* all tests should be testing behaviors, not implementations. Don't add tests for the sake of test coverage
* all tests should liberally use `pytest.fixtures` to encapsulate their setup. Extra bonus points if a fixture is used that already exists, or if fixtures are shared across tests to keep things more DRY
* Tests should have a one sentence docstring of the behavior under test
* Prefer magicmock to monkeypatch
* use static typehints where possible
* Build tests to test end to end visible behaviors consistent with the end user visible behavior, not undocumented internal features.

nominal/experimental/ingest/_ingest_builder.py
```diff
@@ -0,0 +1,552 @@
...
+    file_type: FileType
+    target: common_pb2.IngestSource
+
+
+@dataclass(frozen=True)
+class _PendingItem:
+    """A built ingest item and the uploads whose sources it is still waiting on (1 for most kinds)."""
+
+    item: ingest_service_pb2.IngestItem
+    uploads: tuple[_Upload, ...]
```
drake-nominal: Why couldn't we just store `IngestItem` in lists / dictionaries and avoid needing to maintain and store two separate dataclasses?

nominal/experimental/ingest/_ingest_builder.py
```diff
@@ -0,0 +1,552 @@
...
+        file_type = FileType.from_path(path)
+        if file_type is not FileTypes.AVRO_STREAM:
+            raise ValueError(
+                f"Cannot add path '{file_path}' as avro stream: inferred file type {file_type} not avro stream!"
+            )
+
+        # Avro deliberately omits channel_name_overrides: the backend rejects it for avro
+        # ("channel names come from record data"). TODO(drake): expose it once the backend accepts it.
+        options = file_ingest_pb2.FileIngestOptions(
+            timestamp_metadata=_canonical_avro_timestamp_metadata(),
```
drake-nominal, line 276: why is this not user overridable? The TODO should be on its own line

nominal/experimental/ingest/_ingest_builder.py
```diff
@@ -0,0 +1,552 @@
...
+        Returns:
+            The created ingest job. Track it by polling `job.refresh().status`, or block on its
+            produced files with `list(job.as_files_ingested())`.
+
+        Raises:
+            RuntimeError: this builder was already submitted.
+            ValueError: if no files have been added.
+        """
+        if self._submitted:
+            raise RuntimeError(
```
drake-nominal: can we raise a nominal specific error here

nominal/experimental/ingest/_multipart_uploader.py
drake-nominal: Done in 0caedfe — the cancel path is now a preamble (drain flag, non-waiting shutdowns, cancel pass) and both paths share one set of waiting joins. Join order unified to driver → part → small, which is safe in both modes: drivers wait on their own parts, so the part join is trivial by then, and the small pool has no ordering dependency.

nominal/experimental/ingest/_multipart_uploader.py
drake-nominal: I'd argue for keeping `except BaseException` here, with reasoning: this block never swallows — it cancels sibling parts, best-effort aborts the multipart upload server-side, and re-raises. Narrowing to `Exception` means a `SystemExit`/`KeyboardInterrupt` propagating through a driver would skip the abort and leak an open multipart upload (which costs storage until aborted). `CancelledError` is an `Exception` subclass in `concurrent.futures`, so it's caught either way — the only behavioral delta is the exotic-teardown case, where the broad catch is the safer side. Cleanup-and-reraise is the textbook justified use of `BaseException`. Happy to narrow it if you still prefer — in practice the difference is nearly nil.

nominal/experimental/ingest/_upload_pacing.py
drake-nominal: Done in 0caedfe — the classifier is now typed `isinstance` checks: `RetryError` first, then `requests.HTTPError` (verified: the conjure client's `ConjureHTTPError` subclasses it) with `.response.status_code in (429, 503)`. No getattr. The test fakes now raise real `HTTPError`s with real `Response` objects instead of duck-typed stand-ins.

nominal/experimental/ingest/_upload_pacing.py
drake-nominal: Agreed on the drift risk — 0caedfe removes the defaults from the tuning config (`max_concurrency`, `deadline_seconds` are now required; `create()` is the single owner of default values). On the injected `clock`/`sleep`/`jitter`: those keep real-clock defaults deliberately — they aren't tunable config that can drift (the default IS the production value), and they're what lets the throttle/backoff tests run deterministically with zero real sleeping, per the repo's no-sleep test convention. Without them the tests would have to monkeypatch `time.sleep`, which we also ban.

nominal/experimental/ingest/__init__.py
drake-nominal: Done in 0caedfe — `MultipartUploader` is exported from `nominal.experimental.ingest` alongside `IngestBuilder`.

tests/experimental/ingest/test_multipart_uploader.py
drake-nominal: Swept the suite to these conventions in 857da43: a shared `conftest.py` now provides `write_file` / `make_gate` / `install_test_gate` factory fixtures plus a single `FakeClock` (the second divergent copy is gone); `make_uploader` is a fixture that closes every uploader it built at teardown; the getattr/setattr throttling helper is replaced with explicit per-endpoint wrappers parametrized as callables; and every test now has a one-sentence behavior docstring and static type hints (incl. `WriteFile`/`MakeUploader` aliases). Assertions were already behavior-level (request counts, locations, settle states) and stay that way.

nominal/experimental/ingest/_ingest_builder.py
drake-nominal: Good catch — the per-item grouping was never actually consumed (submit always flattened it), so `_PendingItem` is deleted in 857da43. The builder now keeps two flat parallel lists: the `IngestItem`s to send and the `_Upload`s that must land first. `_Upload` itself stays as the one dataclass — it's the typed (path, file_type, proto-target) pairing that a bare dict/tuple would just re-encode stringly.

nominal/experimental/ingest/_ingest_builder.py
drake-nominal: It's not overridable because the backend rejects `channel_name_overrides` for avro items outright (channel names come from record data), so exposing the kwarg would fail every request server-side — same reason the v2 endpoint item-support matrix forbids it. The TODO now sits on its own line (857da43) so it reads as the follow-up it is: expose the kwarg the moment the backend accepts it.

nominal/experimental/ingest/_ingest_builder.py
drake-nominal: Done in 857da43 — a second `submit()` now raises `NominalIngestError` (reused from `nominal.core.exceptions` rather than minting a new public exception name for an experimental feature). Docstring and both single-use tests updated to match.

PR #908, opened 2026-08-04
feat: add public Workbook.create_template

duane-eldridge:
### Summary

Promotes the private `Workbook._create_template_from_workbook` to the public `Workbook.create_template`.

### Naming

The public name drops the `_from_workbook` suffix, which only repeats the receiver. The keyword arguments override metadata on the new template, never the source. The neighbors already read this way:

- `WorkbookTemplate.create_workbook` creates a workbook from a template.
- `Workbook.clone` creates a workbook from a workbook.

### Workspace parameter

The method now takes `workspace: Workspace | str | None` in place of a bare `workspace_rid` string. This matches `NominalClient.create_workbook_template`, the only other creator that targets a workspace. A caller holding a `Workspace` no longer has to reach for `.rid`. The type is worth settling now, because a public `workspace_rid: str` would cost a deprecation cycle to narrow later.

### Deprecated alias

`_create_template_from_workbook` stays as a deprecated alias that warns and delegates. It keeps the original `workspace_rid` parameter and maps it onto `workspace`, so the old call shape still works. Nothing in the repo calls it, and the name was never public, so deleting it outright is the alternative. The alias warns with `UserWarning`, because Python hides `DeprecationWarning` outside `__main__`.

### Docstring correction

The `workspace_rid` docstring promised the workspace of the current workbook. The code calls `resolve_default_workspace_rid()`, which returns the workspace of the client. That is the configured `workspace_rid` of a pinned client, or a lazily resolved default.

The method would also need another API request to find the workspace of the workbook. The `Workbook` class holds no workspace field, and neither does the `scout_notebook_api.Notebook` that it wraps. The neighbor `Workbook.clone` resolves the workspace the same way, so this PR corrects the text and leaves the behavior alone.

### Testing plan

- `ruff check` and `ruff format --check` pass.
- `mypy` reports no new errors. The 4 remaining errors are in `nominal/experimental/compute_as_code` and come from the optional `nominal_compute` module.
- `tests/core` passes with 326 passed and 12 skipped.
- New tests cover workspace resolution for a default, a RID, and a `Workspace`, and both conditions in the `Raises` block.
- One test covers the `workspace_rid` mapping in the alias, which `mypy` cannot check because the parameter is optional.

### DIFF
--- a/nominal/core/workbook.py
+++ b/nominal/core/workbook.py
@@ -9,9 +9,10 @@ from nominal_api import scout, scout_chartdefinition_api, scout_notebook_api, sc
 from typing_extensions import Self, deprecated
 
 from nominal.core._clientsbunch import HasScoutParams
-from nominal.core._utils.api_tools import HasRid, RefreshableConjureMixin
+from nominal.core._utils.api_tools import HasRid, RefreshableConjureMixin, rid_from_instance_or_string
 from nominal.core._utils.pagination_tools import search_workbooks_paginated
 from nominal.core._utils.query_tools import ArchiveStatusFilter, create_search_workbooks_query
+from nominal.core.workspace import Workspace
 
 logger = logging.getLogger(__name__)
 
@@ -266,7 +267,7 @@ class Workbook(HasRid, RefreshableConjureMixin[scout_notebook_api.Notebook]):
         description: str | None = None,
         labels: Sequence[str] | None = None,
         properties: Mapping[str, str] | None = None,
-        workspace_rid: str | None = None,
+        workspace: Workspace | str | None = None,
     ) -> WorkbookTemplate:
         """Create a workbook template from this workbook.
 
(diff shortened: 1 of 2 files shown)


nominal/core/workbook.py
```diff
@@ -310,6 +321,28 @@ def _create_template_from_workbook(
             commit_message="Initial version",
         )
 
+    @deprecated(
+        "`Workbook._create_template_from_workbook` is deprecated in favor of the public "
+        "`Workbook.create_template`, and will be removed in a future version.",
+        category=UserWarning,
+    )
```
drake-nominal, line 328: It's private, we don't need to deprecate it. use at your own risk!

nominal/core/workbook.py
```diff
@@ -259,14 +260,14 @@ def delete(self) -> None:
...
         self,
         *,
         title: str | None = None,
         description: str | None = None,
         labels: Sequence[str] | None = None,
         properties: Mapping[str, str] | None = None,
-        workspace_rid: str | None = None,
+        workspace: Workspace | str | None = None,
     ) -> WorkbookTemplate:
         """Create a workbook template from this workbook.
```
drake-nominal, line 272: In the past, we've been asked not to expose this publicly due to the fragile nature of the clientside request. We've historically decided to not expose this until the backend provides a clean API route to do this operation.

PR #911, opened 2026-08-06
feat(ingest): declare how numeric timestamps should be read for avro and journal json

drake-nominal:
Lets callers say how a file's numeric timestamps should be read, across every surface that ingests
avro or journal-json — including containerized extractor outputs, which previously could not say
anything at all.

## Why

An avro-stream file's timestamps are plain integers, and the schema does not record what unit they
are in. A manifest extractor that wrote an avro output had no way to declare that unit, so the output
inherited the *job-level* timestamp metadata. When the extractor's own input was a CSV, that
job-level type is commonly a string format — applied, silently, to a field holding integers.

## What changed

**Single-file ingest** — `Dataset.add_avro_stream` takes `timestamp_type` (numeric only: an absolute
epoch or an offset from a start) and `tags`. `Dataset.add_journal_json` takes
`timestamp_column`/`timestamp_type`, naming the field in each JSON payload that holds the timestamp
and the epoch unit it is in.

**Containerized extractors** — `ctx.add_avro_stream(timestamp_type=...)` declares the reading for one
output, overriding the job-level metadata for that file only. It takes a type and no column, because
the schema fixes which field holds the timestamps. Omit it to inherit the job-level metadata, which is
right only when that metadata already describes numeric timestamps.

`ctx.add_tabular` and `ctx.add_journal_json` gained `@overload` signatures, so passing one half of the
`timestamp_column`/`timestamp_type` pair is now a type error rather than a runtime one. That matters
more here than elsewhere in the client: the runtime error surfaces inside a container partway through
an ingest job, not at the call site. The runtime check stays for callers who do not type-check.

**Data scopes** — the wrapper forwards both option sets. See the behavior change below.

**Experimental ingest builder** — `IngestBuilder.add_avro_stream` takes `timestamp_type`. Unlike the
manifest there is no "absent" state here, since the multi-file endpoint wants timestamp metadata on
every file item, so omitting it still sends the canonical epoch-nanosecond reading as before.

**Typing** — a new `_AnyEpochTimestampType` alias lets journal-json accept `"epoch_microseconds"`-style
literals, not just `ts.Epoch` objects, matching every sibling method. Relative types are deliberately
excluded: the journal request carries a time unit with nowhere to record a starting offset, and the
alias documents that so nobody widens it later.

## Behavior change

`add_avro_stream` on a data scope used to raise `RuntimeError` when the selected scope required tags,
because avro ingest could not carry tags and ingesting untagged data was worse than failing. Avro
ingest applies additional file tags now, so those calls ingest, with the scope's tags merged in and
caller-provided tags winning on key collisions. The equivalent journal-json refusal stays — journal
ingest still has no tags field.

## Bug fixed along the way

The pairing guard introduced early in this branch rejected `add_journal_json` calls that passed
*neither* timestamp argument, which is the default call shape. Separately, the non-epoch guard ran
*after* the file upload, so a caller passing an inexpressible type uploaded the whole log file before
failing; it now runs before any I/O, matching the avro path.

## Typing note for reviewers

`ctx.add_tabular`'s `timestamp_type` narrowed from all timestamp types to the numeric-only set. Per-output
metadata never accepted anything else — the previous behavior was a runtime `ExtractorError` — so this
converts a runtime failure into a type error in an experimental module. Consequently, passing the pair
explicitly as `timestamp_column=None, timestamp_type=None` no longer type-checks; omit both instead.

## Testing

`just test` — 602 passed, 14 skipped. `just check` clean (ruff format, mypy strict over 161 source
files, ruff check).

New coverage: per-output avro metadata for epoch and relative types, plus the omitted case and the
rejection of types the manifest cannot express; the timestamp conversions this branch added at the
`ts` layer; the journal-json pairing guard in both directions and its pre-upload validation; the avro
options landing on the ingest request; data-scope tag merging with a colliding key; and the builder's
declared-versus-default reading.

### DIFF
--- a/nominal/core/dataset.py
+++ b/nominal/core/dataset.py
@@ -30,11 +30,15 @@ from nominal.core.log import LogPoint, _write_logs
 from nominal.core.video import _build_video_file_timestamp_manifest
 from nominal.core.video_dataset_file import VideoDatasetFile
 from nominal.ts import (
+    Epoch,
     IntegralNanosecondsUTC,
+    _AnyEpochTimestampType,
+    _AnyNumericTimestampType,
     _AnyTimestampType,
     _InferrableTimestampType,
     _SecondsNanos,
     _to_typed_timestamp_type,
+    _validate_timestamp_pair,
 )
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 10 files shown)


varun-nominal:
fyi the timestamp type stuff for avro only works on environments with v2 enabled (which is most places but just fyi)

varun-nominal:
looks like IngestBuilder.add_journal_json still takes _AnyTimestampType but the siblings are only numeric

varun-nominal (approved):

drake-nominal:
> looks like IngestBuilder.add_journal_json still takes _AnyTimestampType but the siblings are only numeric

@varun-nominal it may not be super obvious naming, but they should actually take `_AnyEpochTimestampType` which does what you'd expect it to do (allows "epoch_nanoseconds" etc. ,any type that's `Epoch` based). I'll switch.

PR #915, opened 2026-08-07
feat: migrate asset service to gRPC

drake-nominal:
Migrates the SDK's asset operations from Conjure to the v2 gRPC `AssetService`. Client method signatures remain unchanged, asset RPCs use `translate_grpc_errors()`, and search uses the shared gRPC paginator. `Asset` uses `RefreshableGrpcMixin` and exposes `updated_at`, allowing the download CLI to sort assets without another fetch.

`Asset` and `Run` each resolve dataset scope names through `_lookup_dataset_scope()`, returning only a dataset RID and series tags. `_DatasetWrapper` therefore stays independent of both wire formats; `Run` continues to read its Conjure scopes directly. Link normalization is shared by the Conjure and asset-proto builders.

Presence and error behavior:

- Missing assets raise `NominalNotFoundError`, including absent entries in a successful batch response.
- Updates distinguish omitted collections from explicit empty collections. Asset descriptions preserve the distinction between an absent value and an explicitly empty string.
- Scope removal compares only the active data-source oneof arm and rejects empty RIDs before making an RPC. Surviving scopes retain field presence, including absent offsets and data sources.

Validation on Python 3.13 after merging current `main`: 1,001 unit tests passed; mypy and Ruff lint/format checks passed. One optional video-stream test module was skipped because the local GStreamer WebRTC library is unavailable. Live-backend tests were not run.

### DIFF
--- a/nominal/cli/download.py
+++ b/nominal/cli/download.py
@@ -95,10 +95,7 @@ class DataDownloader(abc.ABC):
                 return None
 
         # Sort by last updated timestamps
-        raw_assets = self._client._clients.assets.get_assets(
-            self._client._clients.auth_header, [asset.rid for asset in assets]
-        )
-        sorted_assets = sorted(assets, key=lambda asset: pd.to_datetime(raw_assets[asset.rid].updated_at), reverse=True)
+        sorted_assets = sorted(assets, key=lambda asset: asset.updated_at, reverse=True)
         table = Table(
             Column("#", style=Style(color="white", bold=True), ratio=1, overflow="fold"),
             Column("Name", style=Style(color="white", bold=True), ratio=2, overflow="fold"),
@@ -418,15 +415,11 @@ class DataDownloader(abc.ABC):
             return
 
         # get tags from dataset & asset combo
-        scope_tags = None
-        raw_asset = self._client._clients.assets.get_assets(self._client._clients.auth_header, [asset.rid])[asset.rid]
-        for raw_datascope in raw_asset.data_scopes:
-            if raw_datascope.data_scope_name == refname:
-                scope_tags = raw_datascope.series_tags
-                break
-        if scope_tags is None:
+        data_scope = asset._lookup_dataset_scope(refname)
+        if data_scope is None:
             logger.error("Failed to retrieve datascope details for refname %s", refname)
             return
+        _, scope_tags = data_scope
 
         # Select channels (exact-name matching with iterative queries) to download
         channels = self._select_channels(dataset)
--- a/nominal/core/_utils/api_tools.py
+++ b/nominal/core/_utils/api_tools.py
@@ -5,12 +5,24 @@ import importlib.metadata
 import logging
 import platform
 import sys
-from typing import Any, Generic, Literal, Mapping, Protocol, Sequence, TypeAlias, TypedDict, TypeVar, runtime_checkable
+from typing import (
+    Any,
+    Generic,
+    Literal,
+    Mapping,
+    Protocol,
+    Sequence,
+    TypeAlias,
+    TypedDict,
+    TypeVar,
+    runtime_checkable,
+)
 
 from nominal_api import scout_asset_api, scout_compute_api, scout_run_api
 from typing_extensions import NotRequired, Self
 
 from nominal._utils.dataclass_tools import update_dataclass
+from nominal.protos.asset.v2 import asset_pb2
 
 ScopeTypeSpecifier: TypeAlias = Literal["connection", "dataset", "video"]
 
@@ -108,17 +120,31 @@ class LinkDict(TypedDict):
     title: NotRequired[str]
 
 
-def create_links(links: Sequence[str | Link | LinkDict]) -> list[scout_run_api.Link]:
-    links_conjure = []
+def normalize_links(links: Sequence[str | Link | LinkDict]) -> list[tuple[str, str | None]]:
+    """Reduce every accepted link spelling -- url, (url, title), or dict -- to (url, title) pairs."""
+    normalized: list[tuple[str, str | None]] = []
     for link in links:
         if isinstance(link, tuple):
             url, title = link
-            links_conjure.append(scout_run_api.Link(url=url, title=title))
+            normalized.append((url, title))
         elif isinstance(link, dict):
-            links_conjure.append(scout_run_api.Link(url=link["url"], title=link.get("title")))
+            normalized.append((link["url"], link.get("title")))
         else:
-            links_conjure.append(scout_run_api.Link(url=link))
-    return links_conjure
+            normalized.append((link, None))
+    return normalized
+
+
+def create_links(links: Sequence[str | Link | LinkDict]) -> list[scout_run_api.Link]:
+    return [scout_run_api.Link(url=url, title=title) for url, title in normalize_links(links)]
+
+
+def create_proto_links(links: Sequence[str | Link | LinkDict]) -> list[asset_pb2.Link]:
+    """The proto peer of `create_links`.
+
+    `Link` is currently declared per service and only the asset service has migrated, so this builds the
+    asset spelling. It collapses into `create_links` once `Link` becomes a shared type.
+    """
+    return [asset_pb2.Link(url=url, title=title) for url, title in normalize_links(links)]
 
 
 def create_api_tags(tags: Mapping[str, str] | None = None) -> dict[str, scout_compute_api.StringConstant]:
@@ -157,9 +183,3 @@ def filter_scopes(
     scopes: Sequence[scout_asset_api.DataScope], scope_type: ScopeTypeSpecifier
 ) -> Sequence[scout_asset_api.DataScope]:
     return [scope for scope in scopes if scope.data_source.type.lower() == scope_type]
-
-
-def filter_scope_rids(scopes: Sequence[scout_asset_api.DataScope], scope_type: ScopeTypeSpecifier) -> Mapping[str, str]:
-    return {
-        scope.data_scope_name: getattr(scope.data_source, scope_type) for scope in filter_scopes(scopes, scope_type)
-    }
--- a/nominal/core/_utils/query_tools.py
+++ b/nominal/core/_utils/query_tools.py
@@ -8,7 +8,6 @@ from nominal_api import (
     api,
     authentication_api,
     ingest_api,
-    scout_asset_api,
     scout_catalog,
     scout_checks_api,
     scout_notebook_api,
@@ -20,6 +19,7 @@ from nominal_api import (
 
 from nominal.core._event_types import EventType, SearchEventOriginType
 from nominal.core._utils.api_tools import rid_from_instance_or_string
+from nominal.protos.asset.v2 import asset_pb2
 from nominal.protos.event.v2 import event_pb2
 from nominal.protos.registry.v2 import registry_pb2
 from nominal.protos.secrets.v1 import secrets_pb2
@@ -237,22 +237,25 @@ def create_search_assets_query(
     properties: Mapping[str, str] | None = None,
     exact_substring: str | None = None,
     workspace_rid: str | None = None,
-) -> scout_asset_api.SearchAssetsQuery:
+) -> asset_pb2.SearchAssetsQuery:
     queries = []
     if search_text is not None:
-        queries.append(scout_asset_api.SearchAssetsQuery(search_text=search_text))
+        queries.append(asset_pb2.SearchAssetsQuery(search_text=search_text))
     if exact_substring is not None:
-        queries.append(scout_asset_api.SearchAssetsQuery(exact_substring=exact_substring))
+        queries.append(asset_pb2.SearchAssetsQuery(exact_substring=exact_substring))
     if labels is not None:
         for label in labels:
-            queries.append(scout_asset_api.SearchAssetsQuery(label=label))
+            queries.append(asset_pb2.SearchAssetsQuery(label=label))
     if properties:
         for name, value in properties.items():
-            queries.append(scout_asset_api.SearchAssetsQuery(property=api.Property(name=name, value=value)))
+            queries.append(asset_pb2.SearchAssetsQuery(property=types_pb2.Property(name=name, value=value)))
     if workspace_rid is not None:
-        queries.append(scout_asset_api.SearchAssetsQuery(workspace=workspace_rid))
+        queries.append(asset_pb2.SearchAssetsQuery(workspace=workspace_rid))
 
-    return scout_asset_api.SearchAssetsQuery(and_=queries)
+    # `and` is a Python keyword, and generated typing stubs cannot expose it as a named argument.
+    return asset_pb2.SearchAssetsQuery(
+        **{"and": asset_pb2.SearchAssetsQueryList(queries=queries)}  # type: ignore[arg-type]
+    )
 
 
 def create_search_ingest_jobs_query(
--- a/nominal/core/_utils/query_tools.py
+++ b/nominal/core/_utils/query_tools.py
@@ -8,7 +8,6 @@ from nominal_api import (
     api,
     authentication_api,
     ingest_api,
-    scout_asset_api,
     scout_catalog,
     scout_checks_api,
     scout_notebook_api,
@@ -20,6 +19,7 @@ from nominal_api import (
 
 from nominal.core._event_types import EventType, SearchEventOriginType
 from nominal.core._utils.api_tools import rid_from_instance_or_string
+from nominal.protos.asset.v2 import asset_pb2
 from nominal.protos.authorization.markings.v1 import markings_pb2
 from nominal.protos.event.v2 import event_pb2
 from nominal.protos.registry.v2 import registry_pb2
@@ -250,22 +250,25 @@ def create_search_assets_query(
     properties: Mapping[str, str] | None = None,
     exact_substring: str | None = None,
     workspace_rid: str | None = None,
-) -> scout_asset_api.SearchAssetsQuery:
+) -> asset_pb2.SearchAssetsQuery:
     queries = []
     if search_text is not None:
-        queries.append(scout_asset_api.SearchAssetsQuery(search_text=search_text))
+        queries.append(asset_pb2.SearchAssetsQuery(search_text=search_text))
     if exact_substring is not None:
-        queries.append(scout_asset_api.SearchAssetsQuery(exact_substring=exact_substring))
+        queries.append(asset_pb2.SearchAssetsQuery(exact_substring=exact_substring))
     if labels is not None:
         for label in labels:
-            queries.append(scout_asset_api.SearchAssetsQuery(label=label))
+            queries.append(asset_pb2.SearchAssetsQuery(label=label))
     if properties:
         for name, value in properties.items():
-            queries.append(scout_asset_api.SearchAssetsQuery(property=api.Property(name=name, value=value)))
+            queries.append(asset_pb2.SearchAssetsQuery(property=types_pb2.Property(name=name, value=value)))
     if workspace_rid is not None:
-        queries.append(scout_asset_api.SearchAssetsQuery(workspace=workspace_rid))
+        queries.append(asset_pb2.SearchAssetsQuery(workspace=workspace_rid))
 
-    return scout_asset_api.SearchAssetsQuery(and_=queries)
+    # `and` is a Python keyword, and generated typing stubs cannot expose it as a named argument.
+    return asset_pb2.SearchAssetsQuery(
+        **{"and": asset_pb2.SearchAssetsQueryList(queries=queries)}  # type: ignore[arg-type]
+    )
 
 
 def create_search_ingest_jobs_query(
--- a/nominal/core/asset.py
+++ b/nominal/core/asset.py
@@ -8,9 +8,6 @@ from typing import Iterable, Mapping, Protocol, Sequence, TypeAlias
 
 from nominal_api import (
     scout,
-    scout_asset_api,
-    scout_assets,
-    scout_run_api,
 )
 from typing_extensions import Self, deprecated
 
@@ -21,14 +18,13 @@ from nominal.core._utils.api_tools import (
     HasRid,
     Link,
     LinkDict,
-    RefreshableConjureMixin,
+    RefreshableGrpcMixin,
     ScopeTypeSpecifier,
-    create_links,
-    filter_scope_rids,
-    filter_scopes,
+    create_proto_links,
     rid_from_instance_or_string,
 )
 from nominal.core._utils.frontend_urls import asset_url
+from nominal.core._utils.grpc_tools import translate_grpc_errors
 from nominal.core._utils.pagination_tools import search_runs_by_asset_paginated
 from nominal.core._utils.query_tools import ArchiveStatusFilter
 from nominal.core.attachment import Attachment, _iter_get_attachments
@@ -36,11 +32,13 @@ from nominal.core.connection import Connection, _get_connection, _get_connection
 from nominal.core.dataset import Dataset, _create_dataset, _DatasetWrapper, _get_dataset, _get_datasets
 from nominal.core.datasource import DataSource
 from nominal.core.event import Event, _create_event, _search_events
-from nominal.core.exceptions import LegacyVideoDeprecationWarning
+from nominal.core.exceptions import LegacyVideoDeprecationWarning, NominalNotFoundError
 from nominal.core.video import Video, _create_video, _get_video
 from nominal.core.workbook import Workbook, _search_workbooks
+from nominal.protos.asset.v2 import asset_pb2, asset_pb2_grpc
 from nominal.protos.comments.v1 import comments_pb2_grpc
-from nominal.ts import IntegralNanosecondsDuration, IntegralNanosecondsUTC, _SecondsNanos
+from nominal.protos.types import types_pb2
+from nominal.ts import IntegralNanosecondsDuration, IntegralNanosecondsUTC
 
 ScopeType: TypeAlias = Connection | Dataset | Video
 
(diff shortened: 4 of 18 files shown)


nominal/core/asset.py
```diff
@@ -180,31 +190,28 @@ def remove_data_scopes(
...
-        conjure_asset = self._get_latest_api()
+        latest_asset = self._get_latest_api()
 
         data_scopes_to_keep = [
-            scout_asset_api.CreateAssetDataScope(
+            asset_pb2.CreateAssetDataScope(
                 data_scope_name=ds.data_scope_name,
                 data_source=ds.data_source,
                 series_tags=ds.series_tags,
                 offset=ds.offset,
```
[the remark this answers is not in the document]
drake-nominal: Confirmed and fixed. `DataScope.offset` and `.data_source` both report `has_presence=True`, so this was live: every scope without an offset was being rewritten with an explicit zero. Both now go through `HasField`, with a regression test I verified fails without the fix.

nominal/core/asset.py
```diff
@@ -697,32 +713,68 @@ def remove_attachments(self, attachments: Iterable[Attachment] | Iterable[str])
...
+
+
+def _get_asset(clients: Asset._Clients, rid: str) -> asset_pb2.Asset:
+    with translate_grpc_errors():
+        response = clients.assets.GetAssets(asset_pb2.GetAssetsRequest(rids=[rid]))
+    if rid not in response.responses:
+        raise ValueError(f"no asset found with RID {rid!r}: {response.responses!r}")
+    if len(response.responses) > 1:
+        raise ValueError(f"multiple assets found with RID {rid!r}: {response.responses!r}")
+    return response.responses[rid]
```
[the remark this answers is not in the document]
drake-nominal: `_get_asset` now raises `NominalNotFoundError`, and `get_asset`'s docstring documents it. Noting the exception type is a break for `except ValueError` callers; that is called out in the PR description rather than left implicit.

nominal/cli/download.py
```diff
@@ -95,10 +98,15 @@ def _select_asset(self) -> Asset | None:
                 return None
 
         # Sort by last updated timestamps
-        raw_assets = self._client._clients.assets.get_assets(
-            self._client._clients.auth_header, [asset.rid for asset in assets]
+        with translate_grpc_errors():
+            raw_assets = self._client._clients.assets.GetAssets(
+                asset_pb2.GetAssetsRequest(rids=[asset.rid for asset in assets])
+            ).responses
```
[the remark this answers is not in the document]
drake-nominal: Fixed, and then some — `_get_assets`/`_get_asset` were added per the `_get_datasets`/`_get_dataset` precedent, and the CLI no longer imports either. It sorts on a new `Asset.updated_at` field instead of re-fetching, so the extra RPC is gone too.

nominal/core/asset.py
```diff
@@ -697,32 +713,68 @@ def remove_attachments(self, attachments: Iterable[Attachment] | Iterable[str])
...
+    proto_links = []
+    for link in links:
+        if isinstance(link, tuple):
+            url, title = link
+            proto_links.append(asset_pb2.Link(url=url, title=title))
+        elif isinstance(link, dict):
+            proto_links.append(asset_pb2.Link(url=link["url"], title=link.get("title")))
+        else:
+            proto_links.append(asset_pb2.Link(url=link))
+    return proto_links
```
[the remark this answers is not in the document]
drake-nominal: De-duplicated, though via a shared dispatch rather than a message-class parameter: `normalize_links` reduces every spelling to `(url, title)` pairs and each transport maps those to its own `Link`.

tests/core/test_asset_grpc.py
```diff
@@ -0,0 +1,167 @@
...
+    asset = _asset(clients)
+    clients.assets.UpdateAsset.return_value = asset_pb2.UpdateAssetResponse(
+        asset=_proto_asset(asset.rid, title="renamed")
+    )
+
+    asset.update(name="renamed")
+
+    request = clients.assets.UpdateAsset.call_args.args[0]
+    assert request.asset_rid == asset.rid
+    assert request.title == "renamed"
```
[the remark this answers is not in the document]
drake-nominal: `UpdateAssetRequest.title` does have explicit presence (verified), so the omission case is meaningful and is now covered by `test_update_omitting_name_leaves_title_absent`.

tests/core/test_asset_grpc.py
```diff
@@ -0,0 +1,167 @@
...
+
+def test_archive_and_unarchive_address_this_asset() -> None:
+    clients = MagicMock()
+    asset = _asset(clients)
+
+    asset.archive()
+    asset.unarchive()
+
+    assert clients.assets.Archive.call_args.args[0].asset_rid == asset.rid
+    assert clients.assets.Unarchive.call_args.args[0].asset_rid == asset.rid
```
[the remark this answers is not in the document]
drake-nominal: Covered: name-matched and rid-matched removal with a surviving scope asserted intact, plus a dedicated regression for the `offset` presence issue.

tests/core/test_asset_grpc.py
```diff
@@ -0,0 +1,167 @@
...
+
+def test_from_proto_maps_absent_optionals_to_none() -> None:
+    """Description and created_by are optional on the wire; absent must read as None, not empty string."""
+    asset = Asset._from_proto(MagicMock(), _proto_asset())
+
+    assert asset.description is None
+    assert asset.created_by_rid is None
+
+
+def test_from_proto_reads_present_optionals() -> None:
```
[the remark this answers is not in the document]
drake-nominal: Docstrings added to both.

nominal/core/dataset.py
```diff
@@ -1246,12 +1246,10 @@ def _get_dataset_scope(self, data_scope_name: str) -> tuple[Dataset, Mapping[str
         data_scope = dataset_scopes.get(data_scope_name)
         if data_scope is None:
             raise ValueError(f"No such data scope found with data_scope_name {data_scope_name}")
```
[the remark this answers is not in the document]
drake-nominal: Fixed — the clause now reads `If no dataset-backed scope exists with the given data_scope_name.`

nominal/core/_utils/query_tools.py
```diff
@@ -236,22 +236,25 @@ def create_search_assets_query(
...
-            queries.append(scout_asset_api.SearchAssetsQuery(property=api.Property(name=name, value=value)))
+            queries.append(asset_pb2.SearchAssetsQuery(property=types_pb2.Property(name=name, value=value)))
     if workspace_rid is not None:
-        queries.append(scout_asset_api.SearchAssetsQuery(workspace=workspace_rid))
+        queries.append(asset_pb2.SearchAssetsQuery(workspace=workspace_rid))
 
-    return scout_asset_api.SearchAssetsQuery(and_=queries)
+    # `and` is a Python keyword, and generated typing stubs cannot expose it as a named argument.
+    return asset_pb2.SearchAssetsQuery(
+        **{"and": asset_pb2.SearchAssetsQueryList(queries=queries)}  # type: ignore[arg-type]
```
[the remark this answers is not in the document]
drake-nominal: Good catch — added `test_create_search_assets_query_wraps_clauses_in_and`, so the keyword-named field is pinned by a unit test rather than only by e2e.

nominal/experimental/migration/migrator/asset_migrator.py
```diff
@@ -144,8 +143,8 @@ def _copy_asset_datasets(self, source_asset: Asset, destination_asset: Asset, op
 
         for source_data_scope in source_data_scopes:
             source_data_scope_name = source_data_scope.data_scope_name
-            source_dataset_rid = source_data_scope.data_source.dataset
-            if source_dataset_rid is None or source_dataset_rid not in source_datasets:
+            source_dataset_rid = source_data_scope.dataset_rid
+            if source_dataset_rid not in source_datasets:
```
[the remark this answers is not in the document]
drake-nominal: Reworded to name the actual condition: the scope references a dataset that could not be resolved.

nominal/cli/download.py
```diff
@@ -95,10 +96,12 @@ def _select_asset(self) -> Asset | None:
                 return None
 
         # Sort by last updated timestamps
-        raw_assets = self._client._clients.assets.get_assets(
-            self._client._clients.auth_header, [asset.rid for asset in assets]
+        raw_assets = _get_assets(self._client._clients, [asset.rid for asset in assets])
```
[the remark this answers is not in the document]
drake-nominal: Closed by removing the fetch entirely — the CLI now sorts on `Asset.updated_at`, so there is no map to index and no fallback.

nominal/core/_utils/api_tools.py
```diff
@@ -108,17 +120,22 @@ class LinkDict(TypedDict):
     title: NotRequired[str]
 
 
-def create_links(links: Sequence[str | Link | LinkDict]) -> list[scout_run_api.Link]:
-    links_conjure = []
+def normalize_links(links: Sequence[str | Link | LinkDict]) -> list[tuple[str, str | None]]:
```
drake-nominal, line 123: we should have a `create_links` / `create_proto_links` variant that uses this helper vs. using this helper directly downstream.

nominal/core/_utils/api_tools.py
```diff
@@ -159,7 +176,21 @@ def filter_scopes(
     return [scope for scope in scopes if scope.data_source.type.lower() == scope_type]
 
 
-def filter_scope_rids(scopes: Sequence[scout_asset_api.DataScope], scope_type: ScopeTypeSpecifier) -> Mapping[str, str]:
-    return {
-        scope.data_scope_name: getattr(scope.data_source, scope_type) for scope in filter_scopes(scopes, scope_type)
-    }
+def conjure_dataset_scopes(scopes: Sequence[scout_asset_api.DataScope]) -> Sequence[asset_pb2.DataScope]:
```
drake-nominal: This is extremely hacky. Why are we converting conjure dataset scopes at all? It would be better to have separate helper methods.

drake-nominal: Deleted. There is no longer a normalized scope shape for anyone to convert into: `_DatasetWrapper` now declares `_lookup_dataset_scope(name) -> tuple[dataset_rid, series_tags] | None` and each resource resolves the name against the scopes its own transport returns — `Asset` over proto `DataScope`s, `Run` over conjure ones, with the malformed-scope raise moving to the conjure side where the field is actually optional. `filter_scope_rids`, `conjure_dataset_scopes`, and the `asset_pb2` import in `api_tools.py` are all gone, and `dataset.py` no longer imports an asset proto to describe a dataset. Both lookups are now unit-tested; the conjure one never was.

nominal/core/dataset.py
```diff
@@ -1226,18 +1227,18 @@ class _DatasetWrapper(abc.ABC):
...
+    Subclasses must implement `_dataset_scopes`, which is used to resolve scopes.
     """
 
     # static typing for required field
     _clients: Dataset._Clients
 
     @abc.abstractmethod
-    def _list_dataset_scopes(self) -> Sequence[scout_asset_api.DataScope]:
-        """Return the data scopes available to this wrapper.
+    def _dataset_scopes(self) -> Sequence[asset_pb2.DataScope]:
```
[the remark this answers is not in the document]
drake-nominal: Took the other route: rather than normalize both transports into a shared scope type, the wrapper no longer sees a scope at all. `_dataset_scopes` is replaced by an abstract `_lookup_dataset_scope(name) -> tuple[dataset_rid, series_tags] | None`, so `Asset` and `Run` each resolve the name against their own representation. Same outcome you were after — `conjure_dataset_scopes` and both `asset_pb2` imports (`api_tools.py`, `dataset.py`) are gone, and `download.py` no longer touches proto objects — without adding a third scope type to maintain.

nominal/cli/download.py
```diff
@@ -418,12 +415,9 @@ def download(self, channels_per_request: int, points_per_file: int) -> None:
...
-        for raw_datascope in raw_asset.data_scopes:
-            if raw_datascope.data_scope_name == refname:
-                scope_tags = raw_datascope.series_tags
-                break
+        scope_tags = next(
+            (scope.series_tags for scope in asset._dataset_scopes() if scope.data_scope_name == refname), None
+        )
         if scope_tags is None:
             logger.error("Failed to retrieve datascope details for refname %s", refname)
             return
```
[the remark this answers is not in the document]
drake-nominal: Fixed by the scope-contract change above: the CLI now calls `asset._lookup_dataset_scope(refname)` and reads the tags from the pair, so the duplicated `next(...)` lookup is gone and it no longer holds proto objects. `_select_dataset` keeps its signature — the asset fetch is the same one the old `_dataset_scopes()` call made, so no extra RPC.

nominal/core/_utils/query_tools.py
```diff
@@ -237,22 +237,25 @@ def create_search_assets_query(
...
```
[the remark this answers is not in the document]
drake-nominal: You are right and my earlier reply on this thread was wrong — that test never landed. Added now, in `tests/core/test_query_tools.py` alongside the events builder: one case asserting the full ANDed clause list for text/labels/properties/workspace, and one asserting an unfiltered search still wraps an empty clause list. The keyword-named field is pinned by unit tests rather than only by e2e.

nominal/core/_utils/api_tools.py
drake-nominal: That variant exists — `_create_proto_links` in `asset.py:777`, which is `create_links` with `asset_pb2.Link` in place of `scout_run_api.Link`. `normalize_links` has no direct downstream callers; the two builders are its only users. I added a docstring naming it as the proto peer so the pairing is discoverable from either side.

I kept it in `asset.py` rather than next to `create_links` because `Link` is declared per service (`nominal.asset.v2.Link`) with no shared `nominal.types` version, so a `create_proto_links` in `api_tools.py` would bake one service into the shared module and need a sibling or a message-class parameter the moment runs migrate — the same trap as the per-service `Duration` messages in #925. It would also re-add the `asset_pb2` import to `api_tools.py` that this push removes.

Happy to move it if you would rather have both spellings in one file; say the word and I will parametrize on the message class.

drake-nominal: Moved: `create_proto_links` now sits directly under `create_links` in `api_tools.py`, and `asset.py` imports the builder instead of `normalize_links`. Docstring notes it builds the asset spelling only because that is the one migrated service, and that it collapses into `create_links` once `Link` becomes shared. Their agreement is pinned in a new `tests/core/test_api_tools.py`, which covers both builders over the same input — the absent title is the only difference, since proto has no null string.

nominally-zoe (commented):
🤖 **Automated review, posted by a robot acting on behalf of Zoe Cohen (PE).** Left as a COMMENT; the approve/request-changes call stays with a human.

### Verdict
Clean transport swap; I would approve once the staging e2e run is green. Inline suggestions are small, one structural.

### What I verified (not from the description)
- Checked out `edbc5a7` in a worktree: ruff, ruff format, strict mypy, and the unit suite pass. The only mypy/collection failures are in `nominal/experimental/compute_as_code` (needs `nominal-compute`) and reproduce on `main`.
- Read the pinned `nominal-api-protos` descriptors: `UpdateAssetRequest.title/description/labels/properties/links/data_scopes/is_staged`, `DataScope.offset`/`data_source`, `Asset.description`/`created_by`, and `SearchAssetsRequest.next_page_token` all report `has_presence=True`, and `DataSource`'s oneof carries `dataset/connection/log_set/video/spatial`. Every `HasField` guard and `None`-omits-the-field claim holds.
- Server side is ready: `AssetGrpcResource` is registered, and ingress routes `nominal.asset.*` by package wildcard (combined-service ≥ 0.1345).

### Needs a human call
- **E2E never ran.** All three staging e2e jobs failed at setup on 2026-09-10 (`E2E_NOMINAL_STAGING_API_KEY secret is not configured`, then 503s on `/my/profile`); the same workflow passes on `main`. Nothing on this branch has hit a live `AssetService` yet. Re-run "E2E Tests (Staging)" before merge.
- **`get_asset` exception type.** `ValueError` → `NominalNotFoundError`, and `NominalNotFoundError` does not subclass `ValueError`, so `except ValueError` callers break. The body says so; the `feat:` commit does not. Either document it in the release notes or give it a soft landing via the `class X(NominalError, ValueError)` pattern already used in `exceptions.py`.
- **`Asset.updated_at` lands before `is_archived`** in a public frozen dataclass, shifting positional construction. Keyword construction is the norm, so probably a release-note line rather than a fix.
- **Scope contract shape.** `tuple[str, Mapping[str, str]] | None` forces `_, scope_tags = data_scope` at the CLI call site. A `NamedTuple` (`dataset_rid`, `series_tags`) names the fields without adding a concept. Author's call.

### Merge ordering with #916 / #962
This branch conflicts with #962 in `_clientsbunch.py`, `api_tools.py`, `asset.py`, and `run.py`, with #916 in `tests/core/test_clientsbunch.py`, and with `main` in `asset.py` (`spatial` joined `ScopeTypeSpecifier`). Suggested order: #915 → #916 → #962. The `create_proto_links` suggestion below is what lets #962 drop its `_create_run_links` copy when it rebases.

nominal/core/_utils/api_tools.py
```diff
@@ -5,12 +5,24 @@
...
+    Generic,
+    Literal,
+    Mapping,
+    Protocol,
+    Sequence,
+    TypeAlias,
+    TypedDict,
+    TypeVar,
+    runtime_checkable,
+)
```
nominally-zoe, line 19: `Callable` is needed for the parametrized `create_proto_links` below.

```suggestion
from typing import (
    Any,
    Callable,
    Generic,
    Literal,
    Mapping,
    Protocol,
    Sequence,
    TypeAlias,
    TypedDict,
    TypeVar,
    runtime_checkable,
)
```

nominal/core/_utils/api_tools.py
```diff
@@ -5,12 +5,24 @@
...
 
 from nominal_api import scout_asset_api, scout_compute_api, scout_run_api
 from typing_extensions import NotRequired, Self
 
 from nominal._utils.dataclass_tools import update_dataclass
+from nominal.protos.asset.v2 import asset_pb2
```
nominally-zoe, line 25: With `create_proto_links` parametrized on the message class, the shared helper module no longer imports one service's proto.

```suggestion
```

nominal/core/_utils/api_tools.py
```diff
@@ -108,17 +120,31 @@ class LinkDict(TypedDict):
...
+    return [scout_run_api.Link(url=url, title=title) for url, title in normalize_links(links)]
+
+
+def create_proto_links(links: Sequence[str | Link | LinkDict]) -> list[asset_pb2.Link]:
+    """The proto peer of `create_links`.
+
+    `Link` is currently declared per service and only the asset service has migrated, so this builds the
+    asset spelling. It collapses into `create_links` once `Link` becomes a shared type.
+    """
+    return [asset_pb2.Link(url=url, title=title) for url, title in normalize_links(links)]
```
nominally-zoe, line 147: The "only the asset service has migrated" justification lapses with #962, which adds `run_service_pb2.Link` and a second copy of this comprehension (`_create_run_links`). Parametrizing on the message class here means #962 reuses this instead of adding a sibling, and the shared module stops naming one service's proto. `T` is already declared at the top of this file.

```suggestion
def create_proto_links(links: Sequence[str | Link | LinkDict], link_type: Callable[..., T]) -> list[T]:
    """The proto peer of `create_links`; `link_type` is the calling service's `Link` message.

    Each proto package declares its own `Link`, so the caller names the message to build.
    """
    return [link_type(url=url, title=title) for url, title in normalize_links(links)]
```

nominal/core/asset.py
```diff
@@ -120,15 +129,22 @@ def update(
                 new_labels.append(old_label)
             asset = asset.update(labels=new_labels)
         """
-        request = scout_asset_api.UpdateAssetRequest(
+        # None omits a field, leaving it unchanged; an empty wrapper clears it.
+        updated_labels = None if labels is None else types_pb2.LabelUpdateWrapper(labels=list(labels))
+        updated_properties = (
+            None if properties is None else types_pb2.PropertyUpdateWrapper(properties=dict(properties))
+        )
+        updated_links = None if links is None else asset_pb2.LinkList(links=create_proto_links(links))
```
nominally-zoe, line 137:
```suggestion
        updated_links = None if links is None else asset_pb2.LinkList(links=create_proto_links(links, asset_pb2.Link))
```

nominal/core/asset.py
```diff
@@ -82,20 +81,30 @@ def nominal_url(self) -> str:
...
+            response = self._clients.assets.UpdateAsset(request)
+        return self._refresh_from_api(response.asset)
+
+    def _dataset_scopes(self) -> Sequence[asset_pb2.DataScope]:
+        return _filter_proto_scopes(self._get_latest_api().data_scopes, "dataset")
+
+    def _lookup_dataset_scope(self, data_scope_name: str) -> tuple[str, Mapping[str, str]] | None:
+        for scope in self._dataset_scopes():
+            if scope.data_scope_name == data_scope_name:
+                return scope.data_source.dataset, scope.series_tags
```
nominally-zoe, line 99: This returns the proto map container through the boundary the PR describes as transport-neutral. #962's `Run` implementation returns `dict(...)`; matching it here keeps the contract a plain `Mapping` and drops the `dict(series_tags)` wrappers in the tests.

```suggestion
    def _lookup_dataset_scope(self, data_scope_name: str) -> tuple[str, Mapping[str, str]] | None:
        for scope in self._dataset_scopes():
            if scope.data_scope_name == data_scope_name:
                return scope.data_source.dataset, dict(scope.series_tags)
```

nominal/core/asset.py
```diff
@@ -695,38 +721,70 @@ def remove_attachments(self, attachments: Iterable[Attachment] | Iterable[str])
...
+            description=asset.description if asset.HasField("description") else None,
+            properties=MappingProxyType(dict(asset.properties)),
             labels=tuple(asset.labels),
-            created_at=_SecondsNanos.from_flexible(asset.created_at).to_nanoseconds(),
+            created_at=asset.created_at.ToNanoseconds(),
+            updated_at=asset.updated_at.ToNanoseconds(),
             is_archived=asset.is_archived,
             _clients=clients,
-            created_by_rid=asset.created_by,
+            created_by_rid=asset.created_by or None,
```
nominally-zoe, line 759: `created_by` has explicit presence like `description` seven lines up; one idiom for both reads better than mixing `HasField` and `or None`.

```suggestion
            created_by_rid=asset.created_by if asset.HasField("created_by") else None,
```

nominal/core/client.py
```diff
@@ -1227,28 +1227,26 @@ def create_asset(
         labels: Sequence[str] = (),
     ) -> Asset:
         """Create an asset."""
-        request = scout_asset_api.CreateAssetRequest(
+        request = asset_pb2.CreateAssetRequest(
             description=description,
             labels=list(labels),
             properties={} if properties is None else dict(properties),
             typed_properties={},
```
nominally-zoe, line 1234: No-op on proto (an empty map is the default), and the sibling empty collections were already dropped from this request.

```suggestion
```

tests/core/test_clientsbunch.py
```diff
@@ -315,7 +317,7 @@ def test_experimental_as_user_returns_derived_nominal_client(monkeypatch):
     assert impersonated._clients.catalog._requests_session.headers[ON_BEHALF_OF_USER_RID_HEADER] == (
         "ri.authn.dev.user.target"
     )
-    assert impersonated._clients.assets._requests_session.headers[ON_BEHALF_OF_USER_RID_HEADER] == (
+    assert impersonated._clients.run._requests_session.headers[ON_BEHALF_OF_USER_RID_HEADER] == (
```
nominally-zoe, line 320: `run` becomes a gRPC stub in #962, which would have to move this assertion a second time. `attachment` stays on Conjure through all three migrations.

```suggestion
    assert impersonated._clients.attachment._requests_session.headers[ON_BEHALF_OF_USER_RID_HEADER] == (
```

tests/core/test_asset.py
```diff
@@ -140,3 +146,158 @@ def test_search_data_reviews_passes_archive_status(mock_asset):
...
+    assert [scope.data_scope_name for scope in kept] == ["ds"]
+    assert kept[0].data_source.dataset == "ri.dataset.1"
+    assert dict(kept[0].series_tags) == {"vehicle": "a"}
+    assert not kept[0].HasField("offset"), "an unset offset must not be re-sent as an explicit zero"
+
+
+@pytest.mark.parametrize("scopes", [[""], ["ri.video.1", ""]])
+def test_remove_data_scopes_rejects_empty_rids_before_calling_the_service(
+    asset_with_scopes, mock_clients, scopes
+) -> None:
```
nominally-zoe, line 233: The rest of the file carries a one-line docstring per test (D103 is off, so ruff will not remind).

```suggestion
def test_remove_data_scopes_rejects_empty_rids_before_calling_the_service(
    asset_with_scopes, mock_clients, scopes
) -> None:
    """An empty rid would match the unset arms of every scope, so it is rejected before any RPC is made."""
```

tests/core/test_asset.py
```diff
@@ -140,3 +146,158 @@ def test_search_data_reviews_passes_archive_status(mock_asset):
...
+def test_from_proto_reads_optional_fields() -> None:
+    """Description and created_by are optional on the wire; absent must read as None, not empty string."""
+    absent = Asset._from_proto(MagicMock(), _proto_asset())
+    assert (absent.description, absent.created_by_rid) == (None, None)
+
+    present = Asset._from_proto(MagicMock(), _proto_asset(description="d", created_by="ri.user.1"))
+    assert (present.description, present.created_by_rid) == ("d", "ri.user.1")
+
+
+def test_from_proto_preserves_an_explicitly_empty_description() -> None:
```
nominally-zoe, line 292:
```suggestion
def test_from_proto_preserves_an_explicitly_empty_description() -> None:
    """Presence, not truthiness, decides absence: an explicit empty string must survive as ""."""
```

PR #920, opened 2026-08-10
feat: add helper for batch create or update channel

bw-nom:
**Changes** 
- Add helper function `batch_update_or_create_channels` around the batch create or update endpoint
- Consolidate duplicate between batch_create and batch_add across client + tests

### DIFF
--- a/nominal/core/datasource.py
+++ b/nominal/core/datasource.py
@@ -374,7 +374,7 @@ class DataSource(HasRid):
         """
         return self._batch_write_channels(channels, self._clients.series_metadata.batch_create)
 
-    def batch_upsert_channels(
+    def batch_update_or_create_channels(
         self,
         channels: Sequence[CreateChannelRequest],
     ) -> BatchAddChannelsResult:
--- a/nominal/core/datasource.py
+++ b/nominal/core/datasource.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 import logging
 from dataclasses import dataclass, field
 from datetime import timedelta
-from typing import Iterable, Literal, Mapping, Protocol, Sequence, overload
+from typing import Callable, Iterable, Literal, Mapping, Protocol, Sequence, overload
 
 from nominal_api import (
     api,
@@ -372,17 +372,7 @@ class DataSource(HasRid):
             single batch, the entire batch will fail. Callers are responsible for deduplicating
             channel names before passing them to this method.
         """
-        if not channels:
-            return BatchAddChannelsResult(channels=[], missing=[])
-        for batch in batched(channels, _DEFAULT_CHANNEL_BATCH_SIZE):
-            requests = [_build_series_metadata_request(self.rid, req) for req in batch]
-            batch_request = timeseries_metadata_api.BatchCreateSeriesMetadataRequest(requests=requests)
-            self._clients.series_metadata.batch_create(self._clients.auth_header, batch_request)
-        created = {ch.name: ch for ch in self.get_channels(names=[req.name for req in channels])}
-        return BatchAddChannelsResult(
-            channels=list(created.values()),
-            missing=[req for req in channels if req.name not in created],
-        )
+        return self._batch_write_channels(channels, self._clients.series_metadata.batch_create)
 
     def batch_upsert_channels(
         self,
@@ -406,12 +396,19 @@ class DataSource(HasRid):
             batch will fail. Callers are responsible for deduplicating channel names before
             passing them to this method.
         """
+        return self._batch_write_channels(channels, self._clients.series_metadata.batch_create_or_update)
+
+    def _batch_write_channels(
+        self,
+        channels: Sequence[CreateChannelRequest],
+        write_batch: Callable[[str, timeseries_metadata_api.BatchCreateSeriesMetadataRequest], None],
+    ) -> BatchAddChannelsResult:
         if not channels:
             return BatchAddChannelsResult(channels=[], missing=[])
         for batch in batched(channels, _DEFAULT_CHANNEL_BATCH_SIZE):
             requests = [_build_series_metadata_request(self.rid, req) for req in batch]
             batch_request = timeseries_metadata_api.BatchCreateSeriesMetadataRequest(requests=requests)
-            self._clients.series_metadata.batch_create_or_update(self._clients.auth_header, batch_request)
+            write_batch(self._clients.auth_header, batch_request)
         created = {ch.name: ch for ch in self.get_channels(names=[req.name for req in channels])}
         return BatchAddChannelsResult(
             channels=list(created.values()),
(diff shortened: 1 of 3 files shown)


nominal/core/datasource.py
```diff
@@ -372,12 +372,43 @@ def batch_add_channels(
             single batch, the entire batch will fail. Callers are responsible for deduplicating
             channel names before passing them to this method.
         """
+        return self._batch_write_channels(channels, self._clients.series_metadata.batch_create)
+
+    def batch_upsert_channels(
+        self,
+        channels: Sequence[CreateChannelRequest],
```
drake-nominal, line 379: We don't expose the raw request type in our public methods (`CreateChannelRequest`) so that if the API changes in a backwards incompatible way, we can provide our own deprecation cycle. 

Canonically in this repo, we would wrap `CreateChannelRequest` in a frozen dataclass, and then return something akin to a Sequence of `Channel` objects vs. a channelsResult.

tests/core/test_datasource.py
```diff
@@ -26,73 +34,90 @@ def mock_datasource(mock_clients):
...
         CreateChannelRequest(name="ch2", data_type=ChannelDataType.STRING, description="a string"),
         CreateChannelRequest(name="ch3", data_type=ChannelDataType.INT, unit="m/s"),
     ]
-    mock_datasource.batch_add_channels(channels)
-
-    assert mock_clients.series_metadata.batch_create.call_count == 1
-    _, batch_req = mock_clients.series_metadata.batch_create.call_args[0]
+    getattr(mock_datasource, method)(channels)
+
+    endpoint_mock = getattr(mock_clients.series_metadata, endpoint)
```
drake-nominal: no getattr, setattr, or hasattr in repo if at all possible.

tests/e2e/test_core.py
```diff
@@ -464,6 +465,40 @@ def test_batch_add_channels(client: NominalClient, archive: ArchiveFn) -> None:
...
+    archive(ds)
+
+    seed = ds.batch_add_channels(
+        [
+            CreateChannelRequest(name="velocity", data_type=ChannelDataType.DOUBLE, unit="m/s"),
+            CreateChannelRequest(name="status", data_type=ChannelDataType.STRING, description="system status"),
+        ]
+    )
+    assert seed.missing == []
+
```
drake-nominal, line 482: should add some more exotic types, see the ones i'm adding in #919

nominal/core/datasource.py
```diff
@@ -372,12 +372,43 @@ def batch_add_channels(
             single batch, the entire batch will fail. Callers are responsible for deduplicating
             channel names before passing them to this method.
         """
+        return self._batch_write_channels(channels, self._clients.series_metadata.batch_create)
+
+    def batch_upsert_channels(
```
drake-nominal: more canonical naming in repo would be `update_or_create_channels` => inline with the `get_or_create` naming we do for other upserts.

nominal/core/datasource.py
```diff
@@ -372,12 +372,43 @@ def batch_add_channels(
...
+        channels: Sequence[CreateChannelRequest],
+        write_batch: Callable[[str, timeseries_metadata_api.BatchCreateSeriesMetadataRequest], None],
+    ) -> BatchAddChannelsResult:
         if not channels:
             return BatchAddChannelsResult(channels=[], missing=[])
         for batch in batched(channels, _DEFAULT_CHANNEL_BATCH_SIZE):
             requests = [_build_series_metadata_request(self.rid, req) for req in batch]
             batch_request = timeseries_metadata_api.BatchCreateSeriesMetadataRequest(requests=requests)
-            self._clients.series_metadata.batch_create(self._clients.auth_header, batch_request)
+            write_batch(self._clients.auth_header, batch_request)
```
drake-nominal: having different lambdas vs. plumbing a bool seems more complex. This feels like a weakening of the design of the methods to enable testability of a private method, which is a smell. We should focus unit tests on testing behaviors not implementations ( I put this precise line in my agents setup file, has helped immensely )

nominal/core/datasource.py
bw-nom: I think we're doing this at the top of the file?

nominal/core/datasource.py
bw-nom: Good shout, reworked the implementation and tests

tests/core/test_datasource.py
```diff
@@ -96,3 +96,96 @@ def test_batch_add_channels_returns_missing_when_server_drops_channel(mock_datas
...
+    req1 = CreateChannelRequest(name="ch1", data_type=ChannelDataType.DOUBLE)
+    req2 = CreateChannelRequest(name="ch2", data_type=ChannelDataType.DOUBLE)
+    mock_ch1 = MagicMock()
+    mock_ch1.name = "ch1"
+
+    with patch.object(DataSource, "get_channels", return_value=[mock_ch1]):  # ch2 not returned
+        result = mock_datasource.batch_update_or_create_channels([req1, req2])
+
+    assert result.channels == [mock_ch1]
+    assert result.missing == [req2]
```
alkasm: Technically `result.channels` and `result.missing` do not guarantee that they are lists; just sequences. but you're checking for equality against a list. that is correct in today's implementation but the attributes could instead return tuples for instance without breaking the type guarantees, but the tests would fail.

instead, better to do something like:

```suggestion
    assert len(result.channels) == 1
    assert len(result.missing) == 1
    assert result.channels[0] == mock_ch1
    assert result.missing[1] == req2
```

alkasm (approved):

tests/e2e/test_core.py
```diff
@@ -464,6 +465,41 @@ def test_batch_add_channels(client: NominalClient, archive: ArchiveFn) -> None:
...
+    result = ds.batch_update_or_create_channels(
+        [
+            # Existing channel: unit changed in place
+            CreateChannelRequest(name="velocity", data_type=ChannelDataType.UINT, unit="km/h"),
+            # Existing channel: description not supplied, so the existing value is preserved
+            CreateChannelRequest(name="gyro", data_type=ChannelDataType.DOUBLE_ARRAY),
+            # New channel: created fresh
+            CreateChannelRequest(name="temperature", data_type=ChannelDataType.DOUBLE, unit="degC"),
+        ]
```
drake-nominal, line 491: Any reason not to try one of each data type?

Other thing that would be good to test is with units. Here, you're using what seem to be valid UCUM units. HOWEVER, we can set "display only" nonsense channels as well. e.g., `"sheep"` is a valid unit.

Another thing to test would be what happens when the type changes. I would expect int => float to work, but probably not array => float. what about float => string (enum)?

drake-nominal (approved):
approve with 1 nit

tests/e2e/test_core.py
bw-nom: Will spin up a follow on PR to beef up e2e test

PR #929, opened 2026-08-14
feat: add commit and persist for derived datasets

adithya-nominal:
Editing a derived dataset's definition through the SDK doesn't show up in the dataset's history. The edit itself saves correctly, and the dataset always reflects the latest definition, but the backend records it as a temporary "working state" commit rather than a permanent one. Those commits are hidden from history and are eligible for automatic cleanup, so a dataset edited a few times still shows only its original creation in the API and UI, and older versions can disappear entirely, along with the ability to see what the definition used to be or roll back to it.

This PR adds commit_and_persist_derived_definition, which commits and then marks the commit permanent so the edit shows up in history and persist. It also deprecates the old function that does not persist.

### DIFF
--- a/nominal/core/_clientsbunch.py
+++ b/nominal/core/_clientsbunch.py
@@ -165,6 +165,7 @@ class ClientsBunch:
     storage: storage_datasource_api.NominalDataSourceService
     template: scout.TemplateService
     upload: upload_api.UploadService
+    versioning: scout.VersioningService
     video_file: scout_video.VideoFileService
     video: scout_video.VideoService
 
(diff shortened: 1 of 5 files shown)


david-nominal (commented):
hm, I don't think this should be necessary. `commit_derived_definition` should already be making a persisted commit. We don't do this in the derived dataset FE either. Have you seen any issues?

adithya-nominal:
David and I discussed this- looks like there were some changes in the Nominal application that caused us to not see commit history properly. Problem has been addressed and the updates went in. Closing this out.

PR #930, opened 2026-08-14
feat: add markings support for datasets, connections, and videos

drake-nominal:
Adds support for markings — the access-control primitive that restricts a data source to users in the marking's authorized groups. Markings were previously reachable only as raw RID strings on `create_dataset`; this makes them a first-class resource.

## What's new

**`Marking`** — a frozen dataclass with `rid`, `id` (a slug unique within the organization), `description`, `symbol`, `color`, `created_at`, `updated_at`, `is_archived`, plus `update()`, `archive()`, `unarchive()`, `refresh()`, and `authorized_group_rids()`.

**`Symbol`** — display metadata, built through classmethod constructors: `Symbol.icon(...)`, `Symbol.emoji(...)`, `Symbol.image(...)`. A marking's color is a plain hex string (`"#cc0000"`). The service stores only `^#[0-9a-f]{6}$`, so the SDK enforces the six-digit shape at the call boundary the way it does `id`, but accepts either case and lowercases it rather than making the caller do so.

**Markings on resources** — `Dataset`, `Connection`, `StreamingConnection`, and `Video` gain three methods from a shared mixin:

```python
dataset.list_markings()
dataset.apply_markings([itar])              # accepts Marking instances or RIDs
dataset.remove_markings(["ri.marking.x"])
```

**Client methods** — `create_marking`, `get_marking`, `get_marking_by_id`, `search_markings`.

**Markings at creation** — `markings=` on `create_dataset`, `create_streaming_connection`, `create_video`, and the experimental dataset helpers. Accepts `Marking` instances or RID strings; the existing string form is unchanged.

```python
itar = client.create_marking("export-controlled", description="ITAR", color="#cc0000")
dataset = client.create_dataset("flight-data", markings=[itar])
```

## Notes for reviewers

- **Markings apply only to data sources** (datasets, connections, videos). The mixin's membership is the guard — other resource types simply don't have these methods.
- **The interface is add/remove, not set-replace.** This is the one deliberate departure from how labels and properties work, and it is forced by the API. `GetMarkingsForResources` filters out markings the caller cannot read, so the applied set is never fully observable from the client; there is also no set/replace RPC, only an `UpdateMarkingsOnResource` delta of `markings_to_apply`/`markings_to_remove`. An earlier revision had a `set_markings` that computed that delta against `list_markings()`, but it could not honor its own contract: with an unreadable marking applied, "replace with exactly these" silently left it in place, and `set_markings([])` removed only what the caller could see. `apply_markings`/`remove_markings` name their targets outright, so they are correct without a view of the current state and need no read-then-write. The labels/properties idiom does not transfer because labels carry no per-item read permission.
- **`list_markings` may be a strict subset** of what is applied, for the same filtering reason. Its docstring says so, and says not to compute a desired set from it.
- **Creation with markings is atomic**: `marking_rids` is part of each create request, not a follow-up `apply_markings` call.
- **`search_markings` intentionally takes no `workspace` or `archive_status` argument** — markings are organization-scoped, and archived markings are excluded from search regardless. Results are oldest-first; no sort option is exposed. It also takes no case-insensitivity claim: `id` is server-validated as `^[a-z][a-z0-9-]*$`, so an id cannot contain uppercase and the question is unobservable.
- **Creating, updating, and archiving markings requires organization admin permission**; reading requires organization membership. Docstrings state this.
- `Marking` carries no `name` — only the `id` slug.
- `get_marking` and `get_marking_by_id` use the single-get endpoints rather than a one-element batch, so a marking that is missing and one the caller cannot read raise distinct errors (`NominalNotFoundError` vs `NominalPermissionDeniedError`) instead of both reporting as not-found.
- Deferred by design: a batch get-by-RIDs accessor, listing the resources a marking applies to, and `applied_at` timestamps. Also deferred: the RPC supports applying and removing in one atomic call, which nothing currently exposes — swapping one marking for another takes two calls. Worth adding as `update_markings(apply=..., remove=...)` if a caller needs it.

## Testing

730 unit tests pass; `just check` and strict mypy are clean. End-to-end coverage is read-only (`search_markings`, `list_markings`) because the write paths require organization-admin rights the e2e token isn't guaranteed to have — those paths have unit coverage only.

Suggest squash-merging: several commits here are intra-branch cleanups of private helpers that never shipped, and they would otherwise appear in the changelog.

### DIFF
--- a/nominal/core/__init__.py
+++ b/nominal/core/__init__.py
@@ -29,6 +29,7 @@ from nominal.core.event import Event
 from nominal.core.filetype import FileType, FileTypes
 from nominal.core.ingestion_job import IngestionJob, IngestionJobStatus, IngestType
 from nominal.core.log import LogPoint
+from nominal.core.marking import Marking
 from nominal.core.run import Run
 from nominal.core.secret import Secret
 from nominal.core.unit import Unit, UnitLike
@@ -77,6 +78,7 @@ __all__ = [
     "LinkDict",
     "LogPoint",
     "LogStream",
+    "Marking",
     "Comment",
     "NominalClient",
     "Run",
--- a/nominal/core/__init__.py
+++ b/nominal/core/__init__.py
@@ -24,7 +24,7 @@ from nominal.core.data_review import CheckViolation, DataReview, DataReviewBuild
 from nominal.core.dataset import Dataset
 from nominal.core.dataset_file import DatasetFile, IngestWaitType, as_files_ingested, wait_for_files_to_ingest
 from nominal.core.datasource import DataSource
-from nominal.core.elements import Color, Symbol, SymbolKind
+from nominal.core.elements import Symbol, SymbolKind
 from nominal.core.event import Event
 from nominal.core.filetype import FileType, FileTypes
 from nominal.core.ingestion_job import IngestionJob, IngestionJobStatus, IngestType
@@ -52,7 +52,6 @@ __all__ = [
     "filter_channels_with_data",
     "Checklist",
     "CheckViolation",
-    "Color",
     "Connection",
     "ContainerImage",
     "ContainerImageStatus",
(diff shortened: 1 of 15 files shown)


robertmhero (approved):
Approving as it faithfully exposes the existing marking API surface area 👍

PR #931, opened 2026-08-17
feat: map checklist assignee and data review creator during migration

seanmreidy:
## Problem

The migration impersonation resolver only fixes `created_by` attribution implicitly, via the on-behalf-of header on the client used to create each resource. Two checklist-related fields were left behind:

1. **Checklist assignee** — `assignee_rid` is a request field, not a calling identity, so impersonation headers can't set it. The `ChecklistMigrator` never passed it, and `_create_checklist_with_content` defaulted it to the creating user — the source checklist's actual assignee was silently dropped.
2. **Data review creator** — checklist executions ran via `destination_checklist.execute(...)`, which uses the client bound to the checklist (resolved from the checklist *author*). The new data review's `created_by` therefore reflected the checklist author, not the user who executed the source data review. (Data reviews have no assignee field at the API level — `created_by` is the only attribution.)

## Changes

- `MigrationContext` gains `user_rid_mapping` (source→destination user RIDs) and a `map_user_rid()` helper; plumbed through `MigrationRunner` and populated from the existing `migration.impersonation.source_to_destination_user_rids` config in the CLI.
- `ChecklistMigrator` translates the source checklist's `assignee_rid` through the mapping; an unmapped assignee falls back to the previous behavior (creating user). Also wires up the previously-dead `ChecklistCopyOptions.new_assignee_rid` override.
- `AssetMigrator._copy_asset_checklists` resolves the executing client from the **source data review** (whose `created_by_rid` the existing resolver already understands), so the destination data review is created on behalf of the mapped executor.
- README notes for both behaviors.

## Testing

- New unit tests (`tests/core/test_migration_user_attribution.py`) cover assignee mapping (mapped / unmapped fallback / explicit override) and data-review execution via the resolver-resolved client.
- `test_migrate_with_impersonation` (e2e) now creates a run + checklist + data review on the source and asserts the destination checklist's `assignee_rid` and the destination data review's `created_by` both equal the mapped destination user. Not run here — it requires `--impersonation-source-user-rid`/`--impersonation-dest-user-rid` against live environments.
- `just check` and `tests/core` pass locally (546 tests).

## Follow-up commits

- **`fix: propagate user_rid_mapping through the parallel migration path`** — the CLI always runs through `run_parallel_migration`, which builds its own `MigrationContext` and dropped the mapping, making assignee translation a no-op via `nom migrate copy` (caught by Devin review). Fixed and covered by a regression test that drives `run_parallel_migration` directly.
- **`test: fail migration e2e suite on skipped tests in CI`** — `test_migrate_with_impersonation` has silently skipped in CI since it was added (no impersonation user RIDs configured), so the green Migration E2E check overstated coverage. The suite now runs with `--fail-on-skip`, and the impersonation RIDs are wired from new secrets.

- **`test: derive impersonation source user from the source client`** — the source side of the mapping is now derived from the source token's own user (the creator of the test's resources) instead of a separately-configured secret that can drift; `E2E_IMPERSONATION_SOURCE_USER_RID` is no longer read.
- **`fix: address review feedback on migration user attribution`** — warn on unmapped users when a mapping is configured, skip the per-data-review checklist refetch when the clients bundle already matches, debug-log unmapped assignees, document `new_assignee_rid` as destination-side.
- **`test: split assignee-mapping e2e from impersonation` / `test: drop the impersonation created_by e2e test`** — assignee mapping is a plain request field and is covered end-to-end without impersonation (verified passing against staging in CI). E2E-verifying `created_by` attribution requires the staging e2e service account to hold org-admin (impersonation) rights, which is out of scope here — that test was removed; the executor-resolution change remains covered by unit tests.

> **Note for reviewers:** Migration E2E requires the `E2E_IMPERSONATION_DEST_USER_RID` repo secret (configured). A follow-up can add an impersonation e2e test once the staging service account is granted impersonation rights (org `adminCondition` in gatekeeper-orgs).

### DIFF
--- a/nominal/experimental/migration/migration_cli.py
+++ b/nominal/experimental/migration/migration_cli.py
@@ -572,6 +572,9 @@ def copy(
         asset_inclusion_config=asset_inclusion_config,
         destination_client=target_client,
         destination_client_resolver=destination_client_resolver,
+        user_rid_mapping=impersonation_config.source_to_destination_user_rids
+        if impersonation_config is not None
+        else None,
         migration_state_path=migration_state_path,
         dry_run=dry_run,
         video_ingest_timeout=(
--- a/nominal/experimental/migration/migrator/context.py
+++ b/nominal/experimental/migration/migrator/context.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import concurrent.futures
+import logging
 import threading
 from dataclasses import dataclass, field
 from datetime import timedelta
@@ -13,6 +14,8 @@ from nominal.experimental.migration.utils.video_file_utils import DEFAULT_INGEST
 
 DestinationClientResolver = Callable[[Any], NominalClient]
 Resource = TypeVar("Resource")
+
+logger = logging.getLogger(__name__)
 
 
 @dataclass
@@ -45,7 +48,12 @@ class MigrationContext:
         """Translate a source user RID to its destination equivalent, or None if unmapped."""
         if source_user_rid is None:
             return None
-        return self.user_rid_mapping.get(source_user_rid)
+        destination_user_rid = self.user_rid_mapping.get(source_user_rid)
+        # Only warn when a mapping is configured — an empty mapping is the normal
+        # non-impersonation case, not a misconfigured one.
+        if destination_user_rid is None and self.user_rid_mapping:
+            logger.warning("No mapped destination user RID for source user %s.", source_user_rid)
+        return destination_user_rid
 
     def record_mapping(self, resource_type: ResourceType, old_rid: str, new_rid: str) -> None:
         self.migration_state.record_mapping(resource_type=resource_type, old_rid=old_rid, new_rid=new_rid)
--- a/nominal/experimental/migration/migrator/context.py
+++ b/nominal/experimental/migration/migrator/context.py
@@ -4,7 +4,7 @@ import concurrent.futures
 import threading
 from dataclasses import dataclass, field
 from datetime import timedelta
-from typing import Any, Callable, TypeVar, cast
+from typing import Any, Callable, Mapping, TypeVar, cast
 
 from nominal.core import NominalClient
 from nominal.experimental.migration.migration_state import MigrationState
@@ -22,6 +22,9 @@ class MigrationContext:
     destination_client: NominalClient
     migration_state: MigrationState
     destination_client_resolver: DestinationClientResolver | None = None
+    user_rid_mapping: Mapping[str, str] = field(default_factory=dict)
+    """Source-to-destination user RID mapping, used to translate user-valued fields (e.g. checklist
+    assignee) that are set explicitly on requests rather than derived from the calling identity."""
     source_asset_rids: frozenset[str] = field(default_factory=frozenset)
     dry_run: bool = False
     video_ingest_timeout: timedelta | None = DEFAULT_INGEST_POLL_TIMEOUT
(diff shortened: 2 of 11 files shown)


nominal/experimental/migration/migrator/context.py
```diff
@@ -38,6 +41,12 @@ def destination_client_for(self, source_resource: Any) -> NominalClient:
             return self.destination_client
         return self.destination_client_resolver(source_resource)
 
+    def map_user_rid(self, source_user_rid: str | None) -> str | None:
+        """Translate a source user RID to its destination equivalent, or None if unmapped."""
+        if source_user_rid is None:
+            return None
+        return self.user_rid_mapping.get(source_user_rid)
```
seanmreidy: We should probably log a warn if we don't have a mapped destination user RID.

varun-nominal (approved):

nominal/experimental/migration/migration_cli.py
```diff
@@ -572,6 +572,9 @@ def copy(
         asset_inclusion_config=asset_inclusion_config,
         destination_client=target_client,
         destination_client_resolver=destination_client_resolver,
+        user_rid_mapping=impersonation_config.source_to_destination_user_rids
+        if impersonation_config is not None
+        else None,
```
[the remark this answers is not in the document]
seanmreidy: Fixed in 1031aae — `run_parallel_migration` now passes `user_rid_mapping=runner.user_rid_mapping` into the `MigrationContext` it builds, and a regression test (`test_parallel_migration_passes_user_rid_mapping_to_context`) drives `run_parallel_migration` directly with a mapping and asserts the assignee is translated. Verified the test fails without the fix. Good catch — the sequential path used by the e2e tests had the mapping wired, so nothing else would have surfaced this.

nominal/experimental/migration/migrator/context.py
seanmreidy: Added in 0baa832 — `map_user_rid` now logs a warning naming the source user when a configured mapping has no entry for them. It stays quiet when no mapping is configured at all (the normal non-impersonation case), so the warning only fires when attribution is actually degrading. Covered by `test_map_user_rid_warns_only_when_a_configured_mapping_misses`.

seanmreidy:
Responding to the Claude review findings — all four addressed:

1. **`user_rid_mapping` never reaches the migrators in the CLI** — fixed in 1031aae: `run_parallel_migration` passes `runner.user_rid_mapping` into its `MigrationContext`, with a regression test driving the parallel path directly. The suggested consolidation of the two context construction sites (a single `MigrationRunner.build_context()`, which would also remove the `setattr` resolver hack) is deliberately deferred to a follow-up PR to keep this one scoped.
2. **Re-fetching the checklist just to swap the auth header** — fixed in 0baa832, with one deviation from the suggested `dataclasses.replace`: the code reuses the already-bound checklist when the executing client shares its clients bundle (identity check) and refetches only when it differs. Same round-trip savings in the common case, without constructing a `Checklist` whose fields and credentials come from different clients. Covered by `test_data_review_execution_skips_refetch_when_clients_match`.
3. **Unmapped assignee dropped silently** — fixed in 0baa832: debug log names the source assignee with no mapping; additionally `map_user_rid` now warns whenever a configured mapping misses (per @seanmreidy's review comment).
4. **`new_assignee_rid` semantics undocumented** — fixed in 0baa832: field docstring states it's a destination-side RID that bypasses the mapping.

PR #933, opened 2026-08-18
feat: add File Store drives, files, and revisions to nominal.core

drake-nominal:
Adds a File Store API to `nominal.core`: workspace-scoped drives containing files, with revision history.

## Public surface

```python
client.create_drive(id) / get_drive(rid) / get_drive_by_id(id) / list_drives()

drive.get_file(path) / list_files(parent_path) / put_file(local_path, destination_path)
drive.apply_changes([...])          # batched, <=1000
drive.rename() / archive() / unarchive()      # organization-admin only

file.revisions() / move_to(dest) / remove() / download(dir)
revision.restore(dest) / download(dir)
```

22 new exports from `nominal.core`, plus `NominalFileStoreError` / `FileStoreErrorCode` from `nominal.core.exceptions`. New guide at `docs/src/file-store.md`.

## Design decisions worth a reviewer's attention

**Managed vs virtual drives are separate classes.** `client.get_drive()` returns a `Drive` or a `VirtualDrive` chosen from the drive's source, following the existing `Connection` / `StreamingConnection` pattern. A virtual drive is fully readable but its write methods raise `NominalFileStoreError(READ_ONLY_DRIVE)` — client-side, before any request. Files split the same way: `DriveFile` is an abstract base, with `ManagedDriveFile` (has a RID and history) and `VirtualDriveFile` (identified by a provider tuple, pinned by content).

**`apply_changes` returns results; the single-file methods raise.** The asymmetry is deliberate: a batch is applied in order and continues past a failure, so raising would discard successes that already landed. Single-change calls raise, because that reads better for one file.

**Every guard fires before work is spent.** Read-only drives, missing paths, directories, empty files, cross-drive destinations, and unsafe filenames all fail locally before an upload or an RPC. Each guard's test asserts the RPC was *never called*, not merely that an exception was raised.

## The one change with blast radius

`nominal/core/_utils/multipart.py` now threads an upload `destination` and passes the returned `bucket` handle back on every follow-up call — **on all upload paths, not just File Store**. This is what the upload API's own contract specifies, and the value returned on the default path is the default uploads target, so the echo is a no-op semantically. The three public helpers (`put_multipart_upload`, `upload_multipart_io`, `upload_multipart_file`) keep their exact signatures and return values; only private helpers gained parameters.

Datasets, videos, attachments, and the experimental batch uploader all traverse this function, so it got the closest review on the branch. Two regression tests pin the behaviour on both paths. Worth one live upload of a plain dataset before merging, since no unit test can prove the deployed server tolerates the extra parameter.

## Testing

Suite goes from 662 to 727 passed (16 skipped), no regressions. `ruff`, `mypy -p nominal` (strict), and `mkdocs build --strict` all clean.

## Not included

Ingest-from-drive-file (the obvious next PR), virtual drive configuration, ingest rules, purge, batch put, drive descriptions (needs an API bindings bump), CLI commands.

## Note on one commit

`6e83151` is a one-line docstring fix in `nominal/thirdparty/matlab/_matlab.py`, unrelated to the File Store. The strict docs build was already failing on `main` because of it, and this PR adds a docs page that build has to validate. Kept as its own commit so it can be dropped or split out.

## Known pre-existing issue found while working here

`_put_multipart_upload_to` never closes its `requests.Session` and never shuts down its `ThreadPoolExecutor`, leaking both on every upload. Predates this branch and is untouched by it — filing separately rather than changing resource lifetimes in the function this PR reroutes all uploads through.


docs/src/file-store.md
```diff
@@ -0,0 +1,253 @@
+# File Store
```
alkasm, line 1: Public facing name is just "Nominal Drives"

nominal/core/file_store/__init__.py
alkasm, line 1: let's use `fs` instead of `file_store` for the module path

nominal/core/file_store/drive.py
```diff
@@ -0,0 +1,379 @@
...
+        if size_bytes == 0:
+            raise ValueError(f"cannot upload an empty file: {path}")
+
+        # The drive path decides the file's identity and its stored suffix, so the upload is
+        # named after the destination rather than the local file.
+        filename = _basename(destination_path)
+        if not filename:
+            raise ValueError(f"destination path {destination_path!r} has no filename")
+        # `from_path` never raises — it falls back to the default for an unknown extension. The
+        # default is spelled explicitly here because the parameter's own default has a typo.
```
alkasm, line 249: > default is spelled explicitly here because the parameter's own default has a typo.

🤨

nominal/core/file_store/changes.py
```diff
@@ -0,0 +1,117 @@
...
+    _file_from_proto,
+)
+from nominal.protos.file_store.v1 import files_pb2
+
+MAX_CHANGES_PER_REQUEST = 1000
+"""Most changes the backend accepts in a single `Drive.apply_changes` call."""
+
+
+@dataclass(frozen=True)
+class MoveFile:
```
alkasm, line 31: `PutFile` is also a change that can be applied in batch, but I see why it's not included since upload is out-of-band.

when I built something similar before, it was transaction/commit based, the API ended up looking something like:

```python
with drive.open_transaction() as txn:
    txn.put("path.txt")
    txn.move("a.txt", "b.txt")
```
and on context management closing it would do the uploads and construct the changes in a batch and pushed the changes to the management API. I could imagine something similar here even though it's not transaction-based, could be something like ~:

```python
with drive.batch_changes() as batch:
    batch.put("path.txt")
    batch.move("a.txt", "b.txt")
assert all(r.success for r in batch.results())
```

Not super-pressing but I wonder if we should nope out of the `apply_changes` method for now and just worry about single interactions, and build this batch API in another PR

nominal/core/file_store/drive.py
```diff
@@ -0,0 +1,379 @@
...
+
+        Raises:
+            NominalFileStoreError: Always, with code `READ_ONLY_DRIVE`.
+        """
+        raise NominalFileStoreError(
+            FileStoreErrorCode.READ_ONLY_DRIVE,
+            f"drive {self.id!r} is backed by {self.source.value} and is read-only through Nominal",
+        )
+
+    def put_file(
```
alkasm, line 330: can we also get a `put_file_obj`

nominal/core/file_store/file.py
```diff
@@ -0,0 +1,541 @@
...
+class DriveDirectory(DriveEntry):
+    """A directory in a drive listing.
+
+    Directories are implied by the paths of the files beneath them; they carry no metadata
+    of their own.
+    """
+
+
+@dataclass(frozen=True)
+class DriveFileRevision(HasRid):
```
alkasm, line 91: idt we need to prefix everything with drive, we can probably keep everything implied by the module? wdyt?

PR #935, opened 2026-08-25
fix(migration): preserve dataset owners during impersonation

varun-nominal:
Dataset resources do not expose `created_by_rid` or `author_rid`, so the migration impersonation resolver silently fell back to the destination service user. This caused migrated datasets to be owned by the migration caller even when an impersonation mapping existed.

Use the existing Role Service-backed `get_dataset_owner_rid` helper as a dataset-specific fallback. Owner lookup failures retain the existing service-user behavior. Add regression coverage for owner resolution, graceful fallback, selection of the impersonated destination client, and a manual E2E that verifies the destination `ROLE_OWNER`.

### Testing

- `uv run ruff check nominal/experimental/migration/migration_cli.py tests/core/test_migration_impersonation.py tests/e2e/migration/test_migration.py`
- `uv run ruff format --check nominal/experimental/migration/migration_cli.py tests/core/test_migration_impersonation.py tests/e2e/migration/test_migration.py`
- `uv run pytest tests/core/test_migration_impersonation.py tests/core/test_experimental_dataset_utils.py tests/core/test_migration_destination_client_resolution.py tests/core/test_migration_user_attribution.py`
- `uv run pytest tests/e2e/migration/test_migration.py --collect-only -q`
- `uv run mypy -m nominal.experimental.migration.migration_cli`

### Manual E2E

Requires source and destination profiles plus an active destination user that the destination caller may impersonate. The test creates a source asset and dataset, migrates them, and verifies the destination dataset owner through Role Service.

```sh
uv run pytest tests/e2e/migration/test_migration.py \
  -k dataset_owner_with_impersonation \
  --source-profile=<source> \
  --dest-profile=<destination> \
  --impersonation-dest-user-rid=<destination-user-rid> \
  --fail-on-skip -v
```

### DIFF
--- a/nominal/experimental/migration/migration_cli.py
+++ b/nominal/experimental/migration/migration_cli.py
@@ -11,9 +11,10 @@ import click
 import yaml
 
 from nominal.cli.util.global_decorators import client_options, global_options
-from nominal.core import ArchiveStatusFilter, Asset, Checklist, NominalClient, Workbook
+from nominal.core import ArchiveStatusFilter, Asset, Checklist, Dataset, NominalClient, Workbook
 from nominal.core._utils.grpc_tools import translate_grpc_errors
 from nominal.experimental import as_user
+from nominal.experimental.dataset_utils import get_dataset_owner_rid
 from nominal.experimental.migration.config.migration_data_config import AssetInclusionConfig, MigrationDatasetConfig
 from nominal.experimental.migration.config.migration_resources import AssetResources, MigrationResources
 from nominal.experimental.migration.migration_decorators import migration_client_options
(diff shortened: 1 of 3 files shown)


seanmreidy (approved):

seanmreidy:
**Post-merge CI note**: `test_migrate_preserves_dataset_owner_with_impersonation` failed in CI from this PR's merge until 2026-08-26. All root causes were environment configuration, not code — this test was the first in CI to exercise an impersonated write, and it surfaced three latent gaps in the E2E setup:

1. **The staging E2E service user was not an org admin.** Gatekeeper requires the *caller* to be org-admin to act on-behalf-of another user (`AuthorizationResource.checkAdmin`), so the first impersonated write 403'd (`Authorization:NotAuthorizedAdmin`). The impersonation rid secrets had been staged earlier, but nothing had ever exercised the permission until this test.
2. **After swapping `E2E_NOMINAL_STAGING_API_KEY` to an org-admin account, profile validation failed** — the misleading "Failed to authenticate / base_url may be incorrect" error was actually the *workspace* check: `E2E_NOMINAL_STAGING_WORKSPACE_RID` pointed at a workspace the previous service user could see but the new account could not. Fixed by pointing it at a workspace the new account belongs to.
3. **`E2E_IMPERSONATION_DEST_USER_RID` held the production-instance form of the user rid.** Staging resolved the impersonation to the right identity anyway (same uuid), but the ownership assertion compares full rids, so the instance prefix mismatched. Fixed with the staging-instance rid — now set to a real user distinct from the service account, which is the stronger fixture: the test proves impersonation actually changed ownership rather than passing vacuously.

As of https://github.com/nominal-io/nominal-client/actions/runs/32997207727 the Migration E2E suite is fully green, including this test. (The two `test_core.py` pandas-export failures in the general E2E job are unrelated — a staging export-service 500 under separate investigation.)

PR #948, opened 2026-09-02
feat: add nominal-ibis

sgupta229:
[Fixes](https://linear.app/nominal-io/issue/WHS-36/wire-ibis-adapter-for-internal-sql-interface-dogfood)

Adds the `nominal-ibis` package, registering a Nominal backend with Ibis. It uses an authenticated `NominalClient` and its default workspace, discovers table schemas from the SQL catalog, compiles Ibis expressions to Nominal-compatible SQL, and executes queries through the generated SQL gRPC service.

```python
import ibis
from ibis import _
from nominal.core import NominalClient

client = NominalClient.from_profile("prod")
con = ibis.nominal.connect(client)
pts = con.table("points_double")
result = (
    pts.filter(_.dataset_rid == "ri.catalog....", _.channel == "temperature")
    .group_by(minute=_.ts.truncate("m"))
    .agg(n=_.count(), avg=_.value.mean(), last=_.value.argmax(_.ts))
    .to_pandas()
)
```

Catalog schemas use the recursive scalar, array, and map metadata from the SQL catalog API, including nested types such as assets' `ARRAY<MAP<VARCHAR, VARCHAR>>` links. The dynamic `MAP<ANY, ANY>` type is exposed as a string containing JSON, matching Scout’s serialization of `points_struct.value`; users can parse it with `json.loads()` after fetching. Concretely typed maps remain maps. Missing, unknown, or other unresolved `ANY` types raise `NominalSqlError` when opening that table. Discovery and other tables remain usable; `con.sql(...)` can select concrete columns or explicitly cast dynamic values.

**Server dependency:** `re_search` now compiles to Postgres's `~` operator, which the SQL service accepts as of nominal-io/scout#21053. Merge after the scout release carrying that change is deployed.

### Testing plan

- Unit tests


packages/nominal-sql/nominal/sql/_functions.py
```diff
@@ -0,0 +1,26 @@
...
+
+from __future__ import annotations
+
+from ibis import udf
+
+__all__ = ["derivative", "integral"]
+
+
+@udf.agg.builtin
+def derivative(value: float) -> float:
```
sgupta229: i wonder if there's an easy way to expose these automatically ? vs having to manually add them to this file and keep it in sync with what we expose via the SQL API directly?

packages/nominal-ibis/nominal/ibis/__init__.py
```diff
@@ -0,0 +1,32 @@
...
+    from ibis import _
+
+    con = ibis.nominal.connect()  # uses the "default" profile in the Nominal config
+    pts = con.table("points_double")
+    (
+        pts.filter(_.dataset_rid == "ri.catalog....", _.channel == "temperature")
+        .group_by(minute=_.ts.truncate("m"))
+        .agg(n=_.count(), avg=_.value.mean())
+        .to_pandas()
+    )
```
alkasm: the result of this is just thrown away, so should update the example

packages/nominal-ibis/nominal/ibis/_backend.py
```diff
@@ -0,0 +1,404 @@
...
+                environment variable, then to "default".
+            base_url: API base URL; overrides the profile's URL when given.
+            token: API key or auth token. When given, the Nominal config is not read.
+            workspace_rid: Workspace to query in; overrides the profile's workspace.
+                When neither is set, the caller's default workspace is used.
+            timeout_seconds: HTTP timeout for catalog and query requests.
+        """
+        # nominal.core must initialize before nominal.config: the two circularly
+        # import each other and only the core-first order resolves. A plain
+        # import sorts before the from-imports, so formatters keep this order.
```
alkasm: this is a bug we should fix

packages/nominal-ibis/nominal/ibis/_backend.py
```diff
@@ -0,0 +1,404 @@
...
+        import nominal.core  # noqa: F401
+        from nominal.config import NominalConfig
+        from nominal.core._constants import DEFAULT_API_BASE_URL
+        from nominal.core._utils.api_tools import construct_user_agent_string
```
alkasm: separately is it really necessary to lazily import these inline?

packages/nominal-ibis/nominal/ibis/_backend.py
```diff
@@ -0,0 +1,404 @@
...
+        token: str | None = None,
+        workspace_rid: str | None = None,
+        timeout_seconds: float = 120.0,
+    ) -> None:
+        """Connect to the Nominal SQL API.
+
+        Args:
+            profile: Named profile in the Nominal config (see `nom config profile add`).
+                Used when no token is given; defaults to the `NOMINAL_PROFILE`
+                environment variable, then to "default".
```
alkasm: there is intentionally no "default" profile, and the client library should not assume one

packages/nominal-ibis/nominal/ibis/_backend.py
```diff
@@ -0,0 +1,404 @@
...
+            total=3,
+            backoff_factor=0.5,
+            status_forcelist=(429, 502, 503, 504),
+            allowed_methods=frozenset({"GET", "POST"}),
+        )
+        self._session.mount("https://", HTTPAdapter(max_retries=retries))
+        self._catalog_cache: dict[str, sch.Schema] | None = None
+        self._catalog_functions: list[dict[str, Any]] = []
+        self._fn: Functions | None = None
+        self.workspace_rid = workspace_rid or self._default_workspace_rid()
```
alkasm: why is this all inlined rather than using `NominalClient` itself?

packages/nominal-ibis/nominal/ibis/_backend.py
```diff
@@ -0,0 +1,404 @@
...
+    def _default_workspace_rid(self) -> str:
+        response = self._session.get(f"{self.base_url}/workspaces/v1/default-workspace", timeout=self._timeout)
+        self._raise_for_error(response)
+        workspace = response.json() if response.status_code != 204 and response.content else None
+        if not workspace:
+            raise NominalSqlError("No default workspace is configured for this user; pass workspace_rid to connect()")
+        return str(workspace["rid"])
+
+    def _fetch_catalog(self) -> dict[str, sch.Schema]:
+        if self._catalog_cache is None:
```
alkasm: instead of `_catalog_cache` being optional and then mutating it inside this function, make this a cached property:
```python
@functools.cached_property
def catalog_cache(self) -> dict[str, sch.Schema]:
    ...
```

alkasm: then it wont be possible to have invalid state, at least w.r.t. the catalog cache being `None` unexpectedly

packages/nominal-ibis/nominal/ibis/_backend.py
```diff
@@ -0,0 +1,404 @@
...
+            detail = response.text[:2000]
+        raise NominalSqlError(f"HTTP {response.status_code}: {detail}")
+
+    def _default_workspace_rid(self) -> str:
+        response = self._session.get(f"{self.base_url}/workspaces/v1/default-workspace", timeout=self._timeout)
+        self._raise_for_error(response)
+        workspace = response.json() if response.status_code != 204 and response.content else None
+        if not workspace:
+            raise NominalSqlError("No default workspace is configured for this user; pass workspace_rid to connect()")
+        return str(workspace["rid"])
```
alkasm: we already have a utility function for resolving a default workspace

packages/nominal-ibis/nominal/ibis/_backend.py
```diff
@@ -0,0 +1,404 @@
...
+        }
+        if max_rows is not None:
+            body["maxRows"] = max_rows
+        return body
+
+    def _execute_sql(self, sql: str, max_rows: int | None = None) -> pa.Table:
+        response = self._session.post(
+            f"{self.base_url}/sql/v1/query", json=self._query_body(sql, max_rows), timeout=self._timeout
+        )
+        self._raise_for_error(response)
```
alkasm: why is this just hitting raw HTTP endpoints rather than going through conjure or protobuf generated libraries?

packages/nominal-ibis/nominal/ibis/_backend.py
sgupta229: moved

nominal/_exceptions.py
alkasm, line 1: the `nominal` folder MUST be empty since this is a namespace package
this is not the correct solution to the import loop

nominal/core/_utils/grpc_tools.py
```diff
@@ -302,18 +304,37 @@ def _build_grpc_channel(api_base_url: str, service_config: ServiceConfiguration,
 }
 
 
+def _grpc_error_message(error: grpc.RpcError) -> str:
```
alkasm, line 307: not clear what this is for or what it does

nominal/core/_utils/grpc_tools.py
```diff
@@ -23,11 +23,13 @@
...
 from functools import lru_cache
 from pathlib import Path
-from typing import Any, Iterator, Protocol, TypeVar
+from typing import Any, Iterator, Protocol, TypeVar, cast
 from urllib.parse import urlparse
 
 import grpc
 from conjure_python_client import ServiceConfiguration
+from google.protobuf.message import DecodeError
+from google.rpc import error_details_pb2, status_pb2  # type: ignore[import-untyped]
```
alkasm, line 32: don't add inline ignores for untyped modules, instead do module ignores in the mypy config section of pyproject.toml

however, I'm also confused at the ignore in the first place as it looks like this package _is_ actually typed so any ignore should be unnecessary: https://github.com/googleapis/google-cloud-python/tree/main/packages/googleapis-common-protos/google/rpc

docs/src/reference/thirdparty/ibis.md
alkasm, line 1: this should not be in `thirdparty` (which is somewhat vestigial from before we were doing a namespace package)

packages/nominal-ibis/pyproject.toml
```diff
@@ -0,0 +1,36 @@
...
+license = "Apache-2.0"
+license-files = ["LICENSE", "NOTICE"]
+readme = { text = "Ibis backend for the Nominal Python client.", content-type = "text/markdown" }
+requires-python = ">=3.10,<4"
+dependencies = [
+    "nominal>=1.164.0",
+    "ibis-framework>=12,<13",
+    "pyarrow>=14",
+    "pyarrow-hotfix>=0.4",
+    "packaging>=23.0",
```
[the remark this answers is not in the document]
alkasm: I'm confused at what the comment here is saying that negates this, let's discuss

packages/nominal-ibis/pyproject.toml
```diff
@@ -0,0 +1,38 @@
...
+    "nominal-api-protos>=0.1441.0",
+    "ibis-framework>=12,<13",
+    "pyarrow>=14",
+    # ibis.formats imports these at runtime but ibis only declares them in per-backend extras.
+    "pyarrow-hotfix>=0.4",
+    "packaging>=23.0",
+]
+
+[project.entry-points."ibis.backends"]
+nominal = "nominal.ibis"
```
alkasm, line 23: never seen this before, curious what this does/how it works! will look into it

tests/config/test_config.py
```diff
@@ -96,3 +96,25 @@ def test_nominal_config_rejects_unsupported_version(config_path: Path) -> None:
...
+            "from nominal.config import NominalConfig, NominalConfigError; "
+            "from nominal.core.exceptions import NominalConfigError as PublicError, NominalError; "
+            "assert NominalConfigError is PublicError; "
+            "assert issubclass(PublicError, NominalError)",
+        ],
+        check=False,
+        capture_output=True,
+        text=True,
+    )
+    assert result.returncode == 0, result.stderr
```
alkasm, line 120: this is a whack test

packages/nominal-ibis/nominal/ibis/_backend.py
```diff
@@ -0,0 +1,313 @@
...
+    def do_connect(self, client: NominalClient) -> None:
+        """Query the Nominal SQL API through an existing client.
+
+        Args:
+            client: Authenticated Nominal client. Queries run in its default workspace, either the one
+                pinned in the profile or the tenant default.
+        """
+        self._sql: sql_pb2_grpc.SqlServiceStub = client._clients.sql
+        self.workspace_rid = client._clients.resolve_default_workspace_rid()
+        self.__dict__.pop("_catalog", None)
```
alkasm, line 154: ? what is this for? if it's necessary can we get a comment?

packages/nominal-ibis/nominal/ibis/_backend.py
```diff
@@ -0,0 +1,313 @@
...
+
+    @functools.cached_property
+    def _catalog(self) -> sql_pb2.SqlCatalog:
+        with translate_grpc_errors():
+            return self._sql.GetSqlCatalog(sql_pb2.GetSqlCatalogRequest()).sql_catalog
```
alkasm, line 159: nit: prefer to separate request creation, network call, and field extraction into separate lines

nominal/core/_utils/grpc_tools.py
```diff
@@ -302,18 +304,37 @@ def _build_grpc_channel(api_base_url: str, service_config: ServiceConfiguration,
 }
 
 
+def _grpc_error_message(error: grpc.RpcError) -> str:
+    message = f"{error.code()}: {error.details()}"
+    for key, value in cast("tuple[tuple[str, str | bytes], ...]", error.trailing_metadata() or ()):
```
alkasm, line 309: nit: cast here seems overly complex and it's only because the empty tuple fallback I believe. this could be cleaner

pyproject.toml
```diff
@@ -176,6 +183,16 @@ disable_error_code = ["no-any-return", "no-untyped-def"]
...
 
+[[tool.mypy.overrides]]
+module = ["ibis", "ibis.*", "pyarrow", "pyarrow.*"]
+ignore_missing_imports = true
+
+[[tool.mypy.overrides]]
+module = "nominal.ibis.*"
+disallow_subclassing_any = false
+disallow_untyped_decorators = false
+disable_error_code = ["no-any-return", "empty-body"]
```
alkasm, line 194: all nominal packages should be strictly typed with no exceptions

pyproject.toml
```diff
@@ -176,6 +183,16 @@ disable_error_code = ["no-any-return", "no-untyped-def"]
 module = "nptdms"
 ignore_missing_imports = true
 
+[[tool.mypy.overrides]]
+module = ["ibis", "ibis.*", "pyarrow", "pyarrow.*"]
```
alkasm, line 187: pyarrow should have a type stub package available

nominal/_exceptions.py
sgupta229: makes sense. basically, nominal should only contain subpackages. will fix

packages/nominal-ibis/nominal/ibis/_backend.py
sgupta229: sorry, seems like this was left behind from previous commits

docs/src/reference/thirdparty/ibis.md
sgupta229: notes - tdms and ibis shouldn't be in a third party section. there is a nominal.thirdparty

nominal/core/_utils/grpc_tools.py
sgupta229:
```
StatusCode.INVALID_ARGUMENT: SQL query is invalid (instance ID 5e1f0c2a)
```

sgupta229: parses the rich metadata error out of gRPC

nominal/_exceptions.py
sgupta229: nominal.ibs relies on nominal being a namespace package. nominal is a pep420 implicit namespace package. these can only have subpackages in them.

nominal/core/_utils/grpc_tools.py
sgupta229: has type stubs so should no need to ignore typing

sgupta229: bad practice to have to repeat it every time you do this import

packages/nominal-ibis/nominal/ibis/__init__.py
```diff
@@ -0,0 +1,36 @@
...
+them through an authenticated NominalClient, streaming results back as Arrow.
+
+Requires a server that returns recursive catalog column types and applies the
+final SQL projection. MAP<ANY, ANY> columns, including points_struct.value,
+are exposed as strings containing Scout's JSON text; use json.loads() after
+fetching to access their contents. Concretely typed maps remain Ibis maps.
+Other unresolved ANY types require an explicitly typed projection via con.sql().
+
+Example:
+    import ibis
```
sgupta229: check why we don't have to register nominal.ibs for ibis to run our code

packages/nominal-ibis/pyproject.toml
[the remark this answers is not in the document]
sgupta229: notes - don't specify a transitive dependency. directly pull in the extra. 

packaging can probably be deleted

pyproject.toml
sgupta229: remove pyarrow from here since it does have typings. install in dev dependencies pyarrow-types.
- sometimes they are third party
- a lot of these are C / Rust libs that have bindings.

pyproject.toml
```diff
@@ -176,6 +183,16 @@ disable_error_code = ["no-any-return", "no-untyped-def"]
...
+ignore_missing_imports = true
+
+[[tool.mypy.overrides]]
```
sgupta229: delete this entire section

PR #950, opened 2026-09-03
feat: add script to generate 10k offset 30-minute runs

david-nominal:
## Intent

Add a standalone script that generates one continuous 1Hz sinus dataset and cuts 10,000 offset 30-minute runs from it. Each run has numeric typed property `run_instance` from 1 to 10,000 so the runs can be filtered in the Nominal UI.

## Trajectory

- Confirmed the dataset should be continuous 1Hz across the full timeline, including the 30-minute pauses between runs.
- Public `NominalClient.create_run` only accepts string `properties` and hardcodes `typed_properties={}`. The script therefore builds `CreateRunRequest` directly so `run_instance` is a numeric typed property.
- Chose chunked CSV.gz ingest (~1M rows per file) over per-sample streaming so ~36M points do not require a Python `enqueue()` loop.

## What it does

`uv run python scripts/generate_offset_runs.py --profile PROFILE`

- One dataset, channel `sine`, `sin(2π t / 60)` at 1Hz
- Run i occupies `[start + (i-1)*60min, start + (i-1)*60min + 30min)`
- Last pause is omitted: **35,998,200** samples, **36** upload chunks
- All runs attach the same dataset under ref name `dataset`
- `--count`, `--dry-run`, `--start`, durations, chunk size, and worker count are all overridable

Prove the path first with `--dry-run --count 2` before the full 10k create.

## Key decisions

- No SDK API change in this PR. Numeric properties stay on the raw Conjure request.
- Helpers (`total_samples`, `run_windows`, `sine_values`, `_create_run_request`) live in the script and are unit-tested offline.
- Parallel run creation with retry on 408/429/5xx.

## Potential pitfalls

- Full 10k create is a large ingest (~36M points) plus ~10k HTTP creates. Use `--count` to smoke-test against a live org.
- `search_runs(properties=...)` in this client is string-only. Filter `run_instance` in the Nominal UI, not via the Python search helper.
- Nominal auto-assigns `run_number`. That is not `run_instance`.

## Test plan

- [x] Offline unit tests for window math, sample count, sine continuity, CSV chunk timestamps, and numeric `typed_properties`
- [x] `--dry-run --count 2` and default 10k dry-run
- [ ] Live `--count 2` against a Nominal profile (not run here; no org credentials in this agent)

<div><a href="https://cursor.com/agents/bc-feebe448-ab00-4f3d-bf44-5410238ce4c6?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-feebe448-ab00-4f3d-bf44-5410238ce4c6&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -116,7 +116,7 @@ exclude = ["packages/**"]
 
 [tool.ruff]
 line-length = 120
-include = ["nominal/**/*.py", "packages/**/*.py", "tests/**/*.py"]
+include = ["nominal/**/*.py", "packages/**/*.py", "tests/**/*.py", "scripts/**/*.py"]
 
 [tool.ruff.lint.pydocstyle]
 convention = "google"
--- a/scripts/generate_offset_runs.py
+++ b/scripts/generate_offset_runs.py
@@ -1,0 +1,464 @@
+#!/usr/bin/env python3
+"""Generate one continuous 1Hz sinus dataset and cut offset 30-minute runs from it.
+
+Default layout (overridable via flags):
+
+- 10,000 runs, each 30 minutes long
+- 30-minute pause between runs
+- Continuous 1Hz sine on one channel across the full timeline (pauses included)
+- Each run has numeric typed property ``run_instance`` from 1 to N
+
+Public ``NominalClient.create_run`` only accepts string properties. This script
+uses ``CreateRunRequest.typed_properties`` so ``run_instance`` is numeric and
+filterable in the Nominal UI.
+
+Usage::
+
+    uv run python scripts/generate_offset_runs.py --profile PROFILE
+    uv run python scripts/generate_offset_runs.py --dry-run --count 2
+"""
+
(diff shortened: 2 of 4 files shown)


drake-nominal (changes requested):
I don't think this script belongs in this repository.

PR #956, opened 2026-09-08
fix: restrict plaintext grpc to literal loopback addresses

joyhe208:
An accidental `http://` production API URL must not enable unencrypted calls. Reject HTTP base URLs unless their host is a literal IPv4 loopback address (127.0.0.0/8) or IPv6 ::1. This is a client-wide construction requirement, including for callers that only intend to use Conjure HTTP APIs. HTTPS retains the existing TLS and certificate-verification path.

Desktop Core remains supported with `http://127.0.0.1:<port>/api`. Hostnames, including localhost, are rejected without DNS resolution; scoped and IPv4-mapped IPv6 forms and 0.0.0.0 remain excluded. No remote-plaintext override is introduced. This restriction does not add TLS to the Conjure transport.

Centralize URL validation before transport construction, use explicit loopback networks and a gRPC builder with direct returns, and document construction-time errors. Reject embedded user information for both HTTP and HTTPS. Diagnostics identify the host without exposing URL credentials, and CLI profile verification translates configuration failures into friendly Click errors.

### Testing plan

- Full SDK unit suite: 931 passed, 1 skipped, including real plaintext workspace server tests.
- Existing channel tests retain coverage of accepted loopback and rejected remote/ambiguous addresses. New regressions cover rejection before either transport is built, malformed ports, friendly CLI errors, and HTTP/HTTPS URL credential redaction.
- `mypy -p nominal` passes (159 source files); Ruff check, Ruff format check, and git diff check pass.
- Staging E2Es were not rerun locally for these changes. The two previously observed pandas failures also reproduced on main: https://github.com/nominal-io/nominal-client/actions/runs/34169279850.

### DIFF
--- a/docs/src/networking-tls.md
+++ b/docs/src/networking-tls.md
@@ -3,10 +3,16 @@
 The Nominal client talks to the platform over two transports: HTTP (conjure services) and gRPC
 (e.g. the Role Service). HTTPS URLs verify the server's TLS certificate on both transports.
 
-## Plaintext HTTP deployments
+## Local plaintext HTTP deployments
 
-An `http://` URL selects plaintext for both HTTP and gRPC on any host, including remote hosts.
-Credentials and custom headers are sent without TLS encryption. Use `https://` when TLS is required.
+An `http://` URL is accepted for gRPC only when its host is a literal loopback IP address:
+IPv4 `127.0.0.0/8` (normally `127.0.0.1`) or IPv6 `::1`. Use `https://` for remote deployments.
+Other HTTP hosts raise `NominalConfigError` during client construction, before a gRPC channel is created.
+There is no remote-plaintext override or automatic fallback from TLS to plaintext.
+
+Use `127.0.0.1` or `[::1]` rather than `localhost` or another hostname; the check does not trust DNS
+resolution. Loopback HTTP still sends credentials and custom headers without TLS encryption.
+This restriction applies to gRPC channel construction; it does not add TLS to the Conjure HTTP transport.
 
 For example, connect to a local runtime using its HTTP ingress URL, including `/api`:
 
--- a/docs/src/networking-tls.md
+++ b/docs/src/networking-tls.md
@@ -5,14 +5,16 @@ The Nominal client talks to the platform over two transports: HTTP (conjure serv
 
 ## Local plaintext HTTP deployments
 
-An `http://` URL is accepted for gRPC only when its host is a literal loopback IP address:
+An `http://` API base URL is accepted only when its host is a literal loopback IP address:
 IPv4 `127.0.0.0/8` (normally `127.0.0.1`) or IPv6 `::1`. Use `https://` for remote deployments.
-Other HTTP hosts raise `NominalConfigError` during client construction, before a gRPC channel is created.
+Other HTTP hosts raise `NominalConfigError` during client construction, even if you only intend to use
+Conjure HTTP APIs. URLs with embedded user information are rejected for both HTTP and HTTPS; supply
+your API token separately.
 There is no remote-plaintext override or automatic fallback from TLS to plaintext.
 
 Use `127.0.0.1` or `[::1]` rather than `localhost` or another hostname; the check does not trust DNS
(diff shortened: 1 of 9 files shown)


joyhe208 (commented):
@claude review

docs/src/networking-tls.md
```diff
@@ -3,10 +3,16 @@
 The Nominal client talks to the platform over two transports: HTTP (conjure services) and gRPC
 (e.g. the Role Service). HTTPS URLs verify the server's TLS certificate on both transports.
 
-## Plaintext HTTP deployments
+## Local plaintext HTTP deployments
 
-An `http://` URL selects plaintext for both HTTP and gRPC on any host, including remote hosts.
-Credentials and custom headers are sent without TLS encryption. Use `https://` when TLS is required.
+An `http://` URL is accepted for gRPC only when its host is a literal loopback IP address:
```
drake-nominal: Is it a safe assumption that a customer hosted deployment will _never_ use HTTP? What if we are hosting a local version of `scout` on our laptop and running as if it were a cloud deployment, but on e.g. `nominal.test` subnet?

nominal/core/_utils/grpc_tools.py
```diff
@@ -70,6 +72,19 @@ def api_base_url_to_grpc_target(api_base_url: str) -> str:
...
+    """Accept literal loopback IPs without trusting DNS resolution or scoped IPv6 addresses."""
+    if host is None or "%" in host:
+        return False
+    try:
+        address = ipaddress.ip_address(host)
+    except ValueError:
+        return False
+    if isinstance(address, ipaddress.IPv4Address):
+        return address.is_loopback
+    return address == ipaddress.IPv6Address("::1")
```
drake-nominal: Somewhat along the lines of the bot comments, this logic feels really fragile. I would also prefer a route that avoids the duck-typing `isinstance` checks.

nominal/core/_utils/grpc_tools.py
drake-nominal: Conventionally, we prefer to not do conditional assignment in this repo, e.g.:
  ```python
  if is_monday:
    a=123
  else:
    a=124
  ```

  Prefer a helper method with multiple returns instead to produce safer / more obvious code, such as some `_build_grpc_channel`.

nominal/core/_utils/grpc_tools.py
```diff
@@ -268,7 +283,13 @@ def create_grpc_channel(
         ("grpc.max_receive_message_length", _MAX_MESSAGE_LENGTH),
     ]
     target = api_base_url_to_grpc_target(api_base_url)
-    if urlparse(api_base_url).scheme == "http":
+    parsed = urlparse(api_base_url)
+    if parsed.scheme == "http":
+        if parsed.username is not None or parsed.password is not None or not _is_loopback_address(parsed.hostname):
+            raise NominalConfigError(
+                "Plaintext gRPC requires a literal loopback IP address (127.0.0.1 or [::1]). "
```
drake-nominal: why not 0.0.0.0

drake-nominal (changes requested):

docs/src/networking-tls.md
joyhe208: local dev uses https

joyhe208: on prem uses tls

joyhe208: so i think it's a safe assumption for now

nominal/core/_utils/grpc_tools.py
joyhe208: i want to keep it scoped down to the explicit loopback destination rather than the wildcard

nominal/core/_utils/grpc_tools.py
joyhe208: Checking explicit membership in 127.0.0.0/8 and ::1/128, without isinstance

nominal/core/_utils/grpc_tools.py
joyhe208: extracted _build_grpc_channel

drake-nominal (approved):

PR #966, opened 2026-09-10
chore: bump nominal-api and nominal-api-protos to 0.1437.0

sgupta229:
Bumps both generated-binding pins from 0.1379.0 to 0.1437.0. This is the first PR in a stack: 0.1437.0 is the first release that publishes the `SqlService` protos (scout#20687), which the Ibis backend in #948 needs.

One breaking change touches code the client constructs: `CreateDataset` gained a required `channel_search_split_tag_keys`. It applies only to derived datasets and is rejected otherwise, so the three call sites pass `[]`. The remaining breaking rows in the bump scan are new required fields on beans the client only reads.

### Testing plan

- `just verify` green (738 passed); mypy clean
- Bump scan (`reviewing-nominal-api-bumps`): proto side has no breaking change referencing the client; conjure side's 8 client-referenced rows are covered by the `CreateDataset` fix or are read-only beans

### DIFF
--- a/nominal/core/dataset.py
+++ b/nominal/core/dataset.py
@@ -1579,6 +1579,7 @@ def _create_dataset(
         is_v2_dataset=True,
         metadata={},
         origin_metadata=scout_catalog.DatasetOriginMetadata(),
+        channel_search_split_tag_keys=[],
         workspace=workspace_rid,
         marking_rids=list(marking_rids or []),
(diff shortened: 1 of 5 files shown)


drake-nominal:
already done and merged https://github.com/nominal-io/nominal-client/pull/961

PR #969, opened 2026-09-10
feat: diagnose audio timelines in normalize_video and repair only when needed

drake-nominal:
## Problem

A video was rejected during upload with *"This video's frame timestamps do not increase monotonically."* Every video timestamp in the file strictly increases. So does every audio timestamp. The message was wrong about the stream and wrong about the symptom, and the fix it suggested could not help.

The real fault is that an audio track carries **two independent claims about the same sound**. A container stores packet *durations* and derives start times by summing them; the codec fixes how many samples each packet decodes to. Nothing forces the two to agree. A lossy capture leaves the clock ahead of the sound (**holes**); a recorder flushing a late burst stamps packets closer together than the sound they carry (**overlap**).

A file can hold both at once. One real example measured **224s of holes and 47s of overlap**: a single net figure reports −177s and hides the overlap entirely, which is why the two are measured separately and never netted.

## Change

`normalize_video` measures the audio timeline before touching it, and applies `aresample=async=1` **only when the file actually fails** a strict, timestamp-preserving segmentation pass. That check asks ffmpeg whether the file segments rather than modelling its timestamp arithmetic, which is easy to get wrong and costs ~1s even on a 7GB file.

`async=1` is filling and trimming only. Higher values additionally permit *stretching* — resampling audio to chase a drifting clock, which shifts the timing of every later sample. The declared timeline is authoritative here, so content is adjusted to fit it. Swept across four orders of magnitude, every value ≥1 repaired the failures identically, so this takes the least invasive one.

New public API: `diagnose_audio`, `measure_audio_timeline`, `survives_strict_segmentation`, `timeline_from_packets`, and the `AudioTimeline` / `AudioDiagnosis` / `AudioDefect` model.

## Failing fast instead of failing late

After converting, the output is re-checked. If it still cannot be segmented, `normalize_video` raises with the remaining diagnosis rather than handing back a file that would be rejected on arrival — turning minutes of conversion plus upload into a few seconds and an actionable message.

That check also absorbs the awkward cases. Declining an unsafe repair — a gap longer than `max_audio_hole_seconds` (default 10s), which would otherwise synthesize that much silence from one bad timestamp; or audio outlasting the video, which no audio-only repair reconciles — does not have to decide on its own whether the file is doomed. It logs what it found, and the post-conversion check raises only if the result genuinely will not segment. A file with real holes that still segments is reported and returned.

Audio that already segments cleanly is never rewritten: rebuilding it could only move data that is currently fine. `repair_audio=False` opts out of inspection, repair, and verification together, so the escape hatch stays a complete one.

## What an operator sees

```
WARNING Repairing audio of 'example.mp4' with ffmpeg filter 'aresample=async=1':
        1.02s of missing audio across 5 gaps (largest 0.40s). The filter fills gaps
        with silence and drops audio that has no time to play, so the timeline the
        file declares is preserved exactly and no video or audio timestamp moves.
```

Every path says what was found, what is being done, and what is guaranteed. Files needing no repair log at DEBUG. Files left alone despite a defect log at WARNING, because the recording really is missing that audio and no conversion recovers it — that is a capture-side problem worth surfacing to whoever produced the file.

## Guarantee

**No video timestamp and no audio packet timestamp is ever moved.** Only audio content changes, to match the clock the file already declares.

## Testing

17 tests. Four are pure and run without ffmpeg, so a runner lacking it no longer passes while testing nothing.

- **Healthy audio is byte-identical** with the check on vs off, across 15 shapes: no audio, mono, 5.1, PCM, MP3, 44.1/48/96kHz, MJPEG/AVI, MKV, TS, VFR, two audio tracks, late audio, short audio.
- **Video PTS bit-identical** between repaired and unrepaired output.
- A file that fails to segment before is shown to segment after.
- Oversized holes and audio-outlasting-video decline the repair, and are caught by the post-conversion check.
- Container timestamp rounding is not mistaken for damage.

Packet times are read as seconds rather than raw timestamps: a packet's time base is `1/sample_rate` only for MP4/MOV, while Matroska uses milliseconds and MPEG-TS 90kHz. Reading raw values made healthy MKV and TS files report phantom defects. Deviations under a quarter packet are ignored so container rounding cannot accumulate into minutes of imaginary holes.

`ruff`, `mypy`, and the full suite (1003 passed) are clean.

## Notes for review

- This supersedes #963, which added the filter unconditionally. That would have been correct for the file that prompted it, but applies a repair to every file, including ones already fine.
- The policy is deliberately conservative: repair only what would otherwise fail. A file with real holes that still segments is reported and left alone.
- Two findings worth knowing beyond this PR: x264 output is not byte-reproducible for longer encodes, so byte-comparison assertions only hold for short fixtures; and holes and overlap frequently coexist in one file, so any future check that reports a single net figure will hide one of them.


drake-nominal:
Thanks — this was a genuinely useful pass. Six of these were real, four of them correctness bugs. All addressed in 0b67b9b0.

**Dropped packets manufactured the defect we detect.** The sharpest one. A packet with no `dts_time` was skipped, widening its neighbours' interval so it scored as a hole roughly two packets wide. Such a stream is now reported `UNMEASURABLE` and left alone rather than measured approximately.

**Failing to segment isn't proof of a timing fault.** Correct, and the consequence was exactly as described: a healthy file in a codec/container combination ffmpeg can't mux would have had its audio rebuilt while logging that its timeline was coherent. I took the cheaper of the two suggested fixes — the repair now requires a measured defect as well as a segmentation failure. I couldn't reproduce it with Opus (ffmpeg 8 muxes Opus into MPEG-TS fine) and my build has no libvorbis, but the structural point stands independently: applying `aresample` to a coherent timeline can only do harm, since there's nothing for it to correct.

**Opus/Vorbis variable frame sizes.** Right, and it compounds the above. Both dropped from `_CONSTANT_FRAME_CODECS`; they now report unmeasurable.

**The overrun guard compared fields that are often absent.** MKV and TS routinely leave per-stream `duration` as `N/A`. It now uses the measured timeline span, which we already have at that point in the ladder.

**`defect` stored rather than derived.** Fixed — it's a property alongside `audio_seconds`, so it can't disagree with the timeline it summarises, and two constructor arguments disappeared.

**Cost ordering.** Fair hit: the module docstring promised ordering the code didn't do, and an 8s packet walk ran on every healthy file to pick a log level. `audio_repair_filter` now probes the header, checks segmentation, and returns before measuring when the file already segments. `diagnose_audio` still always measures — it's the API for a complete answer — and now says so. The double header probe is gone via an optional `info` parameter.

**Tests.** The byte-identity claim was verified by a scratch script, not by the committed test — you were right that the test covered one shape while the description claimed fifteen. Now parameterized over six real container/codec/rate/layout combinations including MKV and TS, with a PCM case pinning `UNMEASURABLE`. The helper duplicating the production flags is gone in favour of calling `survives_strict_segmentation` directly.

**`video_duration_seconds`** is now `_video_duration_seconds` — it's an implementation detail, not API.

One note on the first comment: the stale `Raises:` block and `max_audio_hole_seconds` wording were already fixed in 22ecbb1b, which also re-introduced `AudioTimelineError` with a narrower meaning — it now fires only when the *converted* file still cannot be segmented, so we fail in seconds instead of after a long upload that would be rejected on arrival.

23 tests, 1009 in the full suite, ruff and mypy clean. The 15-shape corpus is still byte-identical with inspection on versus off.

PR #974, opened 2026-09-11
feat: default streaming to rust, collapse implementations to python and rust

drake-nominal:
Changes the default write stream implementation from json to rust.

```python
dataset.get_write_stream()                         # rust
dataset.get_write_stream(implementation="python")  # pure python
```

Falls back to `python` when `nominal-streaming` is not installed, which is the case on platforms it ships no binary for. Default batching changes from 50k points / 1s to 250k points / 0.25s.

## Why

Rust streaming is faster than the python paths, accepts the same enqueue calls, and can write failed batches to a fallback file. It is no longer experimental.

The argument is renamed from `data_format` to `implementation`, because the choice is an implementation and not a wire format. Callers do not need to know whether the python stream sends json or protobuf, so both names now mean `python`.

## Compatibility

| old | new | status |
|---|---|---|
| `data_format=` | `implementation=` | deprecated, still works |
| `"json"` | `"python"` | deprecated, still works |
| `"protobuf"` | `"python"` | deprecated, still works |
| `"rust_experimental"` | `"rust"` | deprecated, still works |
| `"experimental"` | unchanged | deprecated, not an alias yet |

Every call that worked before still runs. Deprecated spellings emit a `UserWarning`, the argument one through the existing `warn_on_deprecated_argument`.

Two behavior changes:

- `implementation="json"` now sends protobuf.
- `flush()` is a no-op that warns, where the python stream drains. It is not on `WriteStreamBase`, so type-checked callers never had it, and `close()` still gives the drain it promised.


alxhill (approved):
high-level lgtm!

PR #975, opened 2026-09-14
perf: stream Python logs with columnar protobuf

drake-nominal:
Depends on #976.

Send `Dataset.get_log_stream()` batches as columnar protobuf instead of per-channel JSON requests. Group channels into one request while preserving timestamps, messages, and per-record arguments. Keep the Python interface and batching defaults. Use the shared zstd adapter from #976.

Benchmarked the public stream on staging on September 17 using Python 3.13 on macOS arm64. Compared JSON+zstd (`504e70ff`) with columnar+zstd (`24a83a52`), both at zstd level 1. Ran each workload in JSON/columnar/columnar/JSON order with a one-second flush interval, three in-flight batches, and retries disabled. Timed record generation through stream close and final upload acknowledgement. Each phase used 200–300,000 records, except the paced workload, which used 30.

Each metric below is **JSON → columnar**. Points/s measures acknowledged upload throughput. MB/s and bytes/point measure compressed request bodies; MB is decimal and excludes HTTP/TLS overhead. Rates aggregate the two runs per format by total records or bytes divided by total elapsed time. Batch limits are maxima; the flush timer can send smaller batches.

| Workload | Batch limit | Points/s | MB/s | Bytes/point |
| --- | ---: | ---: | ---: | ---: |
| 8-byte messages | 1 | 22 → 24 | 0.0029 → 0.0031 | 130.25 → 133.50 |
| 8-byte messages | 100 | 2,101 → 2,080 | 0.009 → 0.014 | 4.07 → 6.81 |
| 8-byte messages | 1,000 | 13,360 → 15,444 | 0.041 → 0.073 | 3.05 → 4.72 |
| 256-byte messages, 12 arguments | 100 | 1,995 → 2,046 | 0.034 → 0.042 | 16.97 → 20.68 |
| 256-byte messages, 12 arguments | 10,000 | 32,122 → 46,157 | 0.471 → 0.522 | 14.67 → 11.30 |
| 256-byte messages, 12 arguments | 50,000 | 32,381 → 53,562 | 0.480 → 0.631 | 14.83 → 11.77 |
| 4 KiB repeated messages | 1,000 | 5,133 → 5,560 | 0.032 → 0.043 | 6.21 → 7.70 |
| 4 KiB high-entropy messages | 1,000 | 1,003 → 1,047 | 3.359 → 3.497 | 3348.11 → 3341.40 |
| 80-byte messages, 32 arguments | 5,000 | 11,078 → 15,147 | 0.292 → 0.381 | 26.37 → 25.15 |
| 32-byte messages, 8 channels | 1,000 | 1,682 → 6,027 | 0.011 → 0.031 | 6.45 → 5.10 |
| 32-byte messages, 64 channels | 1,000 | 241 → 5,904 | 0.004 → 0.032 | 16.19 → 5.40 |
| 10 points/s producer, 4 arguments | 50,000 | 9.4 → 9.6 | 0.0003 → 0.0003 | 30.73 → 34.78 |

Structured logs improved 65% at the default 50k batch limit and 44% at 10k. With 64 channels, throughput improved 24.5× as each full batch used one request instead of 64. Faster upload can consume more bandwidth per second despite fewer bytes per point.

Columnar is not uniformly better. At 100 tiny records per batch, throughput was effectively unchanged and compressed bytes/point increased 67%. Small structured batches also grew on the wire. Sparse traffic and high-entropy 4 KiB messages showed little throughput benefit. With only two runs per format, small differences should not be treated as reliable gains. These finite synthetic runs do not establish sustained API capacity or request-size limits.

A separate 17-shape local benchmark measured serialization, compression, and request preparation with HTTP replaced by a successful response. Columnar took less time in every tested shape, with 1.14–9.00× higher throughput (median of five samples after warmup). This isolates client work; it does not predict the same improvement over the network.

All 2,064,920 records received successful HTTP responses across 3,209 requests, with no failures or retries. Query readback matched all 64,992 checked records: exhaustive checks for the small and multi-channel runs, and boundary samples for larger runs. The six previously absent benchmark samples are now readable without resending; all 24 old samples match. These checks do not establish completeness for every uploaded record, so the table reports acknowledged throughput.


drake-nominal:
Closing after benchmarking. The largest gains came from large structured batches and many channels. Expected workloads use small batches across a small number of channels, where we measured little throughput benefit and sometimes larger compressed payloads. The results do not justify adopting columnar log streaming for those workloads at this time. The benchmark results remain in the description for future reference.

PR #979, opened 2026-09-14
feat: support typed properties on assets, runs, datasets, and events

david-nominal:
## Summary
- Adds `TypedProperties` (`str | float`) for assets, runs, datasets, and events, plus `property_filters` (`eq`, `in_`, `gt`, `between`, ...) on asset/run/dataset search.
- Events use proto `typed_properties` (gRPC) after the 0.1451.0 API bump; assets/runs/datasets use conjure. Containerized extractors stay `Mapping[str, str]` — their proto is still a string map.
- Branch is current with `origin/main` (includes #992).

## Trajectory
Started as numeric properties on assets/runs/datasets. Merging `main` brought #981 extractor properties and proto `typed_properties` on events. Extractors needed no adaptation; events did. Fork PR #990 was a temporary stand-in while origin push was unavailable; this branch now has that work.

## Key decisions
- Public stored type is `str | float`. Runtime accepts `int` and coerces to `float`; `bool` is rejected. `int` is not advertised.
- Create sends `properties={}` plus the typed map. Reads use `typed_properties` only.
- `properties=` on `search_assets` / `search_runs` / `search_datasets` is deprecated in favor of `property_filters`. Event search stays string equality (`SearchQuery` has no numeric predicates).
- String `eq` is a singleton `in_`.

## Potential pitfalls
- `Asset` / `Run` / `Dataset` / `Event`.properties widen from `Mapping[str, str]` to `Mapping[str, str | float]` (typed breaking change).
- Existing resources that only have a legacy string map will read back empty until the backend mirrors strings into `typed_properties`.
- Event search cannot express numeric filters yet.

## Test plan
- [x] Unit tests for typed property conversion, filters, event create/update/read
- [x] Extractor tests still pass with string property maps
- [ ] e2e numeric property CRUD on assets, runs, datasets, and events
- [ ] e2e search filters (EQ / gt / between) on assets, runs, datasets


nominal/core/client.py
```diff
@@ -802,7 +809,8 @@ def search_runs(
         name_substring: str | None = None,
         *,
         labels: Sequence[str] | None = None,
-        properties: Mapping[str, str] | None = None,
+        properties: Mapping[str, PropertyValue] | None = None,
+        property_filters: Sequence[NumericPropertyFilter] | None = None,
```
drake-nominal: should we deprecate `properties` and build out more featured checks for inclusion / exclusion etc. within some `PropertyFilter` class?

We've long debated the idea of "is our search API complicated enough to warrant builder classes for queries?", to me this feel like an opportunity to really solidify a solid pattern to make it easy to use our search API.

nominal/core/dataset.py
```diff
@@ -89,7 +94,7 @@ def update(
             description=description,
             labels=None if labels is None else list(labels),
             name=name,
-            properties=None if properties is None else dict(properties),
+            typed_properties=None if properties is None else typed_properties_to_conjure(properties),
```
drake-nominal: would be sick if `typed_properties_to_conjure()` handled the optional part too (e.g. take a None, return a None), but maybe that makes other parts of the code harder to use / type check.

nominal/core/run.py
```diff
@@ -48,7 +48,7 @@ class Run(HasRid, RefreshableConjureMixin[scout_run_api.Run], _DatasetWrapper):
     rid: str
     name: str
     description: str
-    properties: Mapping[str, str]
+    properties: Mapping[str, str | float]
```
drake-nominal: PropertyValue. Maybe worth defining `Mapping[str, PropertyValue]` as something like `TypedProperties`

nominal/core/_utils/properties.py
```diff
@@ -0,0 +1,160 @@
...
+
+from dataclasses import dataclass
+from types import MappingProxyType
+from typing import Literal, Mapping, TypeAlias
+
+from nominal_api import api
+
+PropertyValue: TypeAlias = str | int | float
+
+_COMPARISON_OPERATORS: dict[str, api.PropertyComparisonOperator] = {
```
drake-nominal: IMO, something you could represent maybe more cleanly as an enum-- we have lots of precedent in the repo for enums that wrap conjure-based enums, you could replicate that with `PropertyComparisonOperator`

nominal/core/_utils/properties.py
```diff
@@ -0,0 +1,160 @@
...
+    return NumericPropertyFilter(name=name, operator="lte", value=_as_numeric_value(value, what="value"))
+
+
+def between(name: str, min_value: int | float, max_value: int | float) -> NumericPropertyFilter:
+    return NumericPropertyFilter(
+        name=name,
+        operator="between",
+        min_value=_as_numeric_value(min_value, what="min_value"),
+        max_value=_as_numeric_value(max_value, what="max_value"),
+    )
```
drake-nominal: these really generic method names won't scale well if, for example, we expose other similar comparison things. Would love it if instead there was some builder object that we are calling these methods on that produces some `PropertyFilter` to be used, or even some `PropertyQuery` if we want it to fit in with the family of `build_search_query` methods

nominal/core/__init__.py
```diff
@@ -106,4 +120,12 @@
...
     "WorkspaceSearchType",
     "WriteStream",
+    "between",
+    "eq",
+    "gt",
+    "gte",
+    "lt",
+    "lte",
+    "neq",
+    "not_between",
```
drake-nominal: I think these are too generically named to be in the core module init.py.

nominal/core/asset.py
```diff
@@ -773,7 +773,7 @@ def _from_conjure(cls, clients: _Clients, asset: scout_asset_api.Asset) -> Self:
             rid=asset.rid,
             name=asset.title,
             description=asset.description,
-            properties=MappingProxyType(asset.properties),
+            properties=typed_properties_from_conjure(asset.typed_properties),
```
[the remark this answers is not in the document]
david-nominal: anything in the string only map is also returned in the typed map from the BE!

nominal/core/asset.py
```diff
@@ -125,7 +125,7 @@ def update(
         request = scout_asset_api.UpdateAssetRequest(
             description=description,
             labels=None if labels is None else list(labels),
-            properties=None if properties is None else dict(properties),
+            typed_properties=None if properties is None else typed_properties_to_conjure(properties),
```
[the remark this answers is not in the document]
david-nominal: not true, in the BE properties and typed_properties are stored in the same table. properties at this point is just the string type subset of typed_properties. not setting/using properties and just use typed_properties instead is the way to go

nominal/core/_utils/query_tools.py
```diff
@@ -259,9 +262,16 @@ def create_search_assets_query(
...
             queries.append(scout_asset_api.SearchAssetsQuery(label=label))
-    if properties:
-        for name, value in properties.items():
-            queries.append(scout_asset_api.SearchAssetsQuery(property=api.Property(name=name, value=value)))
+    queries.extend(
+        iter_property_filter_clauses(
+            properties,
+            property_filters,
+            query_cls=scout_asset_api.SearchAssetsQuery,
+            string_in_clause=lambda name, values: scout_asset_api.SearchAssetsQuery(
```
david-nominal, line 270: a bit annoying that we have to pass this in but it's unfortunately currently not uniform across the three backend APIs ...

david-nominal:
@claude re-review

drake-nominal:
note: #981 merged which adds new properties into the repo

docs/src/reference/toplevel.md
```diff
@@ -1,6 +1,7 @@
 Also see the [Python Quickstart guide](https://docs.nominal.io/core/sdk/python-client/quickstart). 
 
 [`nominal.core`](./core.md) — platform client.  
+[`nominal.core.properties`](./properties.md) — property filters (`eq`, `gt`, `between`, ...).
```
drake-nominal, line 4: I don't think this needs to become a top level page in the documentation, we normally don't register individual files like this, just packages.

nominal/core/_utils/api_tools.py
```diff
@@ -107,6 +108,15 @@ def property_update(properties: Mapping[str, str] | None) -> types_pb2.PropertyU
     return None if properties is None else types_pb2.PropertyUpdateWrapper(properties=dict(properties))
 
 
+def typed_property_update(properties: TypedProperties | None) -> types_pb2.TypedPropertyUpdateWrapper | None:
+    """Wrap typed properties for a proto update request: None omits the field, any mapping replaces it."""
+    from nominal.core._utils.properties import typed_properties_to_proto
```
drake-nominal, line 113: ideally we don't lazy import this and just import at the top of the file.

tests/core/test_event.py
```diff
@@ -133,6 +136,52 @@ def test_search_event_origin_type_round_trips(origin_type: SearchEventOriginType
...
+    clients = MagicMock()
+    event = _event(clients)
+    clients.event.BatchUpdateEvent.return_value = event_pb2.BatchUpdateEventResponse(events=[_proto_event()])
+
+    event.update(properties={"serial": "A1", "mass_kg": 7.5}, type=None)
+
+    update = clients.event.BatchUpdateEvent.call_args.args[0].updates[0]
+    assert not update.HasField("properties")
+    assert update.typed_properties.typed_properties["serial"].string_value == "A1"
+    assert update.typed_properties.typed_properties["mass_kg"].numeric_value == 7.5
```
drake-nominal, line 182: Some of these tests are just kinda testing routine plumbing, I wonder if we can consolidate to tests of the helpers we added and remove some of the testing footprint without really losing much.

nominal/core/properties.py
```diff
@@ -0,0 +1,245 @@
...
+    """Inclusive range filter: ``min_value <= property <= max_value``."""
+    return _range(name, NumericPropertyRangeOperator.BETWEEN, min_value, max_value)
+
+
+def not_between(name: str, min_value: float, max_value: float) -> NumericRangeFilter:
+    """Filter matching values outside the inclusive range ``[min_value, max_value]``."""
+    return _range(name, NumericPropertyRangeOperator.NOT_BETWEEN, min_value, max_value)
+
+
+__all__ = [
```
drake-nominal, line 230: we don't normally do `__all__` within files like this, preferring to leave methods named with leading `_` to denote private vs. explicitly whitelisting the public methods.

nominal/core/properties.py
```diff
@@ -0,0 +1,245 @@
...
+    "between",
+    "eq",
+    "gt",
+    "gte",
+    "in_",
+    "lt",
+    "lte",
+    "neq",
+    "not_between",
+]
```
drake-nominal, line 245: I'd love to think through some ways where maybe we can reduce the amount of symbols we are adding into the namespace, especially with such short method names that could easily be masked by other similar functionality.

What if we had some class like:

```
PropertyFilter
```

With a bunch of classmethods like:

```
PropertyFilter.eq()
PropertyFilter.between()
```

etc. so that it's all centralized within one type?

david-nominal: that's fair, I was worried about the verbosity but it's not that much really.

nominal/core/dataset.py
```diff
@@ -51,7 +55,7 @@
 class Dataset(DataSource, RefreshableConjureMixin[scout_catalog.EnrichedDataset]):
     name: str
     description: str | None
-    properties: Mapping[str, str]
+    properties: TypedProperties
```
alkasm, line 58: this is a breaking change

david-nominal: it would be a lint failure but not a runtime failure, right? What's the process/policy for dealing with that here?
I could have a `properties_v2: TypedProperties` and keep the `properties: [str, str]` map around?

